PR #43, opened 2017-09-06
New logging infrastructure and code analysis

lubeltra:
Added new logging infrastructure that exposes a C api.
Enabled code analysis and fixed some SAL issues
Minor tweak to the project file generator to better support different locations of the clone


Source/Logger/trace.cpp
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+private:
+    std::atomic<uint32_t> m_tracingClients = 0;
+    std::atomic<std::chrono::high_resolution_clock::time_point> m_initTime =
+        std::chrono::high_resolution_clock::now();
+    std::atomic<HCTraceCallback*> m_clientCallback = nullptr;
+};
+
+TraceState g_traceState;
```
jasonsandlin: global state.  can we avoid and put it in the singleton that's created upon hcglobalinit()?

lubeltra: I was trying to avoid having a dependency from the logging code to the rest of LibHttpClient, which is why it has its own global init and cleanup.

The main reason for this is so that it can be used by other consumers without bringing in the rest of LibHttpClient (if they want)

PR #46, opened 2017-09-12
Improving singleton logic

jasonsandlin:


Source/Common/singleton.cpp
```diff
@@ -30,33 +29,59 @@ http_singleton::http_singleton()
 
 http_singleton::~http_singleton()
 {
+    for (auto& mockCall : m_mocks)
```
lubeltra: Maybe should just call HCMockClearMocks()

Source/Common/singleton.cpp
```diff
@@ -75,13 +100,13 @@ std::shared_ptr<http_task_completed_queue> http_singleton::get_task_completed_qu
     return taskQueue;
 }
 
-void verify_http_singleton()
+void verify_http_singleton(_In_ std::shared_ptr<http_singleton>& httpSingleton)
```
lubeltra: get_http_singleton() already does the checking, do we really want a second function?

Source/HTTP/httpcallresponse.cpp
```diff
@@ -23,10 +22,26 @@ HCHttpCallResponseSetResponseString(
...
     )
 {
-    verify_http_singleton();
     call->responseString = responseString;
-    HC_TRACE_INFORMATION(HTTPCLIENT, "HCHttpCallResponseSetResponseString [ID %llu]: responseString=%s",
-        call->id, responseString);
+
+#if HC_TRACE_ENABLE
+
+#if HC_CHAR_T_IS_WIDE
```
lubeltra: Consider adding an tstrlen function that is defined for the correct variant of CHAR_T to avoid this split code (look at the various printf variants in trace.cpp for what I mean)

Source/HTTP/httpcallresponse.cpp
```diff
@@ -23,10 +22,26 @@ HCHttpCallResponseSetResponseString(
...
+    size_t len = wcslen(responseString);
+#else
+    size_t len = strlen(responseString);
+#endif
+    if (len > 2048)
```
lubeltra: consider using %.2048s to limit the number of characters printed

Source/Common/singleton.cpp
jasonsandlin: Ah, I'll remove

Source/Common/singleton.cpp
jasonsandlin: It seems safer to do the cleanup on the object being destroyed vs calling ClearMocks which cleans up the global again.

lubeltra: Yes, I'd just like to avoid duplicating the code, maybe a helper function on the singleton that's called from both places?

Source/HTTP/httpcallresponse.cpp
jasonsandlin: Ah, that worked

PR #49, opened 2017-09-13
Updates based on API review feedback

jasonsandlin:


Include/httpClient/httpClient.h
```diff
@@ -3,12 +3,14 @@
 
 #pragma once
 #include "types.h"
+#include "task.h"
+#include "httpProvider.h"
```
lubeltra: I think Cullen was suggesting that this doesn't need to be included in the base header, only people implementing a provider need it

Include/httpClient/httpClient.h
```diff
@@ -3,12 +3,14 @@
...
+#include "mock.h"
```
lubeltra: Same for this header, should only be included by people that want to do mocks, where they implement the mocks

Include/httpClient/httpClient.h
```diff
@@ -679,101 +286,109 @@ HCHttpCallRequestGetRequestBodyString(
...
 /// Defaults to true 
 /// </summary>
-/// <param name="call">The handle of the HTTP call</param>
+/// <param name="call">The handle of the HTTP call.  Pass nullptr to set the default for future calls</param>
 /// <param name="retryAllowed">If retry is allowed for this HTTP call</param>
 HC_API void HC_CALLING_CONV
 HCHttpCallRequestSetRetryAllowed(
-    _In_ HC_CALL_HANDLE call,
+    _In_opt_ HC_CALL_HANDLE call,
     _In_ bool retryAllowed
```
lubeltra: Bool is not a type in C

Include/httpClient/httpClient.h
```diff
@@ -783,19 +398,8 @@ HCHttpCallRequestGetTimeout(
...
-/// <summary>
-/// Set the response body string of the HTTP call
-/// </summary>
-/// <param name="call">The handle of the HTTP call</param>
-/// <param name="responseString">the response body string of the HTTP call</param>
-HC_API void HC_CALLING_CONV
-HCHttpCallResponseSetResponseString(
-    _In_ HC_CALL_HANDLE call,
-    _In_ PCSTR_T responseString
+    _In_opt_ HC_CALL_HANDLE call,
```
lubeltra: Why is this optional? (also applies to the other getters below)

Include/httpClient/trace.h
```diff
@@ -267,15 +267,15 @@ HC_API void HC_CALLING_CONV HCTraceSetClientCallback(HCTraceCallback* callback);
 
 #define HC_PRIVATE_TRACE_AREA_NAME(area) g_trace##area
 
-#if HC_CHAR_T_IS_WIDE
+#if HC_PCSTR_IS_WIDE
```
lubeltra: This should be unnecessary now

Samples/Win32Http/main.cpp
```diff
@@ -60,19 +60,18 @@ DWORD g_numActiveThreads = 0;
...
     {
-        HCTaskGetPendingHandle(),
         g_stopRequestedHandle.get()
     };
 
     bool stop = false;
     while (!stop)
     {
-        DWORD dwResult = WaitForMultipleObjectsEx(2, hEvents, false, INFINITE, false);
+        DWORD dwResult = WaitForMultipleObjectsEx(1, hEvents, false, 20, false);
```
lubeltra: You could use WaitForSingleObject now

Source/HTTP/httpcallrequest.cpp
```diff
@@ -9,8 +9,8 @@ using namespace xbox::httpclient;
 HC_API void HC_CALLING_CONV
 HCHttpCallRequestSetUrl(
     _In_ HC_CALL_HANDLE call,
-    _In_ PCSTR_T method,
-    _In_ PCSTR_T url
+    _In_ PCSTR method,
```
lubeltra: Why not use an enum for the method?

Source/Logger/trace.cpp
```diff
@@ -15,12 +15,12 @@ namespace
 {
 
 //------------------------------------------------------------------------------
```
lubeltra: This entire section can be deleted (I'm kind of surprised it even compiles right now)

Source/Logger/trace.cpp
```diff
@@ -165,9 +165,9 @@ void TraceMessageToDebugger(
     std::tm      fmtTime = {};
     localtime_s(&fmtTime, &timeTInSec);
 
-    CHAR_T outputBuffer[BUFFER_SIZE] = {};
+    CHAR outputBuffer[BUFFER_SIZE] = {};
```
lubeltra: char (lowercase)

lubeltra: happens in several places

Include/httpClient/httpClient.h
jasonsandlin: As doc'd in header if null, then it sets default.

Source/Logger/trace.cpp
jasonsandlin: fixed

Include/httpClient/trace.h
jasonsandlin: removed

Include/httpClient/httpClient.h
jasonsandlin: Should be alright.  https://stackoverflow.com/a/1608349/7655781

Source/Logger/trace.cpp
jasonsandlin: Removed

Include/httpClient/httpClient.h
lubeltra: make sure you include stdbool.h then

I've definitely seen vs barf on it without any includes

Include/httpClient/httpClient.h
lubeltra: There is a default response string?

I guess that can make sense, seems a bit weird though

Source/Logger/trace.cpp
lubeltra: Doesn't look you have fixed the places these functions were being called

Include/httpClient/trace.h
```diff
@@ -266,12 +266,7 @@ HC_API void HC_CALLING_CONV HCTraceSetClientCallback(HCTraceCallback* callback);
 // DO NOT USE THESE SYMBOLS DIRECTLY
 
 #define HC_PRIVATE_TRACE_AREA_NAME(area) g_trace##area
-
-#if HC_PCSTR_IS_WIDE
-#define HC_FUNCTION __FUNCTIONW__
-#else
 #define HC_FUNCTION __FUNCTION__
```
lubeltra: This is unnecessary, we can just use the standard __FUNCTION__ macro now

Include/httpClient/httpClient.h
jasonsandlin: Oh right, this one shouldn't be _In_opt_

Source/Logger/trace.cpp
```diff
@@ -14,42 +14,6 @@
...
-
-void OutputDebugStringT(wCHAR const* string)
-{
-    OutputDebugStringW(string);
-}
-
-#else
 
 template<size_t SIZE>
 int stprintf_s(char(&buffer)[SIZE], _Printf_format_string_ char const* format ...)
```
lubeltra: Now that my crazy overloads are gone, we need to change the code that calls to them

jasonsandlin: Feel free to fix in a new PR

PR #50, opened 2017-09-15
More flushout of Win32 support

jasonsandlin:


Samples/Win32Http/Win32Http.vcxproj
```diff
@@ -92,10 +92,9 @@
       copy $(MSBuildProjectDirectory)\..\..\Binaries\Debug\x64\libHttpClient.141.Win32.C\libHttpClient.141.Win32.C.pdb $(OutputPath)</Command>
     </CustomBuildStep>
     <CustomBuildStep>
-      <Outputs>$(OutputPath)\libHttpClient.141.Win32.C.dll</Outputs>
+      <Outputs>always</Outputs>
```
lubeltra: What does this do?

jasonsandlin: Its for the sample to copy the DLL locally always as it wasn't correctly updating it as it changed.  There may be a better way but it'll be removed when we move to static libs

PR #51, opened 2017-09-15
API review feedback part 2

jasonsandlin:


Include/httpClient/httpClient.h
```diff
@@ -141,6 +144,21 @@ typedef enum HC_LOG_LEVEL
     LOG_ERROR,
 
     /// <summary>
+    /// Log warnings and errors
+    /// </summary>
+    LOG_WARNING,
```
lubeltra: Why not typedef HCTraceLevel HC_LOG_LEVEL? So we don't have 2 types floating about that get used in different places

Include/httpClient/types.h
```diff
@@ -81,3 +81,50 @@
...
+    HC_E_POINTER = -2,
+    HC_E_INVALIDARG = -3,
+    HC_E_OUTOFMEMORY = -4,
+    HC_E_BUFFERTOOSMALL = -5,
+    HC_E_NOTINITIALISED = -6,
+    HC_E_FEATURENOTPRESENT = -7,
+} HC_RESULT;
+
+
+#define CONVERT_STD_EXCEPTION(operation) \
```
lubeltra: Instead of using these macros around places that we think will throw, I would suggest making sure that all api entry points have a try catch block that translates exceptions at that level (you can look at how I set up Xal for an example).

The advantage is that it's a lot easier to audit the code to be sure that no exceptions leak.
I've even marked all my C boundary functions with noexcept so that the compiler will emit a crash if an exception were to leave for an extra layer of safety (well determinism anyway)

Source/Common/singleton.cpp
```diff
@@ -65,8 +66,15 @@ void init_http_singleton()
             &httpSingleton,
             newSingleton
         );
```
lubeltra: Just realised this code does not allocate memory the right way, we should pass an allocator to make shared...

Source/Common/singleton.cpp
```diff
@@ -65,8 +66,15 @@ void init_http_singleton()
...
+
+        if (newSingleton == nullptr)
```
lubeltra: make_shared expects the allocator to throw on alloc failure, so you need your macro

Include/httpClient/types.h
jasonsandlin: I'll look into this in a future PR

Include/httpClient/httpClient.h
jasonsandlin: I'd prefer no typedef and only one public type.  For a future PR

lubeltra: HCTraceLevel is in the enum namespace in C, so we need to typdef it into a the type namespace anyway

Source/Common/singleton.cpp
jasonsandlin: Got it

Source/Common/singleton.cpp
jasonsandlin: Will do

PR #55, opened 2017-09-19
Fixing mem allocs

jasonsandlin:


Source/Global/mem.h
```diff
@@ -75,6 +90,21 @@ class http_stl_allocator
     }
 };
 
+template<typename T>
+std::shared_ptr<T> http_allocate_shared()
```
lubeltra: template<typename T, typename... Args>
std::shared_ptr<T> http_allocate_shared(Args&&... args)
{
    return std::allocate_shared<T>(http_stl_allocator<T>(), std::forward<Args>(args)...);
}

1. Made into a variadic template to allow passing arguments through to the constructor
2. Avoid specifying template arguments that can be deduced
3. Written in the browser so it may fail to compile

Source/Global/mem.h
```diff
@@ -75,6 +90,21 @@ class http_stl_allocator
...
+{
+    return std::allocate_shared<T, http_stl_allocator<T>>(http_stl_allocator<T>());
+}
+
+template<typename T>
+std::unique_ptr<T, http_alloc_deleter<http_stl_allocator<T>>> http_allocate_unique()
```
lubeltra:
```c++
template<typename T>
using HC_UNIQUE_PTR = std::unique_ptr<T, http_alloc_deleter<http_stl_allocator<T>>>;

template<typename T, typename... Args>
HC_UNIQUE_PTR<T> http_allocate_shared(Args&&... args)
{
    http_stl_allocator<T> alloc;
    auto p = std::allocator_traits<http_stl_allocator<T>>::allocate(alloc, 1);
    auto o = new(p) T(std::forward<Args>(args)...);
    return HC_UNIQUE_PTR<T>(o);
}
```

1. Made into a variadic template to allow passing arguments through to the constructor
2. Added template alias for std::unique_ptr with custom deleter
3. Used the pointer returned by operator new instead of the one returned by the allocator (they are probably identical but...)
4. Written in the browser so it may fail to compile

Source/Task/task_impl.h
```diff
@@ -45,6 +45,8 @@ struct HC_TASK
 #endif
 };
 
+typedef std::unique_ptr< HC_TASK, http_alloc_deleter<http_stl_allocator<HC_TASK>> > HC_TASK_PTR;
```
lubeltra: typedef HC_UNIQUE_PTR<HC_TASK> HC_TASK_PTR

Source/Global/mem.h
```diff
@@ -48,20 +48,35 @@ class http_memory_buffer
 
 NAMESPACE_XBOX_HTTP_CLIENT_END
 
+template<typename TAllocator>
+struct http_alloc_deleter
```
lubeltra: this doesn't need to be templated on the allocator type, just the object type. It would save some typing elsewhere

lubeltra: I would also add a template alias for the allocator traits to avoid having to repeat that everywhere

jasonsandlin: done

Source/Global/mem.h
jasonsandlin: done

Source/Global/mem.h
jasonsandlin: done

Source/Task/task_impl.h
jasonsandlin: done

PR #56, opened 2017-09-19
Flushing out mock api

jasonsandlin:


jasonsandlin:
Yeah, I'll do that later

PR #86, opened 2018-01-05
Adding XDK support

jasonsandlin:


Source/HTTP/XMLHttp/xmlhttp_http_task.cpp
```diff
@@ -123,6 +122,7 @@ void xmlhttp_http_task::perform_async(
             return;
         }
 
+#if !HC_XDK_API
```
jplafonta: why is this not needed on xdk?  looks like m_hRequestBodyStream is used below

jasonsandlin: Yeah, working on fixing this next

PR #99, opened 2018-02-10
Adding WinHTTP support

jasonsandlin:


Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -0,0 +1,794 @@
...
+
+bool winhttp_http_task::has_error()
+{
+    return m_exceptionPtr != nullptr;
+}
+
+void winhttp_http_task::set_exception(const std::exception_ptr& exceptionPtr)
+{
+    m_exceptionPtr = exceptionPtr;
+}
```
jasonsandlin: remove

Source/Common/uri.h
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#pragma once
```
jplafonta: is it worth making this public? We are using casablanca uri class right now and we could move to using this instead

jasonsandlin: Maybe but its not a C API, so I'm leaning towards no

PR #105, opened 2018-02-15
Fixing WinHttp response bug

jasonsandlin:


Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -256,37 +259,36 @@ void winhttp_http_task::callback_status_data_available(
...
         // No more data available, complete the request.
         if (pRequestContext->m_responseBuffer.size() > 0)
         {
-            auto const& responseString = pRequestContext->m_responseBuffer.as_string();
-            if (responseString.length() > 0)
-            {
-                HCHttpCallResponseSetResponseString(pRequestContext->m_call, responseString.c_str());
-            }
+            char c[1] = { 0 };
+            pRequestContext->m_responseBuffer.insert(pRequestContext->m_responseBuffer.end(), &c[0], &c[0] + 1);
```
jplafonta: why not just pRequestContext->m_responseBuffer.push_back(0)

jasonsandlin: will do

PR #115, opened 2018-03-28
Updating API based on feedback

jasonsandlin:


Include/httpClient/pal.h
```diff
@@ -56,6 +75,27 @@
     #define E_ACCESSDENIED                   _HRESULT_TYPEDEF_(0x80070005L)
     #define E_ABORT                          _HRESULT_TYPEDEF_(0x80000007L)
 
+    #ifdef _In_
```
lubeltra: This seems bad, you are turning off SAL on windows...

why not:
```c++
#ifndef _In_
#define _In_
#endif
```

jasonsandlin: this is inside the !_WIN32 side of the else.  Its the same as XSAPI does today

PR #126, opened 2018-04-10
Fixing unit tests

jasonsandlin:


Source/Task/AsyncLib.cpp
```diff
@@ -2,9 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 #include "pch.h"
-#include "Async.h"
-#include "AsyncProvider.h"
-#include "AsyncQueue.h"
+#include "httpClient\async.h"
```
brianpepin: Do we want to keep these as identical as possible to what's in Windows or just "the same in spirit"?  Windows has no httpClient directory.

jasonsandlin: Oh I can probably fix it a different way then.  I think it needs that include path fix

PR #134, opened 2018-04-19
Fix NT_VERIFY and ASSERT macros not being switched on and off consistently

lubeltra:
Currently they are controlled by ENABLE_ASSERTS, but then delegate to the system assert macro which switches on a different setting leading to scenarios where we think asserts are being compiled but they are not.

Solved by removing ENABLE_ASSERTS and the NT_VERIFY macro completely

### DIFF
--- a/Source/Task/AsyncQueue.cpp
+++ b/Source/Task/AsyncQueue.cpp
@@ -439,12 +439,14 @@ static void EnsureSharedInitialization()
     BOOL pending;
 
     // Only failure modes for InitOnce are usage errors.
-    NT_VERIFY(InitOnceBeginInitialize(&s_sharedInit, 0, &pending, nullptr));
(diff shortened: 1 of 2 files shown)


Source/Task/AsyncQueue.cpp
```diff
@@ -439,12 +439,14 @@ static void EnsureSharedInitialization()
     BOOL pending;
 
     // Only failure modes for InitOnce are usage errors.
-    NT_VERIFY(InitOnceBeginInitialize(&s_sharedInit, 0, &pending, nullptr));
+    BOOL res = InitOnceBeginInitialize(&s_sharedInit, 0, &pending, nullptr);
+    assert(res);
```
matt-vanderkolk: >assert(res); [](start = 4, length = 12)

This doesn't generate an unreferenced local warning?

lubeltra: libHttpClient is built at warning level 3, and that's a lvl 4 warning

PR #136, opened 2018-04-23
Merge with changes in Windows tree

brianpepin:
This change synchronizes libHttpClient with matching code in Windows.  Same sources are now compiling in both trees.

### DIFF
--- a/Include/httpClient/async.h
+++ b/Include/httpClient/async.h
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #pragma once
 
-#include <httpClient/pal.h>
-
 /// <summary>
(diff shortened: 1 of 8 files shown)


Include/httpClient/async.h
```diff
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #pragma once
 
-#include <httpClient/pal.h>
```
lubeltra: why are we doing this?

brianpepin: Why remove pal.h?  Because we want to be able to easily merge these files between libhttpclient and windows, and windows has no "libHttpClient/pal.h".  This header gets included from other uber headers from a user's perspective so it's not mandatory that it include pal.h.  That can come from higher level headers that include async.h.

PR #139, opened 2018-04-25
Add vcxproj targeting Android and update code to support build

shanelile:
This adds an Android-specific vcxproj and updates the code to support building Android/ARM. This does not contain any concrete implementation for http calls yet, but should be the infrastructure needed for both iOS and Android versions of the library.


Include/httpClient/pal.h
```diff
@@ -12,96 +12,252 @@
...
+#endif
+
+typedef unsigned char BYTE;
+typedef unsigned long DWORD;
+typedef void *PVOID;
+typedef void VOID;
+
+typedef BYTE BOOLEAN;
+typedef BYTE *PBYTE;
+typedef PVOID HANDLE;
```
jasonsandlin: these aren't in the api guidelines

Source/Logger/trace.cpp
```diff
@@ -10,38 +10,41 @@
...
 {
-    return vsprintf_s(buffer, format, varArgs);
+    return vsnprintf(buffer, SIZE, format, varArgs);
 }
 
 void OutputDebugStringT(char const* string)
 {
-    OutputDebugStringA(string);
+    // TODO: Shane: abstract this out to a common tracing class.
+//    OutputDebugStringA(string);
```
jasonsandlin: don't kill debug spew for the rest of the platforms.  just ifdef this for now

jasonsandlin:
> 

@brianpepin, can you review these files?

---
Refers to: Source/Task/AsyncLib.cpp:3 in f74d27c. [](commit_id = f74d27c21bdb0fe0c3162b20dfed384688568fc9, deletion_comment = False)

Source/WebSocket/hcwebsocket.cpp
```diff
@@ -7,10 +7,10 @@
...
 
 
-STDAPI 
+STDAPI
 HCWebSocketCreate(
     _Out_ hc_websocket_handle_t* websocket
-    ) HC_NOEXCEPT
+) HC_NOEXCEPT
 try
 {
```
jasonsandlin: what are you doing to my pretty indents! :)

Source/Task/AsyncLib.cpp
```diff
@@ -548,35 +640,21 @@ STDAPI GetAsyncResult(
...
+            result = E_HC_FEATURE_NOT_PRESENT;
         }
         else if (buffer == nullptr)
         {
             return E_INVALIDARG;
         }
         else if (bufferSize < state->providerData.bufferSize)
         {
-            return HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER);
+            return E_HC_BUFFER_TOO_SMALL;
```
brianpepin: Are these defined to be the same values as ERROR_NOT_SUPPORT and ERROR_INSUFFFICIENT_BUFFER?  I'll need to define constants for these on the Windows side and I don't want to leak new values if we already have error codes that work.

Source/Task/AsyncQueue.cpp
```diff
@@ -245,21 +294,34 @@ class Queue
...
             {
+#if HC_PLATFORM_IS_MICROSOFT
                 ResetEvent(m_event);
+#endif
             }
         }
 
-        if (isEmpty && m_dispatchMode == AsyncQueueDispatchMode_FixedThread &&
-            GetThreadId(m_apcThread) == GetCurrentThreadId())
+        if (isEmpty && m_dispatchMode == AsyncQueueDispatchMode_FixedThread)
```
brianpepin: I think you can simplify this by making the entire block wrapped in the MS platform check.  There's no value in doing the first set of if statements and asserting for other platforms.  The whole block is just designed to pump out the user callbacks, which don't exist on other platforms.

Source/Task/AsyncQueue.cpp
```diff
@@ -245,21 +294,34 @@ class Queue
...
         }
     }
 
     bool Wait(uint32_t timeout)
     {
+#if HC_PLATFORM_IS_MICROSOFT
         uint32_t w = WaitForSingleObject(m_event, timeout);
         return (w == WAIT_OBJECT_0);
+#else
+        assert(false);
```
brianpepin: I assume this is temporary -- I imagine that being able to wait on a call to come in will be pretty common, even for a manual queue.

Source/Task/Callback_STL.h
```diff
@@ -41,19 +43,14 @@ class Callback
     {
         Clear();
     }
-    
-    // Disable copy ctor and assignment, as these cannot be implemented without 
-    // potentially throwing exceptions
-    Callback<CallbackType, CallbackDataType, CallbackThunk>(const Callback<CallbackType, CallbackDataType, CallbackThunk>&) = delete;
```
brianpepin: Default copy/assignment doesn't work in this class because there is no copy of the LIST_ENTRY memory.  You're leaving yourself open to crashes.  If you just removed because delete isn't supported on Android consider just adding these as private decls with no implementation.

Source/Task/CriticalThread.cpp
```diff
@@ -76,23 +41,12 @@ STDAPI SetTimeCriticalThread(_In_ bool isTimeCriticalThread)
...
-    ULONG_PTR timeCritical = reinterpret_cast<ULONG_PTR>(TlsGetValue(s_tlsSlot));
-
-    if ((timeCritical & CRITICAL_TRUE) == 0)
-    {
-        DWORD error = GetLastError();
-        return HRESULT_FROM_WIN32(error);
-    }
-
-    RETURN_HR(E_TIME_CRITICAL_THREAD);
+    RETURN_HR(E_HC_TIME_CRITICAL_THREAD);
```
brianpepin: I would prefer to define all these as general "windows E_" errors like E_INVALIDARG.  That way we can share values / constant names with Windows (we will use the same names).-

brianpepin:
Reviewed.  The only thing I’d request changing is to not return HC errors from async lib.  There are three error codes that were defined.  Instead, I’d define these as top level E_ error codes like you already do for things like E_INVALIDARG, and we’ll work on getting them also defined in the Windows headers.

Source/Task/Callback_STL.h
shanelile: Hmm, I'm not sure why Luca removed this, it seems to compile fine for me on Android with it. I'll add it back.

PR #142, opened 2018-05-01
Fix async block cleanup

lubeltra:
The code in asynclib.cpp was accessing AsyncBlock objects after calling calling SignalCompletion on the worker thread. This is unsafe as the completion thread may delete the AsyncBlock at any point after SignalCompletion is called.

The fix is to hold on to the AsyncState pointer if it's needed after the call to SignalCompletion and to use that instead of accessing it via the AsyncBlock

### DIFF
--- a/Source/Task/AsyncLib.cpp
+++ b/Source/Task/AsyncLib.cpp
@@ -77,13 +77,9 @@ private:
 
 struct AsyncBlockInternal
 {
-    AsyncState* state;
-    HRESULT status;
+    AsyncState* state = nullptr;
+    HRESULT status = E_PENDING;
     std::atomic_flag lock = ATOMIC_FLAG_INIT;
-
-    AsyncBlockInternal(AsyncState* s) :
-        state(s), status(E_PENDING)
-    {}
 };
 static_assert(sizeof(AsyncBlockInternal) <= sizeof(AsyncBlock::internal),
     "Unexpected size for AsyncBlockInternal");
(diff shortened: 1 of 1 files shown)


Source/Task/AsyncLib.cpp
```diff
@@ -189,7 +292,17 @@ static HRESULT AllocState(_In_ AsyncBlock* asyncBlock)
...
+    // object so we can check that AsyncBlock::Internal is all 0
+    for (auto i = 0u; i < sizeof(asyncBlock->internal); ++i)
+    {
+        if (asyncBlock->internal[i] != 0)
+        {
+            return E_INVALIDARG;
+        }
+    }
 
     HRESULT hr = AllocStateNoCompletion(asyncBlock);
```
brianpepin: If this guy fails the failed HR needs to become the status of the async block.  I would make the AsyncBlockInternal connection two phased:  early in AllocStateNoCompletion, establish AsyncBlockInternal by blitting it into the internal fields and it should have a null state pointer.  That way you have enough established that you can set the HRESULT.  Completion callbacks are going to call GetAsyncStatus and will need that HR.

Source/Task/AsyncLib.cpp
```diff
@@ -179,6 +279,9 @@ static HRESULT AllocStateNoCompletion(_In_ AsyncBlock* asyncBlock)
 #endif
     }
 
+    // construction is inherently single threaded
+    // (there is nothing we can do if the client tries to use the same
+    // AsyncBlock in 2 calls at the same time)
     new (asyncBlock->internal) AsyncBlockInternal{ state.Detach() };
```
brianpepin: I'd move this to the very beginning so we can set status on the block always with no error path.

PR #156, opened 2018-05-11
Port async unit tests and merge back with Windows tree

brianpepin:
This change adds async unit tests from the windows tree and has minor changes to cleanly merge this code back in with Windows.  Notable:

- Unified HC_PLATFORM_IS_MICROSOFT and _WIN32 -- only using the #if _WIN32 macro.
- Removed includes that use <libhttp/foo> format, since libhttp is not in windows.  I moved these to the windows PCH, which isn't merged.  Please add them to appropriate PCH if you need to share source elsewhere.


Build/libHttpClient.140.XDK.C/libHttpClient.140.XDK.C.vcxproj
```diff
@@ -14,7 +14,7 @@
     <ApplicationEnvironment>title</ApplicationEnvironment>
     <ProjectGuid>{CF37BEB2-A66A-42BA-B348-1F07566F4865}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
-    <ProjectName>libHttpClient.140.XDK.C</ProjectName>
+    <ProjectName>libHttpClient.141.XDK.C</ProjectName>
     <RootNamespace>xbox.httpclient</RootNamespace>
```
jasonsandlin: revert

Source/Task/AsyncLib.cpp
```diff
@@ -771,8 +772,22 @@ STDAPI GetAsyncResult(
         else if (token != state->token)
         {
             // Call/Result mismatch.  This AsyncBlock was initiated by state->function
-            ASSERT(false);
+            char buf[100];
+            if (state->function != nullptr)
+            {
+                sprintf_s(
```
lubeltra: sprintf_s is a Microsoft function, snprintf is the standard alternative

Include/httpClient/async.h
```diff
@@ -71,7 +65,7 @@ typedef struct AsyncBlock
     /// <summary>
     /// Internal use only
     /// </summary>
-    unsigned char internal[ASYNCBLOCK_INTERNAL_SIZE];
+    unsigned char internal[sizeof(void*) * 4];
```
lubeltra: why 4?

Include/httpClient/pal.h
```diff
@@ -37,9 +37,18 @@
...
+#ifndef E_TIME_CRITICAL_THREAD
+#define E_TIME_CRITICAL_THREAD           _HRESULTYPEDEF_(0x800701A0L)
+#endif
+
+#ifndef E_NOT_SUPPORTED
 #define E_NOT_SUPPORTED                  _HRESULTYPEDEF_(0x80070032L)
-#define E_INSUFFICIENT_BUFFER            _HRESULTYPEDEF_(0x8007007AL)
+#endif
+
+#ifndef E_NOT_SUFFICIENT_BUFFER
```
lubeltra: Update the non win32 section, right now it still defines E_INSUFFICIENT_BUFFER

Where does this name come from anyway? http://errors/ does not find it.

Include/httpClient/async.h
brianpepin: I'm a sucker for powers of two.  We won't be able to change this header once published and I wanted a bit of headroom.

Source/Task/AsyncLib.cpp
brianpepin: Fixed.

Include/httpClient/pal.h
brianpepin: It's defined in winerror.h. I found it already defined as E_NOT_SUFFICIENT_BUFFER, which is why the rename. Updated code.

PR #158, opened 2018-05-14
Shanel/android impl

shanelile:
Refactors tracing to include platform specific definitions and fixes some build warnings and issues with the Android project.

### DIFF
--- a/Source/Logger/Android/android_logger.cpp
+++ b/Source/Logger/Android/android_logger.cpp
@@ -1,0 +1,31 @@
+#include <Android/log.h>
+#include <httpClient/pal.h>
+#include <httpClient/trace.h>
+
+void Internal_HCTraceMessage(const char* areaName, HCTraceLevel traceLevel, const char* message)
+{
+    int androidLogPriority = ANDROID_LOG_UNKNOWN;
+
+    switch (traceLevel) {
+        case HCTraceLevel_Off:
+            androidLogPriority = ANDROID_LOG_SILENT;
+            break;
+        case HCTraceLevel_Error:
+            androidLogPriority = ANDROID_LOG_ERROR;
+            break;
+        case HCTraceLevel_Warning:
+            androidLogPriority = ANDROID_LOG_WARN;
+            break;
+        case HCTraceLevel_Important:
+            androidLogPriority = ANDROID_LOG_WARN;
+            break;
+        case HCTraceLevel_Information:
+            androidLogPriority = ANDROID_LOG_INFO;
+            break;
+        case HCTraceLevel_Verbose:
+            androidLogPriority = ANDROID_LOG_VERBOSE;
+            break;
+    }
+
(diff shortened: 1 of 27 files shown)


Source/Logger/trace.cpp
```diff
@@ -156,7 +147,7 @@ void TraceMessageToDebugger(
 
     char outputBuffer[BUFFER_SIZE] = {};
     // [threadId][level][time][area] message
-    auto written = stprintf_s(outputBuffer, "[%04X][%s][%02d:%02d:%02d.%03u][%s] %s",
+    auto written = stprintf_s(outputBuffer, "[%04" PRIX64 "][%s][%02d:%02d:%02d.%03u][%s] %s",
```
lubeltra: have you tried `llX`?
I would like to avoid using the macro type specifiers if possible

Source/Logger/iOS/ios_logger.cpp
```diff
@@ -0,0 +1,8 @@
+#include "pch.h"
+
+#include <httpClient/trace.h>
+
+void Internal_HCTraceMessage(const char* areaName, HCTraceLevel traceLevel, const char* message)
```
lubeltra: We should discuss if we want to move all the debug log formatting in this function as a future bit of work

Source/Logger/trace.cpp
shanelile: llX works, I can change it to use that.

Source/Logger/Android/android_logger.cpp
```diff
@@ -0,0 +1,31 @@
+#include <Android/log.h>
```
matt-vanderkolk: >include [](start = 1, length = 7)

include pch

Source/Logger/Android/android_logger.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include <httpClient/pal.h>
+#include <httpClient/trace.h>
+
+void Internal_HCTraceMessage(const char* areaName, HCTraceLevel traceLevel, const char* message)
+{
+    int androidLogPriority = ANDROID_LOG_UNKNOWN;
+
+    switch (traceLevel) {
```
matt-vanderkolk: > { [](start = 23, length = 2)

new line?

Utilities/CMake/CMakeLists.txt
```diff
@@ -191,6 +191,13 @@ set(Logger_Source_Files
     ../../../Source/Logger/log_publics.cpp
     )
 
+set(Windows_Logger_Source_Files
+   ../../../Source/Logger/Win/win_logger.cpp
+   )
+
+set(Android_Logger_Source_Files
+	../../../Source/Logger/Android/android_logger.cpp
+	)
```
matt-vanderkolk: replace tabs with spaces

PR #160, opened 2018-05-15
Add null check to nested queue

rakhad:


Source/Task/AsyncQueue.cpp
```diff
@@ -534,6 +534,12 @@ STDAPI CreateNestedAsyncQueue(
     _In_ async_queue_handle_t parentQueue,
     _Out_ async_queue_handle_t* queue)
 {
+    if (parentQueue == nullptr)
+    {
+        *queue = nullptr;
+        return S_OK;
+    }
```
jplafonta: Should this just be in the if check below? Instead of return E_INVALIDARG

brianpepin: ? This looks like an error.  Why would I legitimately call this with a null parent?  I'd be tempted to crash.  Note that if you call with the queue field in providerData, that is never null (we put a default version in there if null was passed into the async block).

brianpepin:
Sweet!

On May 15, 2018, at 6:28 PM, Ramsey <notifications@github.com<mailto:notifications@github.com>> wrote:


Closed #160<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FMicrosoft%2FlibHttpClient%2Fpull%2F160&data=02%7C01%7Cbrianpe%40microsoft.com%7Cd65c824feab749c3800308d5bacc5343%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636620309084549960&sdata=hGt6FQYcd5alLdPV8jTil9s%2F9gyPjQOAm70LLT5MBgk%3D&reserved=0>.

PR #164, opened 2018-05-17
Signal wait after completion callback and enable non-win32 waits

brianpepin:
This change does the following:

- Changes the wait event semantics so the event is signaled only after the completion has run.  This makes it trivial to turn an async call into a sync call.  The wait is also alterable now.

- We now use a condition variable for non-win32 waits.  Unfortunately condition variables don't support alertable states so I needed both code paths.

- Fixed errors -- we now define as constants by undef/redef on down level win32 compilers.  This is unnecessary on VS 2017.  Also fixed the numeric error values for E_TIME_CRITICAL_THREAD.


Build/libHttpClient.UnitTest.140.TE/libHttpClient.UnitTest.140.TE.vcxproj.filters
```diff
@@ -193,9 +193,12 @@
     </Filter>
     <Filter Include="C++ Source\Logger">
       <UniqueIdentifier>{4FAC39B2-DCB4-3476-AEEF-548A5C43612D}</UniqueIdentifier>
+<<<<<<< HEAD
+=======
     </Filter>
     <Filter Include="C++ Source\Logger\Win">
       <UniqueIdentifier>{11DB47ED-9A1E-3493-8CA8-5F3BE24F13F1}</UniqueIdentifier>
+>>>>>>> master
     </Filter>
```
jasonsandlin: revert

Include/httpClient/async.h
```diff
@@ -48,7 +48,9 @@ typedef struct AsyncBlock
     async_queue_handle_t queue;
 
     /// <summary>
-    /// Optional event to wait on
+    /// Optional event to wait on.  This will be signale
```
jasonsandlin: >signale [](start = 49, length = 7)

signal

PR #165, opened 2018-05-22
Initial Android HTTP implementation

shanelile:
This is the initial stab at a basic Android HTTP implemention for libHttpClient. It includes a new androidproj with a simple wrapper around OkHttp and enough entry points to manage extracting the data we need from hc_call and then populating the response. I've verified that basic calls work by integrating into our test XAL app and getting far enough to request a D token. This raises some questions but I felt the context of the pull request would make it easier to understand the motivations behind them and I'll add them inline.


Source/HTTP/Android/android_http_request.cpp
```diff
@@ -0,0 +1,355 @@
+#include "pch.h"
+
+#include "android_http_request.h"
+
+#include <httpClient/httpClient.h>
+
+/* static */ JavaVM* HttpRequest::s_javaVM = nullptr;
```
jasonsandlin: >/* static */ [](start = 0, length = 12)

why /* static */ everywhere?  seems like you should just remove this

Include/httpClient/httpClient.h
```diff
@@ -100,7 +100,7 @@ STDAPI HCMemGetFunctions(
 /// Should have a corresponding call to HCGlobalCleanup().
 /// </summary>
```
lubeltra: ></summary> [](start = 4, length = 10)

param comment

Source/HTTP/Android/http_android.cpp
```diff
@@ -3,13 +3,77 @@
 #include "pch.h"
 #if ANDROID_API
 #include <httpClient/httpClient.h>
+#include "android_http_request.h"
+
+void Internal_HCHttpCallInitialize(void* context)
+{
+    HttpRequest::InitializeJavaEnvironment(reinterpret_cast<JavaVM*>(context));
```
lubeltra: >HttpRequest::InitializeJavaEnvironment(reinterpret_cast<JavaVM*>(context)); [](start = 4, length = 75)

should only do this if `HC_PLATFORM == HC_PLATFORM_ANDROID` and assert context is not null on other platforms you should `assert(context == nullptr);`

Source/HTTP/Android/http_android.cpp
```diff
@@ -3,13 +3,77 @@
 #include "pch.h"
 #if ANDROID_API
 #include <httpClient/httpClient.h>
+#include "android_http_request.h"
```
lubeltra: should only do this if HC_PLATFORM == HC_PLATFORM_ANDROID

Source/HTTP/Android/android_http_request.cpp
lubeltra: as a reminder these are static, since it's not part of the definition, only the declaration

---
In reply to: [189750654](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r189750654) [](ancestors = 189750654)

Source/HTTP/Android/android_http_request.cpp
```diff
@@ -0,0 +1,355 @@
...
+/* static */ jclass HttpRequest::s_httpRequestClass = nullptr;
+/* static */ jclass HttpRequest::s_httpResponseClass = nullptr;
+
+/* static */ HRESULT HttpRequest::InitializeJavaEnvironment(JavaVM* javaVM) 
+{
+    s_javaVM = javaVM;
```
lubeltra: >javaVM [](start = 15, length = 6)

assert non null

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -823,6 +823,10 @@ void winhttp_http_task::perform_async()
 
 NAMESPACE_XBOX_HTTP_CLIENT_END
 
+void Internal_HCHttpCallInitialize(void* context)
```
lubeltra: >context [](start = 41, length = 7)

assert is null

lubeltra: does the ios version need updating too?

---
In reply to: [189752159](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r189752159) [](ancestors = 189752159)

Source/HTTP/Android/http_android.cpp
```diff
@@ -3,13 +3,77 @@
...
+        const char* contentType = nullptr;
+        HCHttpCallRequestGetHeader(call, "Content-Type", &contentType);
+        httpRequest.SetMethodAndBody(requestMethod, contentType, requestBody, requestBodySize);
+    }
+
+    result = httpRequest.ExecuteRequest();
+
+    if (!SUCCEEDED(result)) 
+    { 
+        CompleteAsync(asyncBlock, result, 0);
```
lubeltra: >CompleteAsync(asyncBlock, result, 0); [](start = 8, length = 37)

can we set a more detailed error code here?

Source/HTTP/Android/android_http_request.cpp
```diff
@@ -0,0 +1,355 @@
...
+        jint attachThreadResult = s_javaVM->AttachCurrentThread(&jniEnv, nullptr);
+
+        if (attachThreadResult == JNI_OK) 
+        {
+            isThreadAttached = true;
+        }
+        else 
+        {
+            HC_TRACE_ERROR(HTTPCLIENT, "Could not attach to java thread to dispose of global class references");
+        }
```
lubeltra: I would just give an error here

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
```diff
@@ -0,0 +1,51 @@
...
+    public void setHttpUrl(String url) {
+        this.requestBuilder = this.requestBuilder.url(url);
+    }
+
+    public void setHttpMethodAndBody(String method, String contentType, byte[] body) {
+    	if ("GET".equals(method)) {
+            this.requestBuilder = this.requestBuilder.method(method, null);
+    	} else {
+            this.requestBuilder = this.requestBuilder.method(method, RequestBody.create(MediaType.parse(contentType), body));
+    	}
```
lubeltra: tabs mixed in

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
```diff
@@ -0,0 +1,51 @@
...
+        return new HttpClientRequest();
+    }
+
+    public void setHttpUrl(String url) {
+        this.requestBuilder = this.requestBuilder.url(url);
+    }
+
+    public void setHttpMethodAndBody(String method, String contentType, byte[] body) {
+    	if ("GET".equals(method)) {
+            this.requestBuilder = this.requestBuilder.method(method, null);
```
lubeltra: do we need this special case? there are many methods that don't take a body

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
```diff
@@ -0,0 +1,51 @@
...
+    public HttpClientResponse doRequest() {
+        OkHttpClient client = new OkHttpClient.Builder().build();
+
+        try {
+            Response response = client.newCall(this.requestBuilder.build()).execute();
+            return new HttpClientResponse(response); 
+        } catch (IOException e) {
+            Log.e("HttpRequestClient", "Failed to execute request", e);
+            return null;
+        }
```
lubeltra: is there a way to make this non blocking?
maybe not now but in the future it would be good

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientResponse.java
```diff
@@ -0,0 +1,61 @@
...
+package com.xbox.httpclient;
+
+import android.util.Log;
+
+import java.io.IOException;
+import okhttp3.Response;
+
+class HttpClientResponse
+{
+    private Response okHttpResponse;
```
lubeltra: >okHttpResponse [](start = 21, length = 14)

why is it called okHttpResponse if it can be unsuccessful?

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientResponse.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    public String getResponseString() {
+        try {
+            return this.okHttpResponse.body().string();
+        } catch (IOException e) {
+            return null;
+        }
+    }
```
lubeltra: is this used anywhere?

Source/HTTP/Android/android_http_request.h
```diff
@@ -0,0 +1,36 @@
...
+
+    // Response Functions
+    uint32_t GetResponseCode();
+    uint32_t GetResponseHeaderCount();
+    std::string GetHeaderNameAtIndex(uint32_t index);
+    std::string GetHeaderValueAtIndex(uint32_t index);
+    HRESULT ProcessResponseBody(hc_call_handle_t call);
+
+    static HRESULT InitializeJavaEnvironment(JavaVM* javaVM);
+    static void CleanupJavaEnvironment();
```
lubeltra: >CleanupJavaEnvironment [](start = 16, length = 22)

who calls this?

Source/HTTP/Android/http_android.cpp
```diff
@@ -3,13 +3,77 @@
...
 #if ANDROID_API
 #include <httpClient/httpClient.h>
+#include "android_http_request.h"
+
+void Internal_HCHttpCallInitialize(void* context)
+{
+    HttpRequest::InitializeJavaEnvironment(reinterpret_cast<JavaVM*>(context));
+}
 
 void Internal_HCHttpCallPerform(
```
lubeltra: >Internal_HCHttpCallPerform [](start = 5, length = 26)

I think we want to associate a context with the perform function, so that we don't need to use statics for the JavaVM and associated pointers

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
shanelile: Yes, there are non-blocking versions of the call. We'll need to pass a pointer to the native httprequest object so we can call it back on completion.

Discussed offline, I'll open a bug and we can address it in a subsequent update.

---
In reply to: [189760459](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r189760459) [](ancestors = 189760459)

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientResponse.java
lubeltra: That's the worst name for a namespace I've ever seen...

---
In reply to: [190003661](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r190003661) [](ancestors = 190003661,189760525)

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientResponse.java
shanelile: Nope, I've removed it.

---
In reply to: [189760666](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r189760666) [](ancestors = 189760666)

Source/HTTP/Android/http_android.cpp
shanelile: I'm a little confused, should this comment be on ExecuteRequest? We set whatever error code that returns.

That said, it's is returning E_FAIL right now on failure. I think when we look into using a non-blocking call for the HttpRequest we can use that as an opportunity to flesh out some specific error types we'd want call in the async callback.

---
In reply to: [189752347](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r189752347) [](ancestors = 189752347)

lubeltra: We can set the platform specific error code on the http request using HCHttpCallResponseSetNetworkErrorCode

---
In reply to: [190067507](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r190067507) [](ancestors = 190067507,189752347)

shanelile: Ah, adding a call to 
HCHttpCallResponseSetNetworkErrorCode(call, result, static_cast<uint32_t>(result)); 
above it.

---
In reply to: [190071569](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r190071569) [](ancestors = 190071569,190067507,189752347)

lubeltra:
> STDAPI_(void) HCGlobalSetHttpCallPerformFunction(

This should take a context pointer

---
Refers to: Include/httpClient/httpProvider.h:32 in 8003f92. [](commit_id = 8003f92b805c51779508463308efacd2a8d88e48, deletion_comment = False)

lubeltra:
> STDAPI HCGlobalGetHttpCallPerformFunction(

This should return the current context pointer

---
Refers to: Include/httpClient/httpProvider.h:44 in 8003f92. [](commit_id = 8003f92b805c51779508463308efacd2a8d88e48, deletion_comment = False)

Source/HTTP/httpcall.cpp
```diff
@@ -123,7 +123,7 @@ HRESULT perform_http_call(
                     {
                         try
                         {
-                            performFunc(data->async, call);
+                            performFunc(data->async, call, httpSingleton->m_platformContext.get());
```
lubeltra: >httpSingleton->m_platformContext.get() [](start = 59, length = 38)

since you aren't allowing client to override the context, anybody that has set a custom perform function will be given a "random" pointer here they don't know what to do with

Source/Global/global.h
```diff
@@ -55,6 +55,9 @@ typedef struct http_singleton
     uint32_t m_timeoutWindowInSeconds;
     uint32_t m_retryDelayInSeconds;
 
+    // Platform-specific context for calls
+    std::shared_ptr<HCPlatformContext> m_platformContext;
```
lubeltra: You can't own the context the clients will pass in.
(We may need to track both the current perform context and the "default" context separately for cleanup.

I'm not sure why you are using a shared pointer either since you never have more than one copy

Source/HTTP/Android/android_platform_context.cpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    jclass globalRequestClass = reinterpret_cast<jclass>(jniEnv->NewGlobalRef(localHttpRequest));
+    jclass globalResponseClass = reinterpret_cast<jclass>(jniEnv->NewGlobalRef(localHttpResponse));
+
+    *platformContext = new HCPlatformContext(javaVm, globalRequestClass, globalResponseClass);
+    return S_OK;
+}
+
+HCPlatformContext::HCPlatformContext(JavaVM* javaVm, jclass requestClass, jclass responseClass) :
```
lubeltra: >HCPlatformContext [](start = 0, length = 17)

where's the class definition?

Include/httpClient/pal.h
```diff
@@ -11,6 +11,10 @@
 
 #include <httpClient/config.h>
 
+#if HC_PLATFORM == HC_PLATFORM_ANDROID
+#include "jni.h"
+#endif
```
lubeltra: why do you need this?

Source/HTTP/Android/android_platform_context.cpp
shanelile: pal.h

---
In reply to: [190331089](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r190331089) [](ancestors = 190331089)

Source/Global/global.h
shanelile: Exposing the context to clients feels really wrong here. I think the http call has access to the singleton struct and it probably makes more sense to reference it directly.

---
In reply to: [190330199](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r190330199) [](ancestors = 190330199)

Include/httpClient/pal.h
```diff
@@ -263,6 +267,35 @@ typedef struct _LIST_ENTRY {
...
+#else
+
+struct HCPlatformContext 
+{ 
+    // The compiler does not like this class being completely empty,
+    // so define an empty context pointer.
+    void* emptyContext;
+};
+
+#endif
```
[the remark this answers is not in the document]
shanelile: Won't fixing a bunch of comments that are obsolete

---
In reply to: [190332825](https://github.com/Microsoft/libHttpClient/pull/165#discussion_r190332825) [](ancestors = 190332825)

PR #167, opened 2018-05-22
Adding Async at end of async API functions

jasonsandlin:


matt-vanderkolk:
Do iOS files need to be updated as well?

jasonsandlin:
I updated http_ios.mm.  Maybe I missed some?  I'll check again

---
In reply to: [391117219](https://github.com/Microsoft/libHttpClient/pull/167#issuecomment-391117219) [](ancestors = 391117219)

jasonsandlin:
looks fine to me.  I updated ios_websocket.cpp already

---
In reply to: [391117620](https://github.com/Microsoft/libHttpClient/pull/167#issuecomment-391117620) [](ancestors = 391117620,391117219)

PR #176, opened 2018-05-30
Replace existing HTTP execute call with a non-blocking call

shanelile:
Currently the Java implementation is using a blocking call for executing the underlying HTTP request. This updates the call to use the asynchronous version and refactors the C++ side to handle setting the call context and retrieving it when the call has completed.

### DIFF
--- a/Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
+++ b/Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
@@ -42,8 +42,8 @@ public class HttpClientRequest {
 
     public void doRequestAsync(final long sourceCall) {
         OkHttpClient client = new OkHttpClient.Builder()
-        		.retryOnConnectionFailure(false) // Explicitly disable retries; retry logic will be managed by native code in libHttpClient
-        		.build();
+                .retryOnConnectionFailure(false) // Explicitly disable retries; retry logic will be managed by native code in libHttpClient
+                .build();
         client.newCall(this.requestBuilder.build()).enqueue(new Callback() {
             @Override
             public void onFailure(final Call call, IOException e) {
--- a/Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
+++ b/Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
@@ -4,6 +4,8 @@ import android.util.Log;
 
 import java.io.IOException;
 
+import okhttp3.Call;
+import okhttp3.Callback;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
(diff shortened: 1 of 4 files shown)


Source/HTTP/Android/http_android.cpp
```diff
@@ -6,15 +6,40 @@
...
+    {
+        CompleteAsync(sourceRequest->GetAsyncBlock(), E_FAIL, 0);
+    }
+    else 
+    {
+        HRESULT result = sourceRequest->ProcessResponse(sourceCall, response);
+        CompleteAsync(sourceRequest->GetAsyncBlock(), result, 0);
+    }
+
+    delete sourceRequest;
```
shanelile: >delete sourceRequest; [](start = 4, length = 21)

I'd like some advice for what to do about this; on windows it looks like there's a TODO to clean up the windows-specific type that's newed up. I could take ownership as a unique_ptr here and let that clean it up when the function goes out of scope if that's preferred.

I think I need to clean this up in the failure condition below as well.

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
```diff
@@ -38,14 +40,23 @@ public void setHttpHeader(String name, String value) {
         this.requestBuilder = requestBuilder.addHeader(name, value);
     }
 
-    public HttpClientResponse doRequest() {
-        OkHttpClient client = new OkHttpClient.Builder().build();
+    public void doRequestAsync(final long sourceCall) {
+        OkHttpClient client = new OkHttpClient.Builder()
+        		.retryOnConnectionFailure(false) // Explicitly disable retries; retry logic will be managed by native code in libHttpClient
```
lubeltra: >		 [](start = 8, length = 2)

some stray tabs in this file

Source/HTTP/Android/http_android.cpp
lubeltra: using a unique_ptr would be good to protect against weird edge cases

---
In reply to: [191911754](https://github.com/Microsoft/libHttpClient/pull/176#discussion_r191911754) [](ancestors = 191911754)

Source/HTTP/Android/http_android.cpp
```diff
@@ -6,15 +6,40 @@
...
 void Internal_HCHttpCallPerformAsync(
     _In_ AsyncBlock* asyncBlock,
     _In_ hc_call_handle_t call
     )
 {
     auto httpSingleton = xbox::httpclient::get_http_singleton(true);
     AndroidPlatformContext* platformContext = reinterpret_cast<AndroidPlatformContext*>(httpSingleton->m_platformContext.get());
-    HttpRequest httpRequest(platformContext->GetJavaVm(), platformContext->GetHttpRequestClass(), platformContext->GetHttpResponseClass());
-    HRESULT result = httpRequest.Initialize();
+    HttpRequest* httpRequest = new HttpRequest(asyncBlock, platformContext->GetJavaVm(), platformContext->GetHttpRequestClass(), platformContext->GetHttpResponseClass());
```
lubeltra: >HttpRequest* httpRequest = new HttpRequest( [](start = 4, length = 43)

use a unique_ptr here

Source/HTTP/Android/http_android.cpp
```diff
@@ -51,29 +76,16 @@ void Internal_HCHttpCallPerformAsync(
...
     }
 
-    httpRequest.SetMethodAndBody(requestMethod, contentType, requestBody, requestBodySize);
+    httpRequest->SetMethodAndBody(requestMethod, contentType, requestBody, requestBodySize);
 
-    result = httpRequest.ExecuteRequest();
+    HCHttpCallSetContext(call, httpRequest);
+    result = httpRequest->ExecuteAsync(call);
 
     if (!SUCCEEDED(result))
```
lubeltra: >if (!SUCCEEDED(result))  [](start = 4, length = 24)

in the SUCCEEDED branch here release the unique_ptr

PR #185, opened 2018-06-06
Updating name of duplicate async queue handle

jasonsandlin:


Source/Task/AsyncLib.cpp
```diff
@@ -287,7 +287,7 @@ static HRESULT AllocStateNoCompletion(_Inout_ AsyncBlock* asyncBlock, _Inout_ As
 
     if (state->providerData.queue != nullptr)
     {
-        ReferenceAsyncQueue(state->providerData.queue);
+        DuplicateAsyncQueueHandle(state->providerData.queue);
```
brianpepin: For correctness you probably want state->providerData.queue = DuplicateAsyncQueueHandle(...) (even though it returns the input param)

Source/Task/AsyncQueue.cpp
```diff
@@ -590,7 +590,7 @@ STDAPI CreateSharedAsyncQueue(
 
     if (q != nullptr)
     {
-        ReferenceAsyncQueue(q);
+        DuplicateAsyncQueueHandle(q);
```
brianpepin: Same here:  q = DuplicateAsyncQueueHandle(q);

PR #186, opened 2018-06-07
Removing async wait handle

jasonsandlin:


Include/httpClient/async.h
```diff
@@ -64,6 +57,11 @@ typedef struct AsyncBlock
     /// </summary>
     AsyncCompletionRoutine* callback;
 
+    /// <summary>
+    /// Internal use only
+    /// </summary>
+    HANDLE internalHandle;
```
brianpepin: Not necessary.  We create one internally and can just use that.

Source/Task/AsyncLib.cpp
```diff
@@ -438,10 +438,10 @@ static void CALLBACK CompletionCallbackForAsyncBlock(
     {
         asyncBlock->callback(asyncBlock);
     }
-    if (asyncBlock->waitEvent)
+    if (asyncBlock->internalHandle)
```
brianpepin: Woah. I think this is all wrong (existing code).  Note that when we schedule the callback we pass in a state object, not an async block. This should attach to a state pointer, invoke with the state pointer's async block and then set the event in the state pointer.

Include/httpClient/async.h
lubeltra: Agreed

---
In reply to: [193618417](https://github.com/Microsoft/libHttpClient/pull/186#discussion_r193618417) [](ancestors = 193618417)

lubeltra:
>     }

This code could just be removed and state can create the event internally

---
Refers to: Source/Task/AsyncLib.cpp:283 in 9d6b317. [](commit_id = 9d6b317ae2ad9ad2f4138a70293bf40c95422f10, deletion_comment = False)

Source/Task/AsyncLib.cpp
lubeltra:
```c++
            hr = SubmitAsyncCallback(
                asyncBlock->queue,
                AsyncQueueCallbackType_Completion,
                asyncBlock,
                CompletionCallbackForAsyncBlock);
```
context is the AsyncBlock

---
In reply to: [193618839](https://github.com/Microsoft/libHttpClient/pull/186#discussion_r193618839) [](ancestors = 193618839)

PR #226, opened 2018-08-03
Allow optional import of json implementation from json.h

najadojo:
Update to latest UWP SDK
Remove use of U macro in json library code
Allow projects to load without nuget packages present
Correct build dependency issues.

### DIFF
--- a/Include/json_cpp/json.h
+++ b/Include/json_cpp/json.h
@@ -1921,9 +1921,11 @@ public:
     _ASYNCRTIMP utility::istream_t& __cdecl operator >> (utility::istream_t &is, json::value &val);
 }}
 
+#ifndef NO_JSON_CPP_IMPL
 #include "details/asyncrt_utils.hpp"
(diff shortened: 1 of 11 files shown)


Include/json_cpp/json.h
```diff
@@ -1921,9 +1921,11 @@ namespace json
     _ASYNCRTIMP utility::istream_t& __cdecl operator >> (utility::istream_t &is, json::value &val);
 }}
 
+#ifndef NO_JSON_CPP_IMPL
```
jasonsandlin: The intention is that the C++ JSON header is just provided for customers who want a JSON parser, but not part of libHttpClient's header set.  Its basically an optional component.   So if you don't want the JSON impl, then simply don't include json.h

najadojo: I do want to use json.h but I'd like to include this header in multiple complication units. Without this (and another external source to do these includes) I get a linker error that multiple definitions are found.

PR #229, opened 2018-08-03
Add Api to get request URL from HttpCall handle

jplafonta:


Source/HTTP/httpcall.cpp
```diff
@@ -538,6 +538,28 @@ try
...
+	}
+
+	if (call->url.empty())
+	{
+		*url = nullptr;
+	}
+
+	*url = call->url.data();
+	return S_OK;
+}
```
lubeltra: tabs

Source/HTTP/httpcall.cpp
```diff
@@ -538,6 +538,28 @@ try
...
+	_Out_ const char** url
+	) HC_NOEXCEPT
+try
+{
+	if (call == nullptr)
+	{
+		return E_INVALIDARG;
+	}
+
+	if (call->url.empty())
```
lubeltra: data is safe to call even on empty strings, it will return a pointer to an empty c string (so just a null terminator) which may be easier to consume for the client (std::string does not play nice with null pointers)

jplafonta: oh ok ill just return empty string then

PR #240, opened 2018-08-08
NSTimer needs a running RunLoop; 

najadojo:
Check if the current thread's is running if not run the RunLoop until the timer has a chance to fire.


najadojo:
@edingwell any additional information I can provide?

PR #245, opened 2018-08-14
wspp should not be used for unit test projects even though they are win32

jplafonta:
…in32


Build/libHttpClient.internal.props
```diff
@@ -1,16 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(ConfigurationType)' == 'StaticLibrary'">
```
jplafonta: >Condition="'$(ConfigurationType)' == 'StaticLibrary'" [](start = 17, length = 53)

This is actually incorrect.  Remove this condition.

PR #248, opened 2018-08-15
Make libhttpclient.props and openssl.props provide MSBuild references to their projects

lubeltra:


Build/openssl.props
```diff
@@ -1,35 +1,60 @@
...
+>
+<!--Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003"-->
   <PropertyGroup>
-    <HCSettingsFile>$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), hc_settings.props))\hc_settings.props</HCSettingsFile>
+    <HCBuildRoot Condition="'$(HCBuildRoot)'==''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.root))\</HCBuildRoot>
+    <libcryptoName>libcrypto.$(HCLibPlatformType)</libcryptoName>
+    <libsslName>libssl.$(HCLibPlatformType)</libsslName>
+    <opensslSourceRoot>$(HCBuildRoot)External\openssl\</opensslSourceRoot>
+    <opensslGeneratedHeaders>$(HCBuildRoot)External\generatedHeaders\$(HCLibPlatformType)\</opensslGeneratedHeaders>
+    <doref>lol nope</doref>
```
jplafonta: Cant you just leave this undefinded?

Build/openssl.props
```diff
@@ -1,35 +1,60 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup Label="PlatformType">
-    <platformType Condition="'$(ApplicationType)' == 'Android'">Android</platformType>
-    <platformType Condition="'$(ApplicationType)' == ''">Win32</platformType>
-  </PropertyGroup>
```
jplafonta: What was this issue with having this here as opposed to in the project files?

libHttpClient.props
```diff
@@ -1,66 +1,101 @@
...
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(ApplicationType)'=='' AND '$(Platform)'=='Durango'">
-    <HCLibPlatformType>XDK</HCLibPlatformType>
-    <libHttpClientDefine>HC_XDK_API</libHttpClientDefine>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(ApplicationType)'=='Android'">
-    <HCLibPlatformType>Android</HCLibPlatformType>
-    <libHttpClientDefine>HC_ANDROID_API</libHttpClientDefine>
-    <useWsppWebsockets>true</useWsppWebsockets>
-  </PropertyGroup>
```
jplafonta: Why is this removed?  We are now relying on clients to define HCLibPlatformType which seems worse that inferring it here.

Build/openssl.props
lubeltra: Ugh, testing code that shouldn't have been checked in

libHttpClient.props
lubeltra: For both of these comments I want to be sure the settings are consistent across the board (openssl, libhttpclient, xal, xsapi all have to agree). Right now manually setting the values in each project guarantees that.

In a subsequent change, we can create a shared props file that does the guessing, and all projects use it.

PR #257, opened 2018-08-23
Fixed how HC_ANDROID_API is defined

rakhad:


jplafonta:
> #endif

Are theses also wrong then?

---
Refers to: Include/httpClient/pal.h:32 in 77b15d9. [](commit_id = 77b15d966ca2c6fece064427c330d541eee4d5d6, deletion_comment = False)

Include/httpClient/pal.h
```diff
@@ -60,8 +60,8 @@
 
 #else
 
-#ifndef HC_ANDROID_API
-#define HC_ANDROID_API (HC_PLATFORM == HC_PLATFORM_ANDROID)
+#if !defined(HC_ANDROID_API) && HC_PLATFORM == HC_PLATFORM_ANDROID
+#define HC_ANDROID_API
 #endif
```
jplafonta: Perhaps just change how this is used? instead of #ifdef HC_ANDROID_API try #if HC_ANDROID_API?

rakhad: I just looked at how HC_UWP_API worked an it works this way. I'll change to align.

---
In reply to: [212404190](https://github.com/Microsoft/libHttpClient/pull/257#discussion_r212404190) [](ancestors = 212404190)

PR #258, opened 2018-08-28
Added http status codes to pal.h

rakhad:


Include/httpClient/pal.h
```diff
@@ -132,13 +132,39 @@ typedef void* HANDLE;
 #define E_NOT_SUFFICIENT_BUFFER                 _HRESULTYPEDEF_(0x8007007AL)
 #define E_NOINTERFACE                           _HRESULTYPEDEF_(0x80004002L)
 #define E_BOUNDS                                _HRESULTYPEDEF_(0x8000000BL)
+#define HTTP_E_STATUS_AMBIGUOUS                 _HRESULT_TYPEDEF_(0x8019012CL)
```
jplafonta: nit: _HRESULTYPEDEF_ just to be consistent? not sure why we have both defined identically above

PR #266, opened 2018-09-04
Updating reference to 'config.h' to have a proper public header path

matt-vanderkolk:


najadojo:
👍 Is there a CI system in plans so these breaks can be caught before merging?

jasonsandlin:
We have a gated check-in for our internal VSO builds, but we're having trouble making the packageES VSO build system do a gated build for our GitHub projects.  Haven't dug into cause yet.

PR #276, opened 2018-09-11
Merge changes from Windows tree into libHttpClient

brianpepin:
This PR merges the latest changes from the Windows tree into GitHub. Summary of changes:

- Scheduling a call into the future has moved from AsyncLib to AsyncQueue, so it's available as a mechanism to do periodic events.  The timer code here was reworked to be more efficient for Win32 and simpler.
- An async queue handle is now a required parameter to any async call and needs to be set into the async block or you will get an invalid arg error.  This was done to reduce accidents where null is used, bypassing a queue for your app/game.
- You can now "BeginAsyncAlloc", which begins an async call and allocates context data in a single allocation, which reduces the need for several small allocations.


Source/Task/AsyncLib.cpp
```diff
@@ -32,8 +31,17 @@ struct AsyncState
...
-        timer{ this, TimerCallback }
+    void* operator new(size_t size, size_t additional)
+    {
+        return malloc(size + additional);
+    }
+
+    void operator delete(void* ptr)
+    {
+        free(ptr);
+    }
```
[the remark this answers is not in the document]
lubeltra: Oh I see, couldn't you forward `operator new(size_t, size_t)` to `operator new(size_t)`? Just in case someone else overrides the global `operator new`

Source/Task/AsyncLib.cpp
```diff
@@ -217,80 +250,36 @@ static void CALLBACK CompletionCallbackForAsyncBlock(_In_ void* context);
...
-    {
-        state.Detach(); // state is still in use so let it go
-    }
-    else
+    if (contextSize != 0)
     {
-        CompleteAsync(state->providerData.async, hr, 0);
+        // User allocated additional context data.  This was allocated as extra bytes at the end of 
+        // async state.
+        state->providerData.context = (state.Get() + 1);
```
lubeltra: should we 0 that memory?

Source/Task/AsyncQueue.cpp
```diff
@@ -296,30 +350,31 @@ class Queue
...
-        uint32_t w = WaitForSingleObject(m_event, timeout);
-        return (w == WAIT_OBJECT_0);
-#else
-        ASSERT(false);
-        return true;
-#endif
+        std::unique_lock<std::mutex> lock(m_cs);
+        if (!m_hasItems)
+        {
+            m_event.wait_for(lock, std::chrono::milliseconds(timeout));
```
lubeltra: are we worried about wait returning early? Condition variables can be spuriously woken by the os

Source/Task/WaitTimer_stl.cpp
```diff
@@ -0,0 +1,136 @@
...
+        {
+            std::lock_guard<std::mutex> lock(m_mutex);
+            m_abortThread = true;
+            m_cv.notify_all();
+        }
+
+        m_thread.join();
+    }
+
+    HRESULT Initialize(_In_opt_ void* context, _In_ WaitTimerCallback* callback)
```
lubeltra: we are spawning a thread per timer now, the old code only ever created a single thread

Source/Task/WaitTimer_stl.cpp
```diff
@@ -0,0 +1,136 @@
...
+        std::lock_guard<std::mutex> lock(m_mutex);
+        m_deadlineSet = false;
+        m_cv.notify_all();
+    }
+
+private:
+
+    void* m_context;
+    WaitTimerCallback* m_callback;
+    Deadline::duration m_duration;
```
lubeltra: why not store it as a time point (Deadline), this value is set to an absolute point in time (and later you have to do a whole bunch of casting)

Source/Task/iOS/ios_WaitTimer.mm
```diff
@@ -0,0 +1,110 @@
...
+}
+
+void WaitTimerImpl::Start(_In_ uint64_t absoluteTime)
+{
+    Cancel();
+    
+    auto duration = Deadline::duration(absoluteTime);
+    auto timePoint = Deadline(duration) - std::chrono::high_resolution_clock::now();
+
+    m_timer = [NSTimer scheduledTimerWithTimeInterval:timePoint.count() / 1000.0
```
lubeltra: can you use a std::duration_cast to convert to the correct units without making assumptions about the representation of Deadline?

Source/Task/WaitTimer_stl.cpp
brianpepin: Somehow I missed that in the old code.  I will add that back.

Source/Task/AsyncLib.cpp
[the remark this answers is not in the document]
brianpepin: Fixed.

Source/Task/AsyncQueue.cpp
brianpepin: Good catch -- added code to handle this.

Source/Task/WaitTimer_stl.cpp
brianpepin: Ok, I resurrected most of the old code.  I changed "Add" to "Set" because we only allow one value at a time now, and ensured the unit test tested more than one timer value at the same time.

Source/Task/iOS/ios_WaitTimer.mm
brianpepin: Will do. Separate push since I will want to verify on my mac.

Source/Task/WaitTimer_stl.cpp
brianpepin: Yup, cleaner.  Updated.

Source/Task/AsyncQueue.cpp
```diff
@@ -351,9 +351,12 @@ class Queue
     bool Wait(uint32_t timeout)
     {
         std::unique_lock<std::mutex> lock(m_cs);
-        if (!m_hasItems)
+        while (!m_hasItems)
```
lubeltra: there's a version of wait_for that takes a lambda with the predicate, saves you writing the loop and the break
```
m_event.wait_for(lock, std::chrono::milliseconds(timeout), [=](){ return !m_hasItems; });
```

Source/Task/WaitTimer_stl.cpp
```diff
@@ -7,88 +7,194 @@ class WaitTimerImpl
...
-    }
+        std::lock_guard<std::mutex> lock{ m_mutex };
 
-    void Cancel()
-    {
-        std::lock_guard<std::mutex> lock(m_mutex);
-        m_deadlineSet = false;
-        m_cv.notify_all();
+        auto it = std::find_if(m_queue.begin(), m_queue.end(), [timer](auto &item) { return item.Timer == timer; });
+        if (it != m_queue.end())
```
lubeltra: do you have unit tests checking this? It's been too long since I studied heaps to remember if this is a thing that is valid to do (or if it breaks invariants that push_heap relies on)

lubeltra: You may want to do what we do in remove instead, and just clear the timer pointer in the old entry and then add a new entry

PR #278, opened 2018-09-13
Fix potential race during cancel and running async call

brianpepin:
There is a potential narrow race condition in async lib if you cancel an async call while that call is in flight. This can cause the completion callback to be invoked, which is allowed to delete the async block.  If an async call was running at the time, it's provider data structure now contains a dangling pointer.

This was solved earlier by introducing a "zombie block" which is a copy of the user's async block that we swapped in when we were cleaning up.  This can cause the async field of async provider data to switch to a different pointer value, and while narrow, there is still a potential for a race with a running call.

This change keeps a local copy of the user's async block. It promotes the "zombie block" idea such that the async provider data _always_ gets the copy, while user callbacks _always_ get the original pointer.  We keep state transitions of the two in lockstep always.  This way no matter what the user does with the async block the async provider method always has valid data.

### DIFF
--- a/Source/Task/AsyncLib.cpp
+++ b/Source/Task/AsyncLib.cpp
@@ -8,16 +8,28 @@
 // Used by unit tests to verify we cleanup memory correctly.
 std::atomic<uint32_t> s_AsyncLibGlobalStateCount{ 0 };
 
+// Note that there are two AsyncBlock structures in play.
+// There is the pointer allocated and passed to us by the user
+// and there is our own local copy.  We pass the local copy to
+// async provider callbacks, and pass the user pointer to all
+// user-visible callbacks.  The AsyncBlockInternalGuard class keeps
+// the state of these two classes in sync under a lock.  We
+// do this because it is possible for a user to cancel an async
+// call while that call is still running.  We issue the completion
+// callback for the cancel immediatly, which gives the user an
+// opportunity to delete the user async block.  This would leave
+// the async provider callback with a dangling pointer.
+
 struct AsyncState
 {
     uint32_t signature = ASYNC_STATE_SIG;
     std::atomic<uint32_t> refs{ 1 };
     std::atomic<bool> workScheduled{ false };
     bool canceled = false;
-    bool zombieBlockSet = false;
     AsyncProvider* provider = nullptr;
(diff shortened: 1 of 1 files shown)


Source/Task/AsyncLib.cpp
```diff
@@ -425,32 +448,20 @@ static void CALLBACK CompletionCallbackForAsyncState(
     AsyncStateRef state;
     state.Attach(static_cast<AsyncState*>(context));
 
-    AsyncBlock* asyncBlock = state->providerData.async;
+    AsyncBlock* asyncBlock = state->userAsyncBlock;
     if (asyncBlock->callback != nullptr)
     {
         asyncBlock->callback(asyncBlock);
     }
```
jasonsandlin: i wonder if you should pin this in a local var.  aka can asyncBlock->callback change after the if but before it exceutes?

Seems like 

auto callback = asyncBlock->callback;
if( callback != nullptr )
{
     callback(asyncBlock);
}

would be safer.  This applies here and else where in the file

Source/Task/AsyncLib.cpp
```diff
@@ -164,20 +176,42 @@ class AsyncBlockInternalGuard
...
-    {
-        ASSERT(m_internal);
-        while (m_internal->lock.test_and_set()) {}
+        m_internal1{ reinterpret_cast<AsyncBlockInternal*>(asyncBlock->internal) },
+        m_internal2(nullptr)
+    {
+        ASSERT(m_internal1);
+        while (m_internal1->lock.test_and_set()) {}
+        
+        if (m_internal1->state != nullptr)
```
jasonsandlin: >m_internal1->state  [](start = 12, length = 19)

should this be pinned to local var?  can't tell if this can change out from under you from another thread

Source/Task/AsyncLib.cpp
brianpepin: It would narrow the window but it's not legal to change the async block while the call is still active.

Source/Task/AsyncLib.cpp
brianpepin: It shouldn't.  This is the only class that can mutate the state pointer and we grab the lock above.

PR #288, opened 2018-09-21
Fixing issue where websockets caused exception on Android

AesteroidBlues:
New threads need to be managed by the Java VM during their lifetime

### DIFF
--- a/Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
+++ b/Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
@@ -23,6 +23,7 @@
 #include <websocketpp/config/asio_client.hpp>
 #include <websocketpp/config/asio_no_tls_client.hpp>
 #include <websocketpp/client.hpp>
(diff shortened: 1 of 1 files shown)


Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
```diff
@@ -390,6 +393,22 @@ struct wspp_websocket_impl : public hc_websocket_impl, public std::enable_shared
                 auto context = http_allocate_shared<client_context>(client);
                 
                 m_websocketThread = std::thread([context](){
+#if HC_PLATFORM == HC_PLATFORM_ANDROID
+                    auto httpSingleton = xbox::httpclient::get_http_singleton(true);
+                    AndroidPlatformContext* platformContext = reinterpret_cast<AndroidPlatformContext*>(httpSingleton->m_platformContext.get());
```
jasonsandlin: since client.run() is blocking, this auto singleton will keep our singleton alive too long..

Wrap these two lines inside { } so the singleton goes out of scope quickly

PR #293, opened 2018-09-29
Adding support for HC_NOWEBSOCKETS inside websocket cpps.

matt-vanderkolk:
Added #if !HC_NOWEBSOCKETS to cpp files that refer to websockets code.


matt-vanderkolk:
Probably recalling this PR in order to perform a true qfe...

PR #301, opened 2018-10-05
Fix a racing case where could lead to AsyncState double dtor.

jicailiu:


lubeltra:
Reviewing, please don't commit yet

Source/Task/AsyncLib.cpp
```diff
@@ -759,6 +759,13 @@ STDAPI ScheduleAsync(
         RETURN_HR(E_UNEXPECTED);
     }
 
+    // Hold another ref here to prevent async call from finishing then triggering dtor before 
+    // we had a chance to call addRef after submission.
+    AsyncStateRef holder;
+    {
+        AsyncBlockInternalGuard internal{ asyncBlock };
+        holder = internal.GetState();
+    }
```
lubeltra: This will work, but there is a better solution I believe.

You should change the call below to state->AddRef() to state->Detach(), the net effect is the same (we "leak" one reference for the work callback), but it follows a more common pattern for this kind of thing.

PR #307, opened 2018-10-10
Moving async headers to enable swapping to OS version

jasonsandlin:


Include/httpClient/config.h
```diff
@@ -16,7 +16,8 @@
 #define HC_DATAMODEL_LP64 3 // int is 32 bit; long and pointer are 64 bits (64 bit unix)
 // see http://www.unix.org/version2/whatsnew/lp64_wp.html for detailed descriptions
 
-#if defined(_WIN32)
+#if defined(HC_PLATFORM) && defined(HC_DATAMODEL)
```
lubeltra: maybe move the check for defining a datamodel inside the if and output a helpful error for that?
something like: "if manually setting HC_PLATFORM also set HC_DATAMODEL"?

Include/httpClient/httpClient.h
```diff
@@ -5,9 +5,9 @@
 #include <httpClient/pal.h>
 #include <httpClient/mock.h>
 #include <httpClient/trace.h>
-#include <httpClient/async.h>
-#include <httpClient/asyncQueue.h>
-#include <httpClient/asyncQueueEx.h>
+#include <async.h>
+#include <asyncQueue.h>
+#include <asyncQueueEx.h>
```
lubeltra: does this header need asyncQueueEx.h?

jasonsandlin: removing it causes lots of breaks in all samples and clients using httpclient because they use APIs from it.  

similar to xsapi's services.h, httpClient.h is the header we expect clients to include.  its not worth adding complexity on the dev integration story for this

---
In reply to: [224275410](https://github.com/Microsoft/libHttpClient/pull/307#discussion_r224275410) [](ancestors = 224275410)

PR #312, opened 2018-10-22
Fix out of order locking bug in async lib

brianpepin:
Fixes a A,B & B,A locking problem that can cause deadlocks.

### DIFF
--- a/Source/Task/AsyncLib.cpp
+++ b/Source/Task/AsyncLib.cpp
@@ -176,42 +176,40 @@ class AsyncBlockInternalGuard
 {
 public:
     AsyncBlockInternalGuard(_Inout_ AsyncBlock* asyncBlock) noexcept :
-        m_internal1{ reinterpret_cast<AsyncBlockInternal*>(asyncBlock->internal) },
-        m_internal2(nullptr)
+        m_internal(DoLock(asyncBlock))
     {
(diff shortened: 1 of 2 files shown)


Source/Task/AsyncLib.cpp
```diff
@@ -265,8 +257,46 @@ class AsyncBlockInternalGuard
...
+
+        if (internal->state != nullptr && asyncBlock != &internal->state->asyncBlock)
+        {
+            // Grab a state ref here because releasing the lock can allow
+            // the state to be cleared / released.
+            AsyncStateRef state(internal->state);
+            internal->lock.clear();
+            internal = DoLock(&state->asyncBlock);
+
+            // We locked the right object, but we need to check here to see if we
```
lubeltra: I'm confused by this part.

When checking for `internal->state == nullptr`, are we checking if we are racing with someone else?
The heavy reuse of the local variable `internal` with different values makes it hard to follow what's happening. Can you use different names? (I know it inhibits NRVO, but it's a pointer so I don't think NRVO applies anyway)

brianpepin: Yup, I can clarify.  Once we have internal assigned, we've been granted the lock so there is no chance of another actor messing with state.

PR #318, opened 2018-10-25
Switching out Apple timer implementation for the STL timer.

matt-vanderkolk:
This change is here to sidestep an issue regarding how runloops are managed in iOS which can cause libHttpClient to never have its timers finished if the running app does not properly run the runloop.

As a short-term fix, we are switching to using the STL timer until we have a longer term solution.


matt-vanderkolk:
This change is currently in testing. Please do not merge until testing has completed.

edingwell:
> This change is currently in testing. Please do not merge until testing has completed.

Just finished verifying this change, so we should b good to go

PR #334, opened 2018-11-06
Fixing mem leak upon shutdown

jasonsandlin:


Source/Task/WaitTimer_stl.cpp
```diff
@@ -121,6 +134,9 @@ void TimerQueue::Worker() noexcept
     std::unique_lock<std::mutex> lock{ m_mutex };
     while (true)
     {
+        if (m_exitThread)
+            return;
```
lubeltra: please put braces around the if block

lubeltra: I think as written there is a race condition: if m_exitThread is set to true and the condition variable is signalled while we are looping over the queue, we will never exit

I suggest the following change:
```c++
while (!m_exitThread)
{
    // do work
    if (!m_exitThread)
    {
        // wait on condition variable
    }
}

File: Source/Task/WaitTimer_stl.cpp [hunk 2]
```diff
@@ -205,6 +221,7 @@ WaitTimer::~WaitTimer() noexcept
     if (m_impl != nullptr)
     {
         delete m_impl;
+        m_impl = nullptr;
```
lubeltra: unnecessary

Source/Task/WaitTimer_stl.cpp
```diff
@@ -132,7 +145,7 @@ void TimerQueue::Worker() noexcept
             TimerEntry entry = Pop();
 
             // release the lock while invoking the callback, just in case timer
-            // gets destroyed on this thread or readds itself in the callback
+            // gets destroyed on this thread or reads itself in the callback
```
lubeltra: readds (adds again) was correct, we are worried about the callback interacting with TimerQueue

jasonsandlin: Works due to lock

PR #337, opened 2018-11-19
Renaming Source/mock files to lhc_mock

MSFT-Heba:
Renaming Source/mock files to lhc_mock


Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
```diff
@@ -216,8 +216,8 @@
 			isa = PBXGroup;
 			children = (
 				58A7E982209ADEB100CC6774 /* mock_publics.cpp */,
```
matt-vanderkolk: Nit: The order of these dictates the order that Xcode shows them in the editor. alphabetize these please.

PR #339, opened 2018-11-28
Fixing error handling and thread sync bug in WinHTTP implementation

jasonsandlin:


Include/httpClient/httpClient.h
```diff
@@ -223,6 +223,18 @@ STDAPI HCHttpCallPerformAsync(
     _Inout_ AsyncBlock* asyncBlock
     ) HC_NOEXCEPT;
 
+/// <summary>
+/// Get hc_call_handle_t from an HCHttpCallPerformAsync call.
+/// Use HCHttpCallResponse* to get information from the hc_call_handle_t.
+/// </summary>
```
jplafonta: If you are cleaning up the hc_call_handle_t object on failure, make that clear in the comments.

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -31,20 +31,34 @@ winhttp_http_task::winhttp_http_task(
...
+    complete_task(translatedHR, translatedHR);
+}
+
+void winhttp_http_task::complete_task(_In_ HRESULT translatedHR, uint32_t platformSpecificError)
+{
+    HCHttpCallResponseSetNetworkErrorCode(m_call, translatedHR, platformSpecificError);
+    CompleteAsync(m_asyncBlock, S_OK, 0);
+    HCHttpCallSetContext(m_call, nullptr);
+    WinHttpSetStatusCallback(m_hSession, nullptr, WINHTTP_CALLBACK_FLAG_ALL_NOTIFICATIONS, NULL);
+    shared_ptr_cache::fetch<winhttp_http_task>(reinterpret_cast<void*>(this), true, false);
```
jplafonta: >reinterpret_cast<void*> [](start = 47, length = 23)

nit: I believe this is not needed

Source/HTTP/WinHttp/winhttp_http_task.h
```diff
@@ -16,6 +16,39 @@ enum msg_body_type
...
+};
+
+class win32_cs_autolock
+{
+public:
+    win32_cs_autolock(win32_cs* pCS)
+        : m_pCS(pCS)
+    {
+        m_pCS->lock();
+        //HC_TRACE_INFORMATION(HTTPCLIENT, "win32_cs_autolock locked [ID %lu]", GetCurrentThreadId());
```
jplafonta: remove? or uncomment if its useful

jasonsandlin: its kinda spammy but useful when watching threads so i think just leaving it like this is fine for now

---
In reply to: [237280338](https://github.com/Microsoft/libHttpClient/pull/339#discussion_r237280338) [](ancestors = 237280338)

Samples/Win32-Http/main.cpp
```diff
@@ -220,42 +212,57 @@ int main()
...
         std::string errMessage;
 
-        hc_call_handle_t call = static_cast<hc_call_handle_t>(asyncBlock->context);
-        HCHttpCallResponseGetNetworkErrorCode(call, &errCode, &platErrCode);
+        HRESULT hr = GetAsyncStatus(asyncBlock, false);
+        if (FAILED(hr)) 
+        {
+            // This should be a rare error case when the async task fails
+            printf_s("Couldn't get HTTP call object 0x%0.8x\r\n", hr);
+            return;
```
jplafonta: Add HcCallCloseHandle here

PR #340, opened 2018-12-05
Update the platform hook model andadd a "generic" platform

lubeltra:
This work does not fix websockets yet
The generic platform does not implement any of the platform specific functionality and relies on the client to set all the necessary hooks


Include/httpClient/httpProvider.h
```diff
@@ -6,41 +6,52 @@
...
 /// <param name="call">The handle of the HTTP call</param>
 /// <param name="asyncBlock">The asyncBlock of the async task</param>
 typedef void
-(STDAPIVCALLTYPE* HCCallPerformFunction)(
+(CALLBACK* HCCallPerformFunction)(
     _In_ hc_call_handle_t call,
-    _Inout_ AsyncBlock* asyncBlock
+    _Inout_ AsyncBlock* asyncBlock,
+    _In_opt_ void* context,
+    _In_ hc_perform_env env
```
matt-vanderkolk: <param> elements

Include/httpClient/httpProvider.h
```diff
@@ -6,41 +6,52 @@
...
-/// Pass in nullptr to use the default implementation based on the current platform</param>
-/// <returns>Result code for this API operation.  Possible values are S_OK, or E_FAIL.</returns>
-STDAPI_(void) HCSetHttpCallPerformFunction(
-    _In_opt_ HCCallPerformFunction performFunc
+/// </remarks>
+/// <returns>Result code for this API operation.  Possible values are S_OK, or
+/// E_HC_ALREADY_INITIALISED.</returns>
+STDAPI HCSetHttpCallPerformFunction(
+    _In_ HCCallPerformFunction performFunc,
+    _In_opt_ void* performContext
```
matt-vanderkolk: >performContext [](start = 19, length = 14)

<param>

Include/httpClient/httpProvider.h
```diff
@@ -6,41 +6,52 @@
...
+/// <remarks>
+/// This can be used along with HCSetHttpCallPerformFunction() to build a filter that
+/// monitors and modifies all HTTP calls, while still using the default HTTP implementation.
+/// </remarks>
+/// <returns>Result code for this API operation.  Possible values are S_OK,
+/// E_HC_ALREADY_INITIALISED, or E_INVALIDARG.</returns>
 STDAPI HCGetHttpCallPerformFunction(
-    _Out_ HCCallPerformFunction* performFunc
+    _Out_ HCCallPerformFunction* performFunc,
+    _Out_ void** performContex
```
matt-vanderkolk: >performContex [](start = 17, length = 13)

<param>

Source/Logger/Generic/generic_logger.cpp
```diff
@@ -0,0 +1,26 @@
+#include "pch.h"
+
+#include "../trace_internal.h"
+
+uint64_t Internal_ThisThreadId() noexcept
+{
+    auto tidi = GetThreadIdInfo();
+    if (tidi.callback)
```
matt-vanderkolk: >tidi.callback [](start = 8, length = 13)

can these callbacks be set at runtime? If so you should have to do something like
auto callback = foo.callback
if (callback)
{
    callback(...)
}

matt-vanderkolk: here and below

---
In reply to: [239286364](https://github.com/Microsoft/libHttpClient/pull/340#discussion_r239286364) [](ancestors = 239286364)

Source/Global/global.cpp
```diff
@@ -17,12 +17,18 @@ static std::shared_ptr<http_singleton> g_httpSingleton_atomicReadsOnly;
...
 NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
 
-http_singleton::http_singleton(IHCPlatformContext* initialContext) :
-    m_platformContext{ initialContext }
+http_singleton::http_singleton(PerformInfo const& performInfo, PerformEnv&& performEnv) :
+    m_callRoutedHandlersContext{ 0 },
+    m_perform{ performInfo },
+    m_performEnv{ std::move(performEnv) },
+    m_lastId{ 0 },
+    m_retryAllowed{ false },
```
matt-vanderkolk: >m_retryAllowed{ false } [](start = 4, length = 23)

You're switching the default value for this. Is that intentional? Do you have to turn this back on somewhere that I'm missing?

Source/Logger/Generic/generic_logger.cpp
lubeltra: auto tidi is making a copy of the object returned by GetThreadInfo();
(auto& would preserve the reference)

Source/Global/global.cpp
lubeltra: I think I started from older code that defaulted to retries off

PR #344, opened 2018-12-12
Update Async and AsyncQueue to new design guidelines

brianpepin:
This change incorporates api breaking changes from Windows done as a result of design feedback.  Specifically:

**Design Changes**
* AsyncQueue.h -> XTaskQueue.h
* New task queue termination APIs
* New task queue wait apis
* Lots of small changes based on design feedback
* Better perf - task queue is now lock free when submitting new items

**File Renames**
* Async.h -> XAsync.h, AsyncProvider.h -> XAsyncProvider.h.  

**Guideline Updates**
* No more foo_handle_t. Replaced by XFooHandle.
* All APIs prefixed with X and a common category, like XAsync or XTaskQueue


Include/XTaskQueue.h
```diff
@@ -0,0 +1,309 @@
...
+/// <param name='queue'>The queue to submit the callback to.</param>
+/// <param name='port'>The port to invoke the callback on.</param>
+/// <param name='waitHandle'>The handle to monitor.</param>
+/// <param name='callbackContext'>An optional context pointer that will be passed to the callback.</param>
+/// <param name='callback'>A pointer to the callback function.</param>
+/// <param name='token'>A registration token.</param>
+STDAPI XTaskQueueRegisterWaiter(
+    _In_ XTaskQueueHandle queue,
+    _In_ XTaskQueuePort port,
+    _In_ HANDLE waitHandle,
```
lubeltra: How does this work on non windows platforms?

Include/XTaskQueue.h
```diff
@@ -0,0 +1,309 @@
...
+/// <param name='token'>The token returned from XTaskQueueRegisterMonitor.</param>
+STDAPI_(void) XTaskQueueUnregisterMonitor(
+    _In_ XTaskQueueHandle queue,
+    _In_ XTaskQueueRegistrationToken token
+    ) noexcept;
+
+/// <summary>
+/// Returns a handle to the process task queue, or nullptr if there is no
+/// process task queue.  By default, there is a process task queue
+/// that uses the thread pool for both work and completion ports. You
```
lubeltra: What is the default on non windows platforms?

Include/XTaskQueue.h
```diff
@@ -0,0 +1,309 @@
...
+/// can replace the default process task queue by calling 
+/// XTaskQueueSetCurrentProcessTaskQueue.  There is no need to close the
+/// handle returned from XTaskQueueGetCurrentProcessTaskQueue.
```
lubeltra: mmm delicious race conditions

Include/XAsync.h
```diff
@@ -0,0 +1,107 @@
...
+/// <seealso cref='XAsyncBlock' />
+/// <seealso cref='XAsyncRun' />
+typedef HRESULT CALLBACK XAsyncWork(_Inout_ struct XAsyncBlock* asyncBlock);
+
+struct XAsyncBlock
+{
+    /// <summary>
+    /// The queue to queue the call on
+    /// </summary>
+    XTaskQueueHandle queue;
```
lubeltra: is this optional again?

Include/XTaskQueue.h
brianpepin: It returns E_NOTIMPL.

Include/XTaskQueue.h
brianpepin: ThreadPool, ThreadPool, which is the default on all platforms. There is now a simple thread pool implementation in STL.

Include/XTaskQueue.h
brianpepin: Could be. The idea is a process configures this at startup before it begins making calls.  I can see your point though -- if more and more stuff starts using this then the possibility of a race condition before the process queue can be configured is real.

Include/XAsync.h
brianpepin: Yes, but selectively so. IE, an app can set nullptr into the process task queue and then this parameter is no longer optional.  Basically, internal code uses the process task queue if this member is null, and if it doesn't get a valid one it errors with E_NO_TASK_QUEUE.

Source/Task/LocklessList.h
```diff
@@ -0,0 +1,365 @@
...
+
+ Maged M. Michael Michael L. Scott
+ Department of Computer Science University of Rochester Rochester, NY 14627-0226 fmichael,scottg@cs.rochester.ed
+
+ http://cs.rochester.edu/~scott/papers/1996_PODC_queues.pdf
+ 
+ ******************************************************************************/
+
+template <typename TData>
+class alignas(sizeof(std::uintptr_t) * 8) LocklessList
```
lubeltra: does this build in vs2015?
do we still care about 2015 support?

Source/Task/StaticArray.h
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+
+template <typename TData, uint32_t size>
+class StaticArray
```
lubeltra: what does this give us that std::array does not?

Source/Task/TaskQueue.cpp
```diff
@@ -0,0 +1,1590 @@
+// Copyright (c) Microsoft Corporation
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.#include "stdafx.h"
+#include "pch.h"
+#include "referenced_ptr.h"
+#include "TaskQueueP.h"
+#include "TaskQueueImpl.h"
+
+namespace ApiDiag
```
lubeltra: I think this code may run into initialization order issues.
Can any of these globals be touched as a consequence of a constructor running in another compilation unit?

Source/Task/ThreadPool_stl.cpp
```diff
@@ -0,0 +1,130 @@
...
+        {
+            thread.join();
+        }
+
+        m_pool.clear();
+    }
+
+    void Submit() noexcept
+    {
+        std::lock_guard<std::mutex> lock(m_wakeLock);
```
lubeltra: for reduced lock contention you may want to release lock before calling notify_all

Source/Task/WaitTimer_stl.cpp
```diff
@@ -6,6 +6,7 @@ using Deadline = std::chrono::high_resolution_clock::time_point;
 class WaitTimerImpl
```
lubeltra: You should double check these changes, I don't think you want them

Source/Task/LocklessList.h
brianpepin: It built for me.

Source/Task/StaticArray.h
brianpepin: It has some convenience -- you can remove an element from the middle and the contents shift.  Basically, I can treat this like a vector or other container from an API standpoint, but it's fixed size with no allocation.

Source/Task/TaskQueue.cpp
brianpepin: They are only present for tracking leaks in the unit tests. They aren't used at runtime.  If there are parallel copies, no harm.

Source/Task/StaticArray.h
lubeltra: we could build those convenience functions on top of std::array, not that this code is that tricky, but doesn't feel right to rewrite it

Source/Task/TaskQueue.cpp
lubeltra: It's not about parallel copies, it's about code seeing the values before they are initialized, could you add a comment that we only use it for tests and so not care about initialization order?

Source/Task/StaticArray.h
brianpepin: I looked at this a little deeper.  I don't think there's value in using std::array for this.  The only thing I really get from std::array is a [] operator.  The rest of the api in StaticArray would need to be added to std::array through derivation (you still are defnining a new class) or helper methods (which removes the ability to treat this like an object).

lubeltra: ok

PR #346, opened 2018-12-20
Removing ifndef __cplusplus blocks

matt-vanderkolk:


matt-vanderkolk:
@brianpepin The error directives were breaking our c-builds, was their addition intentional or is putting the extern "C" into ifdefs like the rest of our code good enough?

brianpepin:
Yes, these are intentional as part of the design guidelines on the windows side.

PR #353, opened 2019-01-08
Create winhttp session once per init, and fix http close handle bug

jasonsandlin:


Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -13,27 +13,102 @@
 
 using namespace xbox::httpclient;
 
+
+HC_PERFORM_ENV::HC_PERFORM_ENV()
+{
+    m_proxyType = get_ie_proxy_info(proxy_protocol::https, m_proxyUri);
```
jplafonta: Maybe make a note of future work to split out https proxy vs proxy for other protocols

PR #357, opened 2019-01-10
Switching to C++11 enum class types based on latest API style convention

jasonsandlin:


Include/httpClient/trace.h
```diff
@@ -103,38 +109,38 @@
...
 /// Diagnostic level used by tracing
 /// </summary>
-typedef enum HCTraceLevel
+enum class HCTraceLevel : uint32_t
 {
     /// <summary>
     /// No tracing
     /// </summary>
-    HCTraceLevel_Off = HC_PRIVATE_TRACE_LEVEL_OFF,
+    Off = HC_PRIVATE_TRACE_LEVEL_OFF,
```
jplafonta: >O [](start = 4, length = 1)

are enum values supposed to be capitalized or lower case?

PR #358, opened 2019-01-11
Add Begin opcode to async provider

brianpepin:
Adding this opcode allows async implementations to be entirely started / controlled by the async provider callback.


brianpepin:
Approving as Luca and I have bene discussing this in IM.

PR #361, opened 2019-01-16
Adding WinHTTP Websocket support

jasonsandlin:


jplafonta:
> // Copyright (c) Microsoft Corporation

Make sure this builds against win7

---
Refers to: Source/HTTP/WinHttp/winhttp_http_task.cpp:1 in 8553530. [](commit_id = 85535309fa725fb5a1c7d32cf517e55dc300dc91, deletion_comment = False)

Source/WebSocket/WinHTTP/winhttp_websocket.cpp
```diff
@@ -0,0 +1,252 @@
+// Copyright (c) Microsoft Corporation
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "pch.h"
+
+#if HC_WINHTTP_WEBSOCKETS
+#if !HC_NOWEBSOCKETS
```
jplafonta: seems a bit odd - I think these should never be defined together right?

Source/WebSocket/WinHTTP/winhttp_websocket.cpp
```diff
@@ -0,0 +1,252 @@
...
+using namespace xbox::httpclient;
+
+
+struct winhttp_websocket_impl : public hc_websocket_impl, public std::enable_shared_from_this<winhttp_websocket_impl>
+{
+public:
+    winhttp_websocket_impl(HCWebsocketHandle hcHandle)
+    {
+        m_hcWebsocketHandle = HCWebSocketDuplicateHandle(hcHandle);
+        // TODO: when to close
```
jplafonta: >/ TODO: when to close [](start = 9, length = 21)

I think you don't actually want to duplicate the handle here.  iirc this object is owned by the HC_WEBSOCKET, so duplicated the handle here will cause it to never get cleaned up.

jplafonta: the websocket_impl object's lifetime is always the same as the HC_WEBSOCKET

---
In reply to: [248444615](https://github.com/Microsoft/libHttpClient/pull/361#discussion_r248444615) [](ancestors = 248444615)

Source/WebSocket/hcwebsocket.cpp
```diff
@@ -101,25 +109,71 @@ try
...
+            case XAsyncOp::GetResult:
+                break;
+
+            case XAsyncOp::Cancel:
+                break;
+
+            case XAsyncOp::Cleanup:
+            {
+                websocket_connect_context* context = static_cast<websocket_connect_context*>(data->context);
+                //HCWebSocketCloseHandle(context->websocket); // TODO: jasonsa - fix for WinHTTP
```
jplafonta: where does this get duplicated?

Source/WebSocket/hcwebsocket.cpp
```diff
@@ -101,25 +109,71 @@ try
...
+    std::shared_ptr<websocket_connect_context> connectContext = std::make_shared<websocket_connect_context>();
+    connectContext->uri = uri;
+    connectContext->subProtocol = subProtocol;
+    connectContext->websocket = websocket;
+    connectContext->outerAsyncBlock = asyncBlock;
+    websocket_connect_context* rawConnectContext = static_cast<websocket_connect_context*>(shared_ptr_cache::store<websocket_connect_context>(connectContext));
 
-    auto connectFunc = httpSingleton->m_websocketConnectFunc;
-    if (connectFunc != nullptr)
+    HRESULT hr = XAsyncBegin(asyncBlock, rawConnectContext, reinterpret_cast<void*>(HCWebSocketConnectAsync), __FUNCTION__,
```
jplafonta: why did you change this such that the connect is kicked off asynchronously?

Source/WebSocket/WinHTTP/winhttp_websocket.cpp
```diff
@@ -0,0 +1,252 @@
...
+    _In_ HCWebsocketHandle websocket,
+    _Inout_ XAsyncBlock* asyncBlock,
+    _In_ HCPerformEnv env
+    )
+{
+    assert(env != nullptr);
+
+    websocket->uri = uri;
+    websocket->subProtocol = subProtocol;
+    auto wsppSocket = http_allocate_shared<winhttp_websocket_impl>(websocket);
```
jplafonta: rename

jplafonta: seems like we should check that we aren't already connected here? Maybe set websocket->impl to null when we disconnect?  I don't think we are doing this in wspp but should probably be fixed in both places.

Source/WebSocket/hcwebsocket.cpp
jasonsandlin: Ah, good catch.  This is wrong

Source/WebSocket/WinHTTP/winhttp_websocket.cpp
jasonsandlin: You should be able to disconnect/reconnect without tossing handle so impl != null isn't right logic, but I'll add a TODO to guard against double connect

Source/WebSocket/hcwebsocket.cpp
jasonsandlin: bug.  it should be above.  the intent is that we keep the HCWebsocketHandle alive while connecting/connected, and let it be destroyed only when its disconnected.  we'll have to chat a bit about this

PR #382, opened 2019-02-26
Fix analysis warnings

brianpepin:


Source/Task/XTaskQueuePriv.h
```diff
@@ -4,7 +4,7 @@
 
 #pragma once
```
jasonsandlin: revert this one.  Matt made this change for xplat work

Source/Task/XTaskQueuePriv.h
```diff
@@ -4,7 +4,7 @@
...
 
-#include <httpClient/async.h>
+#include "XTaskQueue.h"
```
jasonsandlin: revert this one. Matt made this change for xplat work

brianpepin: This header is shared with Windows though -- there is no httpClient there.  Was this one done wrong?

lubeltra: Brian's change is correct

Issue #385, opened 2019-03-04
Android doesn't propagate failed scenario. 


namrog84:
When letting a http call timeout on windows 
HCHttpCallResponseGetNetworkErrorCode returns a networkErrorCode failed and platformNetworkErrorCode 12002 (timeout)

However, on android they both return as 0, which are successful responses. 

There is a bug that android is not propagating any failures.  (I was testing on a google xl 2)

edit:
As seen in my 2nd comment. There appears to be a bug here.

namrog84:
Just tested on android for timeout
HCHttpCallResponseGetNetworkErrorCode
HCHttpCallResponseGetStatusCode
networkErrorCode: 0, platformNetworkErrorCode : 0, httpStatus: 0

Other than assuming HttpStatus 0 is failed, is there a better way to determine failed state on Android?

namrog84:
It appears like XAsyncComplete is reporting a failure here.
https://github.com/Microsoft/libHttpClient/blob/51e49a7bae3685578a3f662b8bcd8461ebeb500b/Source/HTTP/Android/http_android.cpp#L20

However in that implementation
https://github.com/Microsoft/libHttpClient/blob/92fd932fddf7b7db20a1c482f053f1a50b4d90a4/Source/Task/AsyncLib.cpp#L773
it appears to be dropping the result on the floor and stops propogating it past XAsyncComplete.

I don't really know enough about this section of the code to propose a fix, but I suspect it'll be in XAsyncComplete by setting some result or E_FAIL to something inside of state

namrog84:
Closed because of PR

PR #386, opened 2019-03-04
Enable android propagating failures.

namrog84:
It doesn't appear HCHttpCallResponseSetNetworkErrorCode was ever being called in the Android Path.

Also this might be a good place to convert Java exceptions into platform specific errors in the future.

Fixing my earlier issue of #385


namrog84:
:thumbsup:

PR #394, opened 2019-03-29
Add Windows 10 on ARM64 support to projects that can build for it

marswe:
To support building libhttpclient as ARM64, this change moves projects forward to the latest SDK (17763) and adds ARM64 configurations.  Building as ARM64 allows native Win32 or UWP apps on Windows 10 on ARM to use libhttpclient.

For compatibility, these changes add _WIN32_WINNT=0x601 to the win32 project files so that the resulting built libs can still run on Win7.  Note that the UnitTest.TE project can't build for ARM64 yet and so was left unchanged.

ARM64 versions of all the win32 and UWP test apps run successfully with these changes.


jasonsandlin:
Can you open the template files at https://github.com/Microsoft/libHttpClient/tree/master/Utilities/CMake with the same change?  We use cmake to auto-gen the vcxprojs when we add new source files.  You don't need to run cmake, just update the templates.

Build/libHttpClient.141.Win32.C/libHttpClient.141.Win32.C.vcxproj
```diff
@@ -35,7 +43,7 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
     <PreferredToolArchitecture>x64</PreferredToolArchitecture>
-    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
jasonsandlin: Let me ensure this change doesn't cause disruption

PR #399, opened 2019-04-24
Refactor websocket hooks to follow the same pattern as http hooks, add context pointer

lubeltra:


Source/WebSocket/WinRT/winrt_websocket.cpp
```diff
@@ -578,7 +581,8 @@ void MessageWebSocketSendMessage(
 
 HRESULT CALLBACK Internal_HCWebSocketDisconnect(
     _In_ HCWebsocketHandle websocket,
-    _In_ HCWebSocketCloseStatus closeStatus
+    _In_ HCWebSocketCloseStatus closeStatus,
+    _In_opt_ void* context
```
jplafonta: does this change not need to be made to winhttp_websocket.cpp as well?

lubeltra: I don't think it's actually being build by anyone, so I missed it

jplafonta: I believe its used by gsdk, and optionally on win32

---
In reply to: [278250082](https://github.com/Microsoft/libHttpClient/pull/399#discussion_r278250082) [](ancestors = 278250082)

Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
```diff
@@ -809,7 +811,9 @@ HRESULT CALLBACK Internal_HCWebSocketSendBinaryMessageAsync(
     _In_ HCWebsocketHandle websocket,
     _In_reads_bytes_(payloadSize) const uint8_t* payloadBytes,
     _In_ uint32_t payloadSize,
-    _Inout_ XAsyncBlock* asyncBlock)
+    _Inout_ XAsyncBlock* asyncBlock,
+    _In_opt_ void* context
```
jplafonta: Also shouldn't the implementation be updated so this actually gets passed back to the callback? here and elsewhere?

lubeltra: no this context is for this function, it's not what the consumer registers with their own callbacks (that comes out in get event handlers)

PR #400, opened 2019-04-25
Fixing cleanup when singleton is dtor'd

jasonsandlin:


Source/HTTP/httpcall.cpp
```diff
@@ -449,6 +449,11 @@ try
     retryContext->outerAsyncBlock = asyncBlock;
     retryContext->outerQueue = asyncBlock->queue;
     retry_context* rawRetryContext = static_cast<retry_context*>(shared_ptr_cache::store<retry_context>(retryContext));
+    if (rawRetryContext == nullptr)
+    {
+        XAsyncComplete(asyncBlock, E_HC_NOT_INITIALISED, 0);
+        return E_HC_NOT_INITIALISED;
+    }
```
jplafonta: maybe try to get_http_singleton at the beginning of the function and return E_NOT_INITIALIZED before duplicating the handle/allocating the context? I don't think you need to complete the async block in this case either.

jasonsandlin: Done

Source/HTTP/httpcall.cpp
```diff
@@ -449,10 +451,21 @@ try
...
```
jplafonta: if you don't check that the singleton is alive before duplicating the handle, you are leaking a handle here.  It's supposed to be cleaned up after the async work is completed, but you are not kicking it off even in this case. This check should be at the beginning of HCHttpCallPerformAsync

Source/HTTP/httpcall.cpp
```diff
@@ -449,10 +451,21 @@ try
...
 
     HRESULT hr = XAsyncBegin(asyncBlock, rawRetryContext, reinterpret_cast<void*>(HCHttpCallPerformAsync), __FUNCTION__,
         [](_In_ XAsyncOp op, _In_ const XAsyncProviderData* data)
     {
+        auto httpSingleton = get_http_singleton(false);
+        if (nullptr == httpSingleton)
+        {
+            return E_HC_NOT_INITIALISED;
+        }
```
jplafonta: if this condition is ever true we are leaking a handle as well.  context is never cleaned up and HCHttpCallCloseHandle is never called

Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
```diff
@@ -596,9 +596,22 @@ struct wspp_websocket_impl : public hc_websocket_impl, public std::enable_shared
...
+        }
+
+        auto hr = XAsyncBegin(sendContext->message.async, rawSendContext, (void*)HCWebSocketSendMessageAsync, __FUNCTION__,
             [](XAsyncOp op, const XAsyncProviderData* data)
         {
+            auto httpSingleton = get_http_singleton(false);
+            if (nullptr == httpSingleton)
+            {
+                return E_HC_NOT_INITIALISED;
+            }
```
jplafonta: if this condition is being met we are never removing the context form the shared_ptr_cache.

PR #415, opened 2019-05-14
Fixing websocket callback bug on Xbox after websocket handle is closed

jasonsandlin:
HCWebSocketConnectAsync does a websocket->AddRef() to keep the socket alive until so this isn't needed

### DIFF
--- a/Source/WebSocket/WinRT/winrt_websocket.cpp
+++ b/Source/WebSocket/WinRT/winrt_websocket.cpp
@@ -282,7 +282,7 @@ HRESULT CALLBACK Internal_HCWebSocketConnectAsync(
 {
     UNREFERENCED_PARAMETER(env);
     std::shared_ptr<winrt_websocket_impl> websocketTask = std::make_shared<winrt_websocket_impl>();
-    websocketTask->m_websocketHandle = HCWebSocketDuplicateHandle(websocket);
+    websocketTask->m_websocketHandle = websocket;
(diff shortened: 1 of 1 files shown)


Source/WebSocket/WinRT/winrt_websocket.cpp
```diff
@@ -282,7 +282,7 @@ HRESULT CALLBACK Internal_HCWebSocketConnectAsync(
 {
     UNREFERENCED_PARAMETER(env);
     std::shared_ptr<winrt_websocket_impl> websocketTask = std::make_shared<winrt_websocket_impl>();
-    websocketTask->m_websocketHandle = HCWebSocketDuplicateHandle(websocket);
```
jplafonta: >HCWebSocketDuplicateHandle [](start = 39, length = 26)

I think it is also called in send.

jasonsandlin: Got it

jplafonta: Where did this get closed anyhow? do we need to remove that?

---
In reply to: [284028839](https://github.com/microsoft/libHttpClient/pull/415#discussion_r284028839) [](ancestors = 284028839)

jasonsandlin: Right, good call.  The closes need to be removed as well

PR #420, opened 2019-05-22
Handle httpCall that completes after HCCleanup is called

jplafonta:


Source/HTTP/httpcall.cpp
```diff
@@ -386,6 +386,13 @@ void retry_http_call_until_done(
 
     nestedBlock->callback = [](XAsyncBlock* nestedAsyncBlock)
     {
+        auto httpSingleton = get_http_singleton(false);
+        if (nullptr == httpSingleton)
+        {
+            HC_TRACE_INFORMATION(HTTPCLIENT, "Http completed after HCCleanup was called. Aborting call.");
```
lubeltra: I'm make this a warning

PR #447, opened 2019-07-18
Mock updates

jplafonta:
* Add support for getting an intermediate callback when a mock is matched.
* Add ability to remove a single mock without having to clear them all.


Source/Mock/lhc_mock.cpp
```diff
@@ -15,7 +15,7 @@ bool DoesMockCallMatch(_In_ const HC_CALL* mockCall, _In_ const HC_CALL* origina
     }
     else
     {
-        if (originalCall->url == mockCall->url)
+        if (originalCall->url.substr(0, mockCall->url.size()) == mockCall->url)
```
rakhad: >url [](start = 26, length = 3)

Are you considering
original call: [http://xboxlive.com/achievements?maxItems=10](http://xboxlive.com/achievements?maxItems=10)
   mock call: [http://xboxlive.com/achievements](http://xboxlive.com/achievements)
equal?

rakhad: I guess one scenario you might do this for is if you want to create a mock for all calls to [http://xboxlive.com/achievements](http://xboxlive.com/achievements) regardless of query parameters

---
In reply to: [305495401](https://github.com/microsoft/libHttpClient/pull/447#discussion_r305495401) [](ancestors = 305495401)

jplafonta: Ya, I want those to be equal. If you don't want to match you just specify the full url.  Peoplehub also has examples of this where the responses to /batch and /friends queries are identical.

---
In reply to: [305498711](https://github.com/microsoft/libHttpClient/pull/447#discussion_r305498711) [](ancestors = 305498711,305495401)

PR #464, opened 2019-09-14
Fix a subtle race shutting down a task queue with deferred calls

brianpepin:
If a task queue is executing a deferred callback and is asked to terminate, it's possible the the deferred call runs while the queue task port is destructing.  This can cause a crash.


Source/Task/WaitTimer_win32.cpp
```diff
@@ -94,6 +97,16 @@ HRESULT WaitTimer::Initialize(_In_opt_ void* context, _In_ WaitTimerCallback* ca
     return S_OK;
 }
 
+void WaitTimer::Terminate() noexcept
+{
+    WaitTimerImpl* impl = static_cast<WaitTimerImpl*>(InterlockedExchangePointer(reinterpret_cast<volatile PVOID*>(&m_impl), nullptr));
```
lubeltra: should we use atomics for all accesses to m_impl? should we do the same for the stl timer?

(the fact we aren't using any "thread safe" access for m_impl right now worries me)

brianpepin: We could but it's not required because these are only cleaned up in the port dtor.  Since that is already ref-counted it's guarded there.

brianpepin: Ok, moved this to an atomic in case this timer is used outside the task queue code.

Source/Task/AsyncLib.cpp
```diff
@@ -71,7 +79,12 @@ struct AsyncState
     {
         if (provider != nullptr)
         {
-            (void)provider(XAsyncOp::Cleanup, &providerData);
+            // Have we already called cleanup?
+            ProviderCleanupLocation loc = providerCleanup.exchange(ProviderCleanupLocation::CleanedUp);
```
lubeltra: throw in an assert that loc is either CleanedUp or Destructor?

PR #465, opened 2019-09-16
Fix hang during wspp shutdown

jplafonta:


Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
```diff
@@ -44,6 +44,7 @@
 
 #define SUB_PROTOCOL_HEADER "Sec-WebSocket-Protocol"
 #define WSPP_PING_INTERVAL_MS 1000
+#define WSPP_SHUTDOWN_TIMEOUT 5000
```
rakhad: Add time interval to this

PR #476, opened 2019-10-21
Fixing HC_CALL memory management

jasonsandlin:


Source/HTTP/httpcall.cpp
```diff
@@ -371,7 +374,8 @@ void retry_http_call_until_done(
     auto httpSingleton = get_http_singleton(false);
     if (nullptr == httpSingleton)
     {
-        XAsyncComplete(retryContext->outerAsyncBlock, S_OK, 0);
+        HC_TRACE_WARNING(HTTPCLIENT, "Http call after HCCleanup was called. Aborting call.");
```
jplafonta: > HC_TRACE_WARNING(HTTPCLIENT, "Http call after HCCleanup was called. Aborting call."); [](start = 7, length = 86)

Does HC_TRACE_WARNING rely on the global state?

Source/Common/utils.cpp
```diff
@@ -82,16 +82,20 @@ HRESULT RunAsync(
     uint64_t delayInMs
 )
 {
-    auto workPointer = new AsyncWork(std::move(work));
+    auto workPointer = Make<AsyncWork>(std::move(work));
```
lubeltra: The pattern I would suggest for all this kind of code:
```C++
// start with the async block and/or context being owned by a smart pointer
// if you allocate a context object, just put the async block inside it, it saves you an allocation
std::unique_ptr<XAsyncBlock> async = Make<XAsyncBlock>();
if (!async) { return E_OUTOFMEMORY; }
async->callback = [](XAsyncBlock* asyncPtr)
{
    std::unique_ptr<XAsyncBloc> async{ asyncPtr };
    // do callback
};

// blah blah more setup

HRESULT hr = XAsyncBegin(
    async.get(), // don't release yet! we could fail still
    contextPtrIfAny,
    stuff,
    "stuff",
    [](XAsyncOp op, const XAsyncProviderData* data)
    {
        switch (op)
        {
            case XAsyncOp::Begin:
                return XAsyncSchedule(data->async); // in this case you need to know the delay, you can pass it in through your context object
            case XAsyncOp::DoWork:
                // do work
        }
});
if (FAILED(hr)) { return hr; }
async.release(); // at this point we know do work will be called eventually

Comment (lubeltra) [reply Source/Common/utils.cpp]: You'll want a typedef of std::unique_ptr with a custom deleter for this

File: Source/HTTP/httpcall.cpp (hunk 1 above)
Comment (jasonsandlin) [reply Source/HTTP/httpcall.cpp]: No, it uses

TraceState& GetTraceState() noexcept
{
    static TraceState state;
    return state;
}

and bails out if it that's not setup

File: Source/Common/utils.cpp [hunk 3]
```diff
@@ -82,25 +82,25 @@ HRESULT RunAsync(
...
 
     asyncOp->queue = queue;
-    asyncOp->context = workPointer;
+    asyncOp->context = workPointer.get();
     asyncOp->callback = [](XAsyncBlock* asyncOp)
     {
-        auto context = static_cast<AsyncWork*>(asyncOp->context);
-        Delete(context);
-        Delete(asyncOp);
+        HC_UNIQUE_PTR<AsyncWork> context{ reinterpret_cast<AsyncWork*>(asyncOp->context) };
```
lubeltra: `static_cast` should be sufficient

Source/HTTP/httpcall.cpp
```diff
@@ -401,57 +405,64 @@ void retry_http_call_until_done(
         }
     }
 
+    XAsyncBlock* nestedBlock = Make<XAsyncBlock>();
```
lubeltra: you should do the same changes here

jasonsandlin: done

PR #479, opened 2019-10-24
Update iOS Xcode Build to Succeed on Clean Build

sawalkerXbox:
When building Xal and given the case when building for the first time and building libssl and libcrypto for the first time, the build fails due to the following error:

virtual filesystem overlay file '.../DerivedData/<Xal_Build>/Build/Intermediates.noindex/libHttpClient.build/Release-iphonesimulator/libHttpClient_iOS.build/all-product-headers.yaml' not found

The error is caused because the ssl build script deletes previously written files to `libHttpClient_iOS.build/` where the yaml file exists. This PR moves the ssl script to a new target that is now a dependency of libHttpClient target.


Build/libHttpClient.Apple.C/Dummy/Dummy.h
```diff
@@ -0,0 +1,12 @@
+//
+//  Dummy.h
```
jasonsandlin: from your name from dummy.h & dummy.m.

also can you give the PR a good title?  Its not clear what your trying to do

sawalkerXbox: Hey Jason. I'll clean this up. Matt asked me to create a pr that seeks to solve a problem that I have now added a description of.

jasonsandlin: Ah, cool.  Thanks

PR #482, opened 2019-11-04
android x86_64 build configuration changes

virtuallyjeff-ms:
configuration dependencies and Target API changes to support x86_64 for Android


Build/libHttpClient.141.Android.Java/app/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: 'com.android.model.library'
 
 model {
     android {
-        compileSdkVersion = 19
+        compileSdkVersion = 21
```
jplafonta: >compileSdkVersion = 21 [](start = 8, length = 22)

what's the reason for changing the sdk version?

virtuallyjeff-ms: it's because Android devices didn't support 64-but before 21.  however, thanks for the comment because it also caused me to go back and see that I still have minSdkVersion set to 19, so I should fix that.

PR #490, opened 2020-02-10
Fix lifetime issues with http_singleton

jplafonta:
*Fix lifetime issues with http_singleton - it will only be cleaned up when HCCleanup is called.
*Fix a couple of minor bugs in HCWebsocket
*Repair broken unit tests (unrelated to bug)

### DIFF
--- a/Source/Global/global.cpp
+++ b/Source/Global/global.cpp
@@ -104,6 +104,10 @@ HRESULT http_singleton::cleanup_async(
         {
         case XAsyncOp::Begin:
         {
+            return XAsyncSchedule(data->async, 0);
+        }
+        case XAsyncOp::DoWork:
+        {
             auto singleton{ static_cast<http_singleton*>(data->context)->shared_from_this() };
             singleton->m_self.reset();
 
--- a/Source/Global/global.cpp
+++ b/Source/Global/global.cpp
@@ -108,22 +108,21 @@ HRESULT http_singleton::cleanup_async(
         }
         case XAsyncOp::DoWork:
         {
-            auto singleton{ static_cast<http_singleton*>(data->context)->shared_from_this() };
-            singleton->m_self.reset();
-
-            shared_ptr_cache::cleanup(singleton);
+            auto& self{ static_cast<http_singleton*>(data->context)->m_self };
 
             // Wait for all other references to the singleton to go away
             // Note that the use count check here is only valid because we never create
             // a weak_ptr to the singleton. If we did that could cause the use count
             // to increase even though we are the only strong reference
-            while (singleton.use_count() > 1)
+            if (self.use_count() > 1)
             {
-                std::this_thread::sleep_for(std::chrono::milliseconds{ 10 });
+                return XAsyncSchedule(data->async, 10);
             }
 
-            // Destroy singleton on this thread
-            singleton.reset();
+            shared_ptr_cache::cleanup(self);
+
+            // self is the only reference at this point, singleton will be destroyed on this thread.
+            self.reset();
 
             XAsyncComplete(data->async, S_OK, 0);
             return S_OK;
--- a/Source/Global/global.cpp
+++ b/Source/Global/global.cpp
@@ -17,13 +17,14 @@ NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
 
 enum class singleton_access_mode
 {
+    create,
     get,
-    set
+    cleanup
 };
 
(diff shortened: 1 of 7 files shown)


Source/Global/global.cpp
```diff
@@ -13,10 +13,129 @@
...
+            auto singleton{ static_cast<http_singleton*>(data->context)->shared_from_this() };
+            singleton->m_self.reset();
+
+            shared_ptr_cache::cleanup(singleton);
+
+            // Wait for all other references to the singleton to go away
+            // Note that the use count check here is only valid because we never create
+            // a weak_ptr to the singleton. If we did that could cause the use count
+            // to increase even though we are the only strong reference
+            while (singleton.use_count() > 1)
```
lubeltra: uh... I don't know why I didn't notice this back when we did the Xsapi change, but this could cause deadlocks unless we are 100% sure no other work item produced by libhttpclient will ever schedule itself twice

I think the simplest fix would be to, instead of doing a busy loop, to call `XAsyncSchedule` again with a small delay, therefore freeing the queue to process other work

lubeltra: If you take this suggestion, you'll have to move the `m_self.reset` until after all the uses are gone

Source/Global/global.cpp
```diff
@@ -13,10 +13,129 @@
...
+    {
+        switch (op)
+        {
+        case XAsyncOp::Begin:
+        {
+            return XAsyncSchedule(data->async, 0);
+        }
+        case XAsyncOp::DoWork:
+        {
+            auto singleton{ static_cast<http_singleton*>(data->context)->shared_from_this() };
```
lubeltra: good news! you don't need `shared_from_this` for this to work, you could just copy `m_self` instead

jplafonta: oh good point

---
In reply to: [377389426](https://github.com/microsoft/libHttpClient/pull/490#discussion_r377389426) [](ancestors = 377389426)

Source/Global/global.cpp
```diff
@@ -13,10 +13,129 @@
...
+    {
+        auto rawSingleton = new (http_memory::mem_alloc(sizeof(http_singleton))) http_singleton(
+            GetUserHttpPerformHandler(),
+#if !HC_NOWEBSOCKETS
+            GetUserWebSocketPerformHandlers(),
+#endif
+            std::move(performEnv)
+            );
+
+        auto singleton = std::shared_ptr<http_singleton>(rawSingleton, http_alloc_deleter<http_singleton>());
```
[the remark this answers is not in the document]
jplafonta: I'd rather keep the ctor private to make it more obvious that singleton should only be created with http_singleton::create

---
In reply to: [377386340](https://github.com/microsoft/libHttpClient/pull/490#discussion_r377386340) [](ancestors = 377386340)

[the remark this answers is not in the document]
lubeltra: that makes sense

Source/Global/global.cpp
```diff
@@ -13,10 +13,129 @@
...
+}
+
+HRESULT http_singleton::create(
+    _In_ HCInitArgs* args
+) noexcept
+{
+    if (singleton_access(singleton_access_mode::get))
+    {
+        return E_HC_ALREADY_INITIALISED;
+    }
```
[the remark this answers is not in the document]
jplafonta: do you have a suggestion on how to fix this? I could add a second static lock in http_singleton::create or make the logic in singleton_access a bit more complicated

---
In reply to: [377386531](https://github.com/microsoft/libHttpClient/pull/490#discussion_r377386531) [](ancestors = 377386531)

Source/Global/global.cpp
```diff
@@ -108,22 +108,21 @@ HRESULT http_singleton::cleanup_async(
...
             {
-                std::this_thread::sleep_for(std::chrono::milliseconds{ 10 });
+                return XAsyncSchedule(data->async, 10);
             }
 
-            // Destroy singleton on this thread
-            singleton.reset();
+            shared_ptr_cache::cleanup(self);
+
+            // self is the only reference at this point, singleton will be destroyed on this thread.
```
lubeltra: "the singleton"?

Source/Global/global.cpp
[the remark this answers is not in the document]
lubeltra: the way I resolve this in Xal is that the access function has a create mode, which does the construction inside that lock

PR #491, opened 2020-02-11
Fixing bug with http platform can but initialized multiple times

jplafonta:


Source/Global/global.cpp
```diff
@@ -22,9 +22,10 @@ enum class singleton_access_mode
     cleanup
 };
 
-std::shared_ptr<http_singleton> singleton_access(
+HRESULT singleton_access(
     _In_ singleton_access_mode mode,
-    _In_opt_ PerformEnv createArgs = {}
+    _Out_ std::shared_ptr<http_singleton>& singleton,
```
lubeltra: usually `_Out_` params go last, but if you are really set on having createArg be optional, meh

Source/Global/global.cpp
```diff
@@ -77,26 +82,16 @@ HRESULT http_singleton::create(
...
     _In_ XAsyncBlock* async
 ) noexcept
 {
-    auto singleton{ singleton_access(singleton_access_mode::cleanup) };
-    if (!singleton)
-    {
-        return E_HC_NETWORK_NOT_INITIALIZED;
-    }
+    std::shared_ptr<http_singleton> singleton{};
+    RETURN_IF_FAILED(singleton_access(singleton_access_mode::cleanup, singleton));
```
lubeltra: singleton access does not fail right now in a double cleanup scenario

Source/Global/global.cpp
```diff
@@ -165,7 +160,13 @@ http_singleton::~http_singleton()
 
 std::shared_ptr<http_singleton> get_http_singleton()
 {
-    return singleton_access(singleton_access_mode::get);
+    std::shared_ptr<http_singleton> singleton{};
+    auto hr = singleton_access(singleton_access_mode::get, singleton);
```
lubeltra: you aren't ever failing in this scenario either

PR #492, opened 2020-02-12
Make libHttpClient initialization refCounted

jplafonta:
Make libHttpClient initialization refCounted

### DIFF
--- a/Source/Global/global.cpp
+++ b/Source/Global/global.cpp
@@ -15,13 +15,6 @@ using namespace xbox::httpclient;
 
 NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
 
-enum singleton_access_mode
-{
-    create,
-    get,
-    cleanup
-};
-
 HRESULT http_singleton::singleton_access(
     _In_ singleton_access_mode mode,
     _In_opt_ HCInitArgs* createArgs,
@@ -38,7 +31,7 @@ HRESULT http_singleton::singleton_access(
     case singleton_access_mode::create:
     {
         // Create the singleton only for the first client calling create
-        if (!s_useCount++)
+        if (!s_useCount)
         {
             PerformEnv performEnv;
             RETURN_IF_FAILED(Internal_InitializeHttpPlatform(createArgs, performEnv));
@@ -55,7 +48,9 @@ HRESULT http_singleton::singleton_access(
             s_singleton->m_self = s_singleton;
         }
 
+        ++s_useCount;
         singleton = s_singleton;
+
         return S_OK;
     }
     case singleton_access_mode::get:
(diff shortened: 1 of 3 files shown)


Source/Global/global.cpp
```diff
@@ -15,42 +15,45 @@ using namespace xbox::httpclient;
 
 NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
 
-enum class singleton_access_mode
+enum singleton_access_mode
```
lubeltra: why drop the class?

Source/Global/global.cpp
```diff
@@ -15,42 +15,45 @@ using namespace xbox::httpclient;
...
+    static uint8_t s_useCount{ 0 };
 
     std::lock_guard<std::mutex> lock{ s_mutex };
     switch (mode)
     {
     case singleton_access_mode::create:
     {
-        if (s_singleton)
+        // Create the singleton only for the first client calling create
+        if (!s_useCount++)
```
lubeltra: could you write it as
```c++
if (s_useCount == 0)
{
    ...
}
++s_useCount;

Comment (lubeltra) [reply Source/Global/global.cpp]: In fact right now you potentially increment the use count before creating the singleton, which could lead to a situation where we have use count > 0 but no object

File: Source/Global/global.cpp [hunk 3]
```diff
@@ -67,12 +70,15 @@ HRESULT singleton_access(
 
         if (!s_singleton)
         {
+            assert(!s_useCount);
```
lubeltra: s_useCount == 0

Source/Global/global.cpp
```diff
@@ -67,12 +70,15 @@ HRESULT singleton_access(
...
             return E_HC_NOT_INITIALISED;
         }
 
-        singleton = std::move(s_singleton);
-        s_singleton.reset();
-
+        singleton = s_singleton;
+        if (!--s_useCount)
```
lubeltra:
```c++
--s_useCount
if (s_useCount == 0)
{
    ...
```

Source/Global/global.cpp
```diff
@@ -67,12 +70,15 @@ HRESULT singleton_access(
...
+        {
+            s_singleton.reset();
+        }
         return S_OK;
```
lubeltra: you need to return an error if `s_useCount > 0` otherwise `cleanup_async` will try to run the rest of the cleanup

Source/Global/global.cpp
```diff
@@ -70,14 +65,17 @@ HRESULT http_singleton::singleton_access(
...
         }
 
+        --s_useCount;
         singleton = s_singleton;
-        if (!--s_useCount)
+
+        if (s_useCount == 0)
         {
             s_singleton.reset();
+            return E_HC_INTERNAL_STILLINUSE;
```
lubeltra: wrong way round?

jplafonta: oh good call

---
In reply to: [377989551](https://github.com/microsoft/libHttpClient/pull/492#discussion_r377989551) [](ancestors = 377989551)

PR #494, opened 2020-02-14
Adding missing platform to config.h

rakhad:


libHttpClient.props
```diff
@@ -15,12 +15,14 @@
       UWP
       Win32
       XDK
+      GXDK
```
jplafonta: >GXDK [](start = 6, length = 4)

GXDK or GSDK? seems weird that its GXDK here but the macro is HC_PLATFORM_GSDK

rakhad: I agree but this would be a breaking change at the xsapi level. Not sure if this could lead to broken titles.

---
In reply to: [379604712](https://github.com/microsoft/libHttpClient/pull/494#discussion_r379604712) [](ancestors = 379604712)

PR #495, opened 2020-02-27
Move task queue over to new LocklessQueue

brianpepin:
This change replaces LocklessList.h with LocklessQueue.h.  LocklessQueue performs much better than LocklessList and allows new items to be added to a task queue with zero allocations. Overall performance of task queue inserts and removals is about 6x faster.


Source/Task/LocklessQueue.h
```diff
@@ -0,0 +1,740 @@
...
+        Node* n = to_node(a);
+        m_heap.free(n, a);
+    }
+    
+    //
+    // Pushes the given data onto the back
+    // of the queue.  A copy is made of TData.
+    // If the push fails this returns false.
+    //
+    bool push_back(_In_ TData data) noexcept
```
lubeltra: why not take TData by && to potentially avoid the copy? (it's now the caller responsibility to decide)

Source/Task/LocklessQueue.h
```diff
@@ -0,0 +1,740 @@
...
+     *
+     */
+    
+    // Address - This represents the address of a node.
+    // Nodes live in a contiguous memory block, and there are multiple
+    // blocks. Address represents the position of the
+    // node and must be 64 bits.
+    struct Address
+    {
+        union
```
lubeltra: type punning through a union is technically undefined behavior (breaks strict aliasing I think)

lubeltra: an alternative would be to have accessors for index/block/aba that do the appropriate bit masking and shifting

Source/Task/LocklessQueue.h
```diff
@@ -0,0 +1,740 @@
...
+            if (mem == nullptr)
+            {
+                return false;
+            }
+            
+            Block* block = new (mem) Block;
+
+            block->id = blockId;        
+            block->next = nullptr;
+            block->nodes = new (block + 1) Node[m_blockSize];
```
lubeltra: where are the nodes destructed?

Source/Task/LocklessQueue.h
```diff
@@ -0,0 +1,740 @@
...
+    // Node - each entry in a list is backed by a node,
+    // which contains a single 64 bit value representing
+    // the next node of a singly linked list and
+    // a payload of type TData. Nodes must be properly aligned
+    // in memory so std::atomic works consistently.  We do this
+    // by using an aligned allocator.
+    struct Node
+    {
+        std::atomic<Address> next;
+        TData data;
```
lubeltra: should we only construct TData when data is actually pushed into the node and destroy it when it's popped? (conceptually like vector does it)

Build/libHttpClient.UnitTest.142.TAEF/libHttpClient.UnitTest.142.TAEF.vcxproj
```diff
@@ -20,7 +20,7 @@
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectName>libHttpClient.UnitTest.142.TAEF</ProjectName>
-    <ProjectGuid>{EBFE0314-0869-46B8-ABCE-A3EE223893B8}</ProjectGuid>
+    <ProjectGuid>{9E0EB8C2-40CD-448F-9B98-DAF9830EF9AD}</ProjectGuid>
```
lubeltra: uh oh... this is not meant to happen

Build/libHttpClient.142.XDK.C/libHttpClient.142.XDK.C.vcxproj
```diff
@@ -12,7 +12,7 @@
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ApplicationEnvironment>title</ApplicationEnvironment>
-    <ProjectGuid>{9E0EB8C2-40CD-448F-9B98-DAF9830EF9AD}</ProjectGuid>
+    <ProjectGuid>{57C681F7-280F-45B8-9FA9-D556C24F2053}</ProjectGuid>
```
lubeltra: guids changing is not a good sign, have the templates gone out of sync again? are these guids copy pasted from somewhere else?

Build/libHttpClient.142.Android.C/libHttpClient.142.Android.C.vcxproj
```diff
@@ -35,7 +35,7 @@
     </ProjectConfiguration>
   </ItemGroup>
   <PropertyGroup Label="Globals">
-    <ProjectGuid>{5e0ce391-1ac5-4930-921e-2577a4b5c530}</ProjectGuid>
+    <ProjectGuid>{80FE2440-4F89-4A47-88C6-E86447BC2319}</ProjectGuid>
```
lubeltra: yet another one

Source/Task/TaskQueue.cpp
```diff
@@ -301,36 +314,35 @@ HRESULT __stdcall TaskQueuePortImpl::QueueItem(
...
+    entry.callbackContext = callbackContext;
+    entry.waitRegistration = nullptr;
+    entry.id = m_nextId++;
 
     if (waitMs == 0)
     {
-        entry->enqueueTime = 0;
-        RETURN_HR_IF_FALSE(E_OUTOFMEMORY, AppendEntry(entry.get()));
+        entry.enqueueTime = 0;
+        RETURN_HR_IF(E_OUTOFMEMORY, !AppendEntry(entry));
```
lubeltra: why not RETURN_HR_IF_FALSE?

Source/Task/LocklessQueue.h
```diff
@@ -0,0 +1,740 @@
...
+ copy, and popping data copies that same data out to your variable.
+ 
+ An initial block size is created up front and the size of the block can be customized
+ via the constructor.  If the queue runs out of space, it can create new blocks.  This process
+ is also lock / wait free, although it has an increassed "push" cost as the new block
+ is allocated and initialized. If there is not enough memory to create a new block
+ push_back returns false.
+ 
+ N.B. LocklessQueue MUST be either dynamically allocated or stack allocated as a
+ local variable.  It cannot be used as a member variable in a class or structure
```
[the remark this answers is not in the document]
brianpepin: That wasn't my experience with the compiler Windows is using.  If I declared as a member of a class that had different alignment it wouldn't follow through and the alignment was off.  Maybe compiler bug?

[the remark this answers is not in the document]
lubeltra: Thinking about it, how would the compiler preserve alignment? It would have to know how offset every instance is and magically change the padding...

Ignore me

Source/Task/TaskQueue.cpp
brianpepin: Some branches in Windows are consolidating these macros.

jasonsandlin:
Is this changing the DirectXTK submodule?  Seems like it shouldn't need to

Build/libHttpClient.142.XDK.C/libHttpClient.142.XDK.C.vcxproj
jasonsandlin: agreed.  template checked in should be keeping it unchanged:

C:\git\forks\libHttpClient>findstr /sip 9E0EB8C2 *.vcxproj
Utilities\CMake\template-libHttpClient.142.XDK.C.vcxproj:    <ProjectGuid>{9E0EB8C2-40CD-448F-9B98-DAF9830EF9AD}</Projec

---
In reply to: [385430849](https://github.com/microsoft/libHttpClient/pull/495#discussion_r385430849) [](ancestors = 385430849)

Source/Task/LocklessQueue.h
brianpepin: Would be slower.  I've read a little more on this. I get why this is bad for complex types, and ambiguous for passing pointers around.  But all the "cleaner" ways of doing this I've seen aren't...cleaner.  Best I've seen is a using memcpy to blit the data out.  Looks terrible in code but apparently memcpy is so well optimized by modern compilers that it folds down to a single instruction if the data size fits in a register.

Source/Task/LocklessQueue.h
brianpepin: There are definite advantages to that but I think not right now.  Your comment above about not calling dtors is important.  I've fixed that.  Demand constructing nodes would require that the block objects keep track of which nodes are allocated so they know which to destruct, or that block destruction be coordinated with list contents (which gets hard since lists can share blocks).

Source/Task/LocklessQueue.h
brianpepin: Great catch.

Build/libHttpClient.142.Android.C/libHttpClient.142.Android.C.vcxproj
```diff
@@ -46,7 +46,6 @@
     <ApplicationTypeRevision>3.0</ApplicationTypeRevision>
     <HCLibPlatformType>Android</HCLibPlatformType>
     <HCLibImpl>true</HCLibImpl>
-	<AndroidSdkBuildToolsVersion>25.0.3</AndroidSdkBuildToolsVersion>
```
brianpepin: Why did this get nuked?

Source/Task/LocklessQueue.h
lubeltra: That's fair

Source/Task/LocklessQueue.h
```diff
@@ -272,27 +291,48 @@ class alignas(8) LocklessQueue
...
+            return v != ov;
+        }
+
+        inline operator uint64_t () const
+        {
+            C_ASSERT(sizeof(Address) == sizeof(uint64_t));
+            uint64_t v;
+
+            // Note: this looks horribly inefficient.  General consensus
+            // is this is the best way of doing type puning in a c++ compliant
```
lubeltra: typo

Source/Task/LocklessQueue.h
brianpepin: Provided an overload below.

brianpepin:
Moving this to a new branch that doesn't have the GUID project changes and DirectXTK changes.

PR #502, opened 2020-03-16
Fixing HCHttpCallResponseSetResponseBodyBytes logs bogus bodySize

jasonsandlin:


Source/HTTP/httpcall_request.cpp
```diff
@@ -77,7 +77,7 @@ try
     call->requestBodyBytes.assign(requestBodyBytes, requestBodyBytes + requestBodySize);
     call->requestBodyString.clear();
 
-    if (call->traceCall) { HC_TRACE_INFORMATION(HTTPCLIENT, "HCHttpCallRequestSetRequestBodyBytes [ID %llu]: requestBodySize=%lu", call->id, requestBodySize); }
+    if (call->traceCall) { HC_TRACE_INFORMATION(HTTPCLIENT, "HCHttpCallRequestSetRequestBodyBytes [ID %llu]: requestBodySize=%zu", call->id, requestBodySize); }
```
chwarr: Isn't the `requestBodySize` parameter here a `uint32_t`? There's [`PRIu32` from `cinttypes`](https://en.cppreference.com/w/cpp/types/integer) for that.

jasonsandlin: Ah good point.  I'll switch to %lu which should be good

chwarr: Doesn't `%lu` mean `long unsigned int`? That's 4 bytes for LLP64 but 8 for LP64.

jasonsandlin: Oops, I read that %lu works but I missed that you also need to cast to unsigned long.  I'll do that since I don't want to pull in more headers for such a minor fix

chwarr: Yeah, a cast'll do too.

PR #503, opened 2020-03-19
Allow overriding LIB_OUTPUT env var in openssl.bash for iOS builds

sashaweiss:
Allow setting `OPENSSL_LIB_OUTPUT`and `OPENSSL_TMP_DIR` to override the build locations for libssl/libcrypto.

This change will result in OpenSSL being built under the `DerivedData` folder alongside the other files built as part of the iOS project. Currently, since builds all share the same destination directory (`Binaries/iOS/lib`), building once for `i386+x86_64` (i.e., for Simulator) and then later for `ARM+ARM64` (i.e., for device) will overwrite each other, forcing a recompile if you switch back to sim.


sashaweiss:
@virtuallyjeff-ms made some non-trivial changes. Could I get you to look at it again?

PR #504, opened 2020-03-19
More %llu fixes

jasonsandlin:
Problem discussed in full here:
https://stackoverflow.com/questions/14073877/printf-and-llu-vs-lu-on-os-x

You can either use PRIu64 macro or cast to unsigned long long to solve.

### DIFF
--- a/Source/Common/pch_common.h
+++ b/Source/Common/pch_common.h
@@ -128,6 +128,7 @@ HC_DECLARE_TRACE_AREA(WEBSOCKET);
     catch (...) { ::xbox::httpclient::detail::UnknownExceptionToResult(file, line); return errCode; }
 
 #define RETURN_IF_PERFORM_CALLED(call) if (call->performCalled) return E_HC_PERFORM_ALREADY_CALLED;
+#define TO_ULL(x) static_cast<unsigned long long>(x)
 
 NAMESPACE_XBOX_HTTP_CLIENT_DETAIL_BEGIN
 
--- a/Source/HTTP/httpcall.cpp
+++ b/Source/HTTP/httpcall.cpp
@@ -49,7 +49,7 @@ try
     call->retryIterationNumber = 0;
     call->id = ++httpSingleton->m_lastId;
 
-    HC_TRACE_INFORMATION(HTTPCLIENT, "HCHttpCallCreate [ID %llu]", call->id);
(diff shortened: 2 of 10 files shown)


Source/Common/pch_common.h
```diff
@@ -128,6 +128,7 @@ HC_DECLARE_TRACE_AREA(WEBSOCKET);
     catch (...) { ::xbox::httpclient::detail::UnknownExceptionToResult(file, line); return errCode; }
 
 #define RETURN_IF_PERFORM_CALLED(call) if (call->performCalled) return E_HC_PERFORM_ALREADY_CALLED;
+#define TO_ULL(x) static_cast<unsigned long long>(x)
```
chwarr: Could use a templated function instead of a macro if you want. Adjust the `enable_if` as needed. (There's probably an even better way to write the condition. I was going for "no truncation".)

```suggestion
template <typename T>
typename std::enable_if<
    std::is_unsigned<T>::value && sizeof(T) <= sizeof(unsigned long long),
    unsigned long long>::type
ToUll(const T val)
{
    return val;
}
```

Source/HTTP/httpcall.cpp
```diff
@@ -496,7 +496,7 @@ try
         return E_INVALIDARG;
     }
 
-    if (call->traceCall) { HC_TRACE_INFORMATION(HTTPCLIENT, "HCHttpCallPerform [ID %llu] uri: %s", call->id, call->url.c_str()); }
+    if (call->traceCall) { HC_TRACE_INFORMATION(HTTPCLIENT, "HCHttpCallPerform [ID %llu] uri: %s", TO_ULL(call->id), call->url.c_str()); }
```
jplafonta: > TO_ULL( [](start = 98, length = 8)

what is the impact if you don't include these casts?

jasonsandlin: It causes trouble on some platforms as discussed on this thread: https://stackoverflow.com/questions/14073877/printf-and-llu-vs-lu-on-os-x
but probably hard to see in practice since the IDs start off small

Source/Common/pch_common.h
jasonsandlin: There shouldn't be truncation as discussed here: https://stackoverflow.com/questions/14073877/printf-and-llu-vs-lu-on-os-x cause on some platforms its unsigned long long which is fine, and some platforms its unsigned long, so casting to unsigned long long is fine. 

I'd prefer not to add complex templates for a small issue as this.

PR #505, opened 2020-03-26
Fix corner case error handling

brianpepin:
This change fixes a corner case where we could orphan a call if an async call was started after a queue was terminated.  It also fixes an API signature problem with XAsyncBeginAlloc -- the recommended way to use these apis is to call XAsyncSchedule in your provider during XAsyncOp::Begin, but you can't do that if you were never able to initialize your context object with the call parameters. This adds an initContext parameter to both the XAsyncBeginAlloc call and to the provider data.

### DIFF
--- a/Source/Common/ResultMacros.h
+++ b/Source/Common/ResultMacros.h
@@ -18,3 +18,5 @@
     HC_TRACE_ERROR(HTTPCLIENT, fmt, ##__VA_ARGS__);    \
     ASSERT(false);                                     \
 
+
+#define FAIL_FAST_IF_FAILED(hr)                                 do { HRESULT __hrRet = hr; if (FAILED(__hrRet)) { FAIL_FAST_MSG("%s 0x%08", #hr, __hrRet); }} while (0, 0)
--- a/Tests/UnitTests/Tests/AsyncBlockTests.cpp
+++ b/Tests/UnitTests/Tests/AsyncBlockTests.cpp
@@ -4,6 +4,7 @@
 #include "UnitTestIncludes.h"
 #include "XAsync.h"
 #include "XAsyncProvider.h"
+#include "XAsyncProviderPriv.h"
 #include "XTaskQueue.h"
 #include "XTaskQueuePriv.h"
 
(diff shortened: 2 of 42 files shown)


Tests/UnitTests/Tests/AsyncBlockTests.cpp
```diff
@@ -987,4 +999,82 @@ DEFINE_TEST_CLASS(AsyncBlockTests)
...
+            return S_OK;
+        };
+
+        // Terminate the queue
+        XTaskQueueTerminate(async.queue, true, nullptr, nullptr);
+
+        bool callbackCalled = false;
+        async.context = &callbackCalled;
+
+        // XAsyncBegin should succeed.  The callback should have been called anyway, even
```
lubeltra: does that mean we're potentially calling callback:
- from inside the XAsyncBegin callstack
- on the wrong thread

Both of these scare me, because now I have to write code to account for them happening all the time :(

The first one is especially bad if I have non recursive locks I'm taking

brianpepin: Yes, it does.  The alternative, if we hit this case, is that we never call your callback and your task never completes.

brianpepin: After much discussion we decided not to do this but to instead early check for termination and, if we later on get surprised with failing to submit a completion, ,we failfast the process.

Source/Common/ResultMacros.h
```diff
@@ -18,3 +18,5 @@
     HC_TRACE_ERROR(HTTPCLIENT, fmt, ##__VA_ARGS__);    \
     ASSERT(false);                                     \
 
+
+#define FAIL_FAST_IF_FAILED(hr)                                 do { HRESULT __hrRet = hr; if (FAILED(__hrRet)) { FAIL_FAST_MSG("%s 0x%08", #hr, __hrRet); }} while (0, 0)
```
lubeltra: is this macro defined on all platforms?

brianpepin: Looks to be. brought in through pch_common.

PR #506, opened 2020-04-06
Add explicit platform macros for Nintendo Switch and Playstation 4

jplafonta:


Include/httpClient/config.h
```diff
@@ -9,6 +9,8 @@
 #define HC_PLATFORM_ANDROID 11
 #define HC_PLATFORM_IOS 21
 #define HC_PLATFORM_MAC 22
+#define HC_PLATFORM_NINTENDO_SWITCH 31
```
lubeltra, line 12: I would put them up at 111 and 121, to logically group with external

Include/httpClient/config.h
```diff
@@ -99,6 +101,11 @@
 (HC_PLATFORM == HC_PLATFORM_MAC || HC_PLATFORM == HC_PLATFORM_IOS)
 #endif
 
+#if !defined(HC_PLATFORM_IS_EXTERNAL)
```
lubeltra, line 104: can you add a comment describing what we mean by external here? just to avoid confusion in the future

PR #510, opened 2020-04-10
Fixing platform name

jasonsandlin:


Include/httpClient/config.h
```diff
@@ -5,7 +5,8 @@
 #define HC_PLATFORM_WIN32 1
 #define HC_PLATFORM_UWP 2
 #define HC_PLATFORM_XDK 3
-#define HC_PLATFORM_GSDK 4
+#define HC_PLATFORM_GSDK 4 // For backcompat reasons
```
lubeltra, line 8: instead 
```c++
#define HC_PLATFORM_GSDK HC_PLATFORM_GDK
```
just after this list?

Issue #516, opened 2020-05-11
HTTPS Struggles


Cry-Luca:
I'd like to point out two limitations in the library I found while trying to deal with self-signed certificates we sometimes use for development purposes.

First of all there seem to be no way to disable SSL peer verification. I totally understand the security implications, but sometimes it's fine to accept "bad" certificates from hosts which we know. Other HTTP libraries I know expose that option (e.g. CURLOPT_SSL_VERIFYPEER), so it only makes sense to me that libHttpClient exposes it as well.

Another related limitation is that the most information the library provides about HTTPS failure (on Windows) is a WINHTTP error code. The one I was always getting is extremely generic (ERROR_WINHTTP_SECURE_FAILURE) and the only way to get more information is to use a WINHTTP_STATUS_CALLBACK. This callback slot is already used by the library and even if I wanted to hook my own callback, I couldn't find a way to retrieve the HINTERNET required to set it up.

jasonsandlin:
Feel free to expose an option to disable SSL verification if you need that and submit a pull request.  

For the error code issue, we can't expose HINTERNET as non-Windows platforms won't know what that is.  But what logic do you want to see in WINHTTP_STATUS_CALLBACK to pull out extra information?  Is it just for debugging or does your code need to handle different types of HTTPS failures differently?

Cry-Luca:
Sure, I wouldn't have any problem adding it myself. Do you have suggestions about how this option  should be named and where it should be placed?

Regarding the status callback, it would be enough for me to have some tracing logs to make debugging easier.

jasonsandlin:
For the option, I would expose it per http call handle, you can expose it similar to HCHttpCallSetTracing.  
https://github.com/microsoft/libHttpClient/blob/master/Include/httpClient/httpClient.h#L280
You'll see how that sets a bool.  Call it something like HCHttpCallSetSSLValidation and default to true and doc in the summary that calling it is optional to turn off ssl validation.

Feel free to add logging lines to that callback and send a PR request.  
Here's an example:

   HC_TRACE_ERROR(HTTPCLIENT, "HCHttpCallPerform [ID %llu]: failed", TO_ULL(static_cast<HC_CALL*>(call)->id)); }

Then in your game, use
https://github.com/microsoft/libHttpClient/blob/master/Include/httpClient/trace.h#L152
to set the trace level, and use
https://github.com/microsoft/libHttpClient/blob/master/Include/httpClient/trace.h#L189
or
[...]

PR #517, opened 2020-05-12
Add HCHttpCallRequestSetSSLValidation(). Implemented for WinHttp only.

Cry-Luca:
#516


jasonsandlin:
Thanks.  Can you sign the Contributor License Agreement?

Cry-Luca:
Sorry for making you wait, I am unable to until I get the green light from our legal department.

jasonsandlin:
Thanks.  No rush

Cry-Luca:
@jasonsandlin Sorry it took so long, but it should be ready to go now.

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -863,6 +863,20 @@ HRESULT winhttp_http_task::send(
...
+    {
+        DWORD dwOption = SECURITY_FLAG_IGNORE_UNKNOWN_CA | SECURITY_FLAG_IGNORE_CERT_CN_INVALID;
+        if (!WinHttpSetOption(
+            m_hRequest, 
+            WINHTTP_OPTION_SECURITY_FLAGS, 
+            &dwOption, 
+            sizeof(dwOption)))
+        {
+            DWORD dwError = GetLastError();
+            HC_TRACE_WARNING(HTTPCLIENT, "winhttp_http_task [ID %llu] [TID %ul] WinHttpSetOption errorcode %d", HCHttpCallGetId(m_call), GetCurrentThreadId(), dwError);
```
jplafonta, line 876: are you intentionally eating this error here?  Should it be returned?

Include/httpClient/httpClient.h
```diff
@@ -451,6 +451,20 @@ STDAPI HCHttpCallRequestSetTimeoutWindow(
...
+///
+/// Optional, call prior to calling HCHttpCallPerformAsync.
+/// </summary>
+/// <param name="call">The handle of the HTTP call</param>
+/// <param name="traceCall">Validate certificates</param>
+/// <returns>Result code for this API operation. Possible values are S_OK, E_INVALIDARG, or E_FAIL.</returns>
+STDAPI HCHttpCallRequestSetSSLValidation(
+    _In_ HCCallHandle call,
+    _In_ bool sslValidation
+) noexcept;
```
jplafonta, line 467: @ do we want to #ifdef this API to platforms that use winhttp? or just have it non-functional on other platforms?

jasonsandlin: Hmm, yeah #ifdef for now around the public header until this is supported every would avoid complexity of someone using it and it not doing anything.  

the two platforms that use WinHTTP are Win32 & GDK.  

Cry-Luca, Can you wrap this public header and the src impl functions with:

#endif

Source/HTTP/WinHttp/winhttp_http_task.cpp
Cry-Luca: Yes, I didn't feel like an error in setting that option was serious enough to prevent the request from going through. Do you prefer if we return in case of error?

Include/httpClient/httpClient.h
Cry-Luca: Sure will do that.

Source/HTTP/WinHttp/winhttp_http_task.cpp
jasonsandlin: Yeah eating the error is fine for this

jasonsandlin:
Merging.  The build failure doesn't look related to this PR.  Some internal build machine copy step

PR #518, opened 2020-05-26
Add modern Gradle+CMake Android builds

sashaweiss:
Currently, libHttpClient/crypto/ssl are built for Android using Visual Studio `.vcxproj`s, with a `.androidproj` for libHttpClient's Android Java component. This PR adds Android builds for those three libraries using pure Gradle+CMake, the recommended build system for compiling native Android code. These builds enable us to build libHttpClient with more modern versions of the Android NDK (since VS is limited to outdated NDK versions), as well as to develop libHttpClient Android via Android Studio.

Note: this PR does not remove the VS-based Android builds, as we do not want it to be a breaking change (yet).

Adds:
- CMake files for the Android builds of libHttpClient/crypto/ssl at `Utilities/CMake/Android`.
- Gradle builds for libHttpClient, libcrypto, and libssl at `Build/libHttpClient.Android`, `Build/libcrypto.Android`, and `Build/libssl.Android`.
    - These builds rely on the CMake files above.
    - They use Gradle 5.4.1 and Android plugin 3.5.3, both very modern versions.
    - Using Gradle for libcrypto/ssl is not strictly necessary since they don't have a Java component, but it handles setting various CMake build flags (e.g. ANDROID_ABI) that are otherwise a pain to set in a script.


Build/libHttpClient.Android/app/src/main/AndroidManifest.xml.template
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
lubeltra, line 1: what is the template for?

Build/libHttpClient.Android/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
```diff
@@ -0,0 +1,80 @@
+package com.xbox.httpclient;
```
lubeltra, line 1: I'm not sure build is the right place to put source code
is it because the build files have to be in the same folder?
in that case it may be better to have those in the source

Build/libHttpClient.Android/app/src/main/AndroidManifest.xml.template
sashaweiss: That is a vestige of a copy-paste. Good catch!

Build/libHttpClient.Android/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
sashaweiss: Yeah, it's because for a Gradle project it expects the code to be in subdirectories. I also think it's a bit weird, but FWIW the current libHttpClient Java code also lives under `Build`, in `Build/libHttpClient.14(1|2).Android.Java/app`.

So I was mostly just following the convention there.

sashaweiss:
@lubeltra we don't strictly have to have the Gradle files (e.g. `gradlew`, `gradle-wrapper.properties`) multiple times, but the convention is that a Gradle build has the Gradle wrapper etc bundled in it. So, it'd be odd to have one shared copy.

In principle, each of those Gradle builds represents a standalone build that might use its own version of Gradle etc. In practice they won't - but it'd be breaking convention in a fairly confusing way to keep one shared set of Gradle files.

Build/libHttpClient.Android/app/src/main/AndroidManifest.xml.template
sashaweiss: removed

Build/libHttpClient.Android/app/src/main/java/com/xbox/httpClient/HttpClientRequest.java
sashaweiss: It also, stupidly, means that there are now three copies of the Java code for libHC (one in the 141, 142, and this build def) in the repo. Fortunately those basically never change, but it isn't great. Not sure what the best way otherwise to organize it would be though.

lubeltra: I really dislike this, but I guess it's fine for now

virtuallyjeff-ms:
> @lubeltra we don't strictly have to have the Gradle files (e.g. `gradlew`, `gradle-wrapper.properties`) multiple times, but the convention is that a Gradle build has the Gradle wrapper etc bundled in it. So, it'd be odd to have one shared copy.
> 
> In principle, each of those Gradle builds represents a standalone build that might use its own version of Gradle etc. In practice they won't - but it'd be breaking convention in a fairly confusing way to keep one shared set of Gradle files.

are you referring to the fact that EACH project has its own build.gradle, settings.gradle, etc.?  as opposed to somehow sharing these across related projects?

sashaweiss:
@virtuallyjeff-ms 

I think that is what @lubeltra meant, yeah, but I think he might have meant the more boilerplate stuff like `gradlew[.bat]`, `gradle-wrapper.properties`, etc. In any case, I would argue that having the multiple copies of those is good per both the convention (of having a Gradle wrapper in each project) as well as the potential for different projects to use different versions of Gradle.

Am down to bump the versions! Wanted to keep as they are for now, but I can add that to my list of tweaks I want to make.

What do you mean re: the backward compat comment?

virtuallyjeff-ms:
I meant “sorry that you had to reverse those commits”, but the backward compat is important as we discussed.

sashaweiss:
Oh gotcha - no problem! It's the wise thing to do, and hey that's what `git` is for :)

PR #521, opened 2020-05-29
Tweak binary dir for libHttpClient Java

sashaweiss:
Since product flavors are only used to separate native lib ABIs, no need to put the product flavor in the built Java binary name.

### DIFF
--- a/Build/libHttpClient.Android/app/build.gradle
+++ b/Build/libHttpClient.Android/app/build.gradle
@@ -57,7 +57,7 @@ android {
 
     libraryVariants.all { variant ->
         variant.outputs.each { output ->
-            output.outputFileName = "../../../../../../Binaries/Android/java/libHttpClient-${variant.name}.aar"
+            output.outputFileName = "../../../../../../Binaries/Android/java/${variant.getFlavorName()}/libHttpClient-${variant.getBuildType().getName()}.aar"
         }
     }
 }
--- a/Build/libHttpClient.Android/app/build.gradle
+++ b/Build/libHttpClient.Android/app/build.gradle
@@ -56,8 +56,20 @@ android {
     }
 
     libraryVariants.all { variant ->
+        def variantFlavor = variant.getFlavorName()
+        def variantAbi = "unknown"
+        if (variantFlavor == "arm7") {
+            variantAbi = "armeabi-v7a"
+        } else if (variantFlavor == "arm64") {
+            variantAbi = "arm64-v8a"
+        } else if (variantFlavor == "x86") {
+            variantAbi = "x86"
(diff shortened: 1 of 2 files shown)


Build/libHttpClient.Android/app/build.gradle
```diff
@@ -57,7 +57,7 @@ android {
 
     libraryVariants.all { variant ->
         variant.outputs.each { output ->
-            output.outputFileName = "../../../../../../Binaries/Android/java/libHttpClient-${variant.name}.aar"
+            output.outputFileName = "../../../../../../Binaries/Android/java/${variant.getFlavorName()}/libHttpClient-${variant.getBuildType().getName()}.aar"
```
lubeltra, line 60: could we get the aars to be put next to the .so?

sashaweiss: There is no `.so`, you mean the `.a`?

Currently the C++ binaries go into:
```
Binaries/Android/native/<debug|release>/<abi>/libNAME.a
```
and the Java binary goes to:
```
Binaries/Android/java/libHttpClient-<abi><debug|release>.aar
```

Since for the `.aar` the <abi> is meaningless (the fact that it's there is an artifact of how the Gradle files are structured), this change would put it in:
```
Binaries/Android/java/<abi>/libHttpClient-<debug|release>.aar
```

So while they're not directly alongside the `.a`, everything is included under `Binaries/Android`.

sashaweiss: I can align the Java and native lib paths to have the <abi> and <debug|release> in the same order, and/or take the <debug|release> out of the filename for the Java or something (or put it into the filename for the native libs). Open to feedback here.

sashaweiss: per offline conversation, i'll align to have both native and java binaries in:
```
Binaries/Android/<abi>/<debug|release>
```

sashaweiss:
@virtuallyjeff-ms it does, yeah. Before the binary names had the toolset and platform in them, i.e. `libHttpClient.141.Android.a`, since they came straight from the name of the `.vcxproj` used to build them.

In the new build system, I have the binary names being picked up from the CMake PROJECT_NAME, which is (same as it was in the old build system) hardcoded to the same as the directory for the project - namely, `lib(HttpClient|crypto|ssl).Android`. So, the binaries will come out as `libHttpClient.Android.a`, `libcrypto.Android.a`, and `libssl.Android.a`. Happy to add more annotations in the names, that's no problem.

sashaweiss:
To clarify - *this change* does not change the binary names compared to the VS build - the original CMake PR did. This just changes the path that they're placed under.

PR #522, opened 2020-06-04
Jitter Retry-After delay to spread load out

jasonsandlin:


Source/HTTP/httpcall.cpp
```diff
@@ -243,8 +236,14 @@ bool http_call_should_retry(
         std::chrono::milliseconds waitTime = std::chrono::milliseconds(static_cast<int64_t>(secondsToWait * 1000.0));
         if (retryAfter.count() > 0)
         {
-            // Use either the waitTime or Retry-After header, whichever is bigger
-            call->delayBeforeRetry = std::chrono::milliseconds(std::max(waitTime.count(), retryAfter.count()));
+            // Jitter to spread the load of Retry-After out between the devices trying to retry
```
jplafonta, line 239: is the retry time not already jittered above?

jasonsandlin: waitTime is jittered, but the logic  on line 246 was

            call->delayBeforeRetry = std::chrono::milliseconds(std::max(waitTime.count(), retryAfter.count()));	

so it'd always use retryAfter which wasn't jittered.

PR #523, opened 2020-06-09
[Android CMake] Get HC_NOWEBSOCKETS from env, standardize names

sashaweiss:
In the Android Gradle+CMake native build:

- Detect `HC_NOWEBSOCKETS` as an environment variable, and if present set the corresponding build flag. I chose a standalone env var because:
    - The CXX_FLAGS variable is only considered by CMake if this is the first configuration run, and some users may want to sometimes build the same libHttpClient distribution with/without websockets for different purposes, without clearing all CMake configuration.
    - libHttpClient is included as source and built as a dependency of many (most?) projects, each of whom need to be able to control at their build time whether websockets are included or not.
- Add a README explaining this change.

This PR also includes the following standardization changes:
- Move libHttpClient flags to a separate file, as per convention for other projects.
- Prefix some out-variables with `OUT_`, as per convention for other projects.
- Rename `SetLibHC...` files to `GetLibHC...`, as per convention for other files.


sashaweiss:
going to break this into a couple PRs since things changed in the meantime and it got unwieldy.

PR #535, opened 2020-07-24
Adding XNetworking support

jasonsandlin:


Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -362,6 +505,30 @@ void winhttp_http_task::callback_status_request_error(
...
+    _In_ HINTERNET hRequestHandle,
+    _In_ winhttp_http_task* pRequestContext,
+    _In_ void* /*statusInfo*/)
+{
+    if (hRequestHandle != nullptr)
+    {
+        HRESULT hr = XNetworkingVerifyServerCertificate(hRequestHandle, pRequestContext->m_securityInformation);
+        if (FAILED(hr))
+        {
+            pRequestContext->complete_task(hr, hr);
```
jplafonta, line 520: >hr, hr [](start = 43, length = 6)

intentionally "hr" for both error values?

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -30,29 +30,165 @@ void get_proxy_name(
...
+
+    if (!m_globalProxy.empty())
+    {
+        (void)SetGlobalProxyForHSession(hSession, m_globalProxy.c_str());
+    }
+
+    return hSession;
+}
+
+HINTERNET HC_PERFORM_ENV::GetSessionForHttpSecurityProtocolFlags(_In_ uint32_t enabledHttpSecurityProtocolFlags)
```
jplafonta, line 157: >GetSessionForHttpSecurityProtocolFlags [](start = 26, length = 38)

Any reason to split this function and CreateSession above? looks like you always want to create the session if it doesn't exist yet

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -785,11 +962,79 @@ http_internal_wstring flatten_http_headers(_In_ const http_header_map& headers)
...
+    XAsyncBlock asyncBlock{};
+    asyncBlock.queue = m_env->GetImmediateQueue();
+    HRESULT hr = XNetworkingQuerySecurityInformationForUrlUtf16Async(wUrlHost.c_str(), &asyncBlock);
+    if (FAILED(hr))
+    {
+        HC_TRACE_ERROR(HTTPCLIENT, "winhttp_http_task [ID %llu] [TID %ul] XNetworkingQuerySecurityInformationForUrlUtf16Async 0x%0.8x", TO_ULL(HCHttpCallGetId(m_call)), GetCurrentThreadId(), hr);
+        return hr;
+    }
+
+    hr = XAsyncGetStatus(&asyncBlock, true);
```
jplafonta, line 980: >hr = XAsyncGetStatus(&asyncBlock, true); [](start = 4, length = 40)

fine to block here?

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -785,11 +962,79 @@ http_internal_wstring flatten_http_headers(_In_ const http_header_map& headers)
...
+        cUri.Port();
+    http_internal_wstring wUrlHost = utf16_from_utf8(cUri.Host());
+
+#if HC_PLATFORM != HC_PLATFORM_GDK
+    uint32_t enabledHttpSecurityProtocolFlags = HC_PERFORM_ENV::GetDefaultHttpSecurityProtocolFlagsForWin7();
+#else
+    hr = query_security_information(wUrlHost);
+    if (FAILED(hr))
+        return hr;
+    uint32_t enabledHttpSecurityProtocolFlags = m_securityInformation->enabledHttpSecurityProtocolFlags;
```
jplafonta, line 1033: > m_securityInformation->enabledHttpSecurityProtocolFlags [](start = 47, length = 56)

 Minor thing, but reading the code it wasn't immediately clear how this get set. If this is the only field of m_securityInformation you need, maybe just return it as an _Out_ param from query_security_information rather than storing it as a member of the win_http_task?

Source/HTTP/WinHttp/winhttp_http_task.cpp
jasonsandlin: It was just to make it easier to read GetSessionForHttpSecurityProtocolFlags as this function just dealing with the std::map

Source/HTTP/WinHttp/winhttp_http_task.cpp
jasonsandlin: Its actually used in the callback as well in:

            case WINHTTP_CALLBACK_STATUS_SENDING_REQUEST:
            {
                callback_status_sending_request(hRequestHandle, pRequestContext, statusInfo);
                break;
            }
#endif // HC_PLATFORM_GDK

Source/HTTP/WinHttp/winhttp_http_task.cpp
jasonsandlin: Yeah, Doug recommended this approach with immediate queue as he does this approach elsewhere

Source/HTTP/WinHttp/winhttp_http_task.cpp
jasonsandlin: Yeah, I wanted this hr to flow up so its obvious to the caller the cause.  Probably won't happen too often

PR #540, opened 2020-09-25
Fix handling of redirects for winhttp

bbowman:
This change adds in handling of redirect requests for the windows http task. In handling of the redirect, the default behavior is to forward along all request headers from the original request except for authorization. This is important because the redirect should not expose an auth token to a host that different than the one it was originally intended for. A more fully extensible model may wish to allow a caller to customize this behavior in cases that the redirect is to another page on the same host that also requires auth or that the original request blindly redirects without performing auth. This however is a good default as it reduces the security risk and will let a developer see the 401 upfront rather than being surprised with where else secret information like an auth token may be sent.


bbowman:
@jplafonta Is there anything else I should be doing here to get this into reviewable shape? Let me know and I'm happy to adjust / add to this :-)

jasonsandlin:
Thanks for the PR.  It is on our radar.  We'll need to test it to ensure it doesn't break anything before accepting.  It'll probably won't be till end of year until we have cycles to circle back.

bbowman:
@jasonsandlin Sounds good and thanks for the update! Let me know if there are issues here as we have locally patched a project to include this. More than happy to help assist in validation or do anything else you folks might need. :-)

jasonsandlin:
can you retarget this PR to "main" and repo thanks?

PR #541, opened 2020-09-25
Expose async cleanup

lubeltra:
The current blocking version of HCCleanup has 3 major flaws:
- It always uses the system task queue, which may be undesirable by some clients
- It blocks waiting for the internal work to complete, which could cause a deadlock
- It does not expose a result so in the case of multiple libHttpClient initializations it is impossible to tell if the memory hooks are safe to remove

This change simply exposes the internal asynchronous cleanup functionality allowing the caller control to bypass all these issues

### DIFF
--- a/Source/Global/global.cpp
+++ b/Source/Global/global.cpp
@@ -5,6 +5,7 @@
 #include "../HTTP/httpcall.h"
 #include "buildver.h"
 #include "global.h"
+#include "../Logger/trace_internal.h"
 #include "../Mock/lhc_mock.h"
 
 #if !HC_NOWEBSOCKETS
@@ -33,6 +34,8 @@ HRESULT http_singleton::singleton_access(
         // Create the singleton only for the first client calling create
         if (!s_useCount)
         {
+            HCTraceImplInit();
+
             PerformEnv performEnv;
             RETURN_IF_FAILED(Internal_InitializeHttpPlatform(createArgs, performEnv));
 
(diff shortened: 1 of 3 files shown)


Source/Global/global_publics.cpp
```diff
@@ -5,10 +5,11 @@
 #include "../HTTP/httpcall.h"
 #include "buildver.h"
 #include "global.h"
-#include "../Logger/trace_internal.h"
 
 using namespace xbox::httpclient;
 
+XTaskQueueHandle g_HcCleanupHackQueueHandle = nullptr;
+
```
jplafonta, line 12: is this used somewhere?

Source/Global/global.cpp
```diff
@@ -113,7 +116,38 @@ HRESULT http_singleton::cleanup_async(
...
+                return S_OK;
+            }
+            default:
+            {
+                assert(false);
+                return S_OK;
+            }
+            }
+        });
+    }
```
jplafonta, line 150: outside the scope of this change, but seems like there should be an easier way to return a synchronous error to the XAsync callback.  Seems clunky to have to write a provider just to get the error to flow to caller correctly

jplafonta (approved):
:shipit:

Source/Global/global_publics.cpp
lubeltra: No, that's a leftover from some testing

Source/Global/global.cpp
lubeltra: It would be handy, yes

jasonsandlin (approved):

PR #544, opened 2020-10-01
Making hcwebsocket thread safe

rakhad:


Source/WebSocket/hcwebsocket.cpp
```diff
@@ -195,6 +196,7 @@ HRESULT HC_WEBSOCKET::Send(
     {
         try
         {
+            std::lock_guard<std::recursive_mutex> lock{ m_mutex };
```
rakhad, line 199: >td::lock_guard<std::recursive_mutex> lock{ m_mutex }; [](start = 13, length = 53)

Maybe we don't need to lock here since we aren't setting state

rakhad: same below

---
In reply to: [498433665](https://github.com/microsoft/libHttpClient/pull/544#discussion_r498433665) [](ancestors = 498433665)

jasonsandlin: in general i wouldn't hold a lock when calling handlers or a big function if you can help it. best to pin the data under a lock and then get out.  but note that locking isn't needed just on shared writes or reads, but its about locking a block of logic that does if checks so that a set of state doesn't change on you while doing logic against it (if, looping, etc)

---
In reply to: [498433747](https://github.com/microsoft/libHttpClient/pull/544#discussion_r498433747) [](ancestors = 498433747,498433665)

PR #545, opened 2020-10-05
Adding initial state to better manage websocket lifecycle

rakhad:


Source/WebSocket/hcwebsocket.cpp
```diff
@@ -69,12 +69,19 @@ HRESULT HC_WEBSOCKET::Connect(
             {
                 auto thisPtr{ static_cast<HC_WEBSOCKET*>(async->context) };
                 HRESULT hr = HCGetWebSocketConnectResult(async, &thisPtr->m_connectResult);
-                
-                if (SUCCEEDED(hr) && SUCCEEDED(thisPtr->m_connectResult.errorCode))
+
+                bool doDisconnect{ false };
+                bool doDecRef { false };
```
jplafonta, line 74: Could you add comments to these to bools explaining the situations where we should/shouldn't disconnect and decRef? 

i.e. we auto disconnect if the client has already closed all handles to the socket and we release the provider's reference if we don't expect any more callbacks from the WebSocket provider

Source/WebSocket/hcwebsocket.cpp
```diff
@@ -468,28 +482,40 @@ void HC_WEBSOCKET::CloseFunc(
 )
 {
     UNREFERENCED_PARAMETER(context);
+
+    bool shouldNotClose { false };
+
     {
         std::lock_guard<std::recursive_mutex> lock{ websocket->m_mutex };
 
+        shouldNotClose = (websocket->m_state == State::Connecting || websocket->m_state == State::Disconnected);
```
jplafonta, line 491: >shouldNotClose = (websocket->m_state == State::Connecting || websocket->m_state == State::Disconnected); [](start = 8, length = 104)

Did you test this on other platforms?  I assume the expectation is that if connect fails, the provider should complete the connect XAsyncBlock & not call the closeFunc, which seems reasonable.  Are we sure all our providers are actually doing this though?  I think we need to make it 100% clear what the expectations of the provider are is essential as it seems we have some inconsistency across the various providers here

In any case, I think this would read a bit clearer if you just broke this up into separate cases and commented on each:

switch(websocket->m_state)
{
case State::Connecting:
  // provider error: providers should complete connect async context rather than calling closeFunc if a connect attempt fails
  return;
case State::Disconnected:
  // provider error: providers should call closeFunc exactly once per successful attempt
  return;
case State::Initial:
  // ...
 return;
case State::Connected:
case State::Disconnecting:
  // Expected. Invoke client callback and update our state
  ....
}

This would also help us catch any inconsistency in provider implementation more easily.

jplafonta (changes requested):
:clock1:

Source/WebSocket/hcwebsocket.cpp
```diff
@@ -468,28 +485,52 @@ void HC_WEBSOCKET::CloseFunc(
...
 {
     UNREFERENCED_PARAMETER(context);
+
+    // We release the provider's reference if we handle the close.
+    bool shouldDecRef { false };
     {
         std::lock_guard<std::recursive_mutex> lock{ websocket->m_mutex };
 
+        auto state = websocket->m_state;
         websocket->m_state = State::Disconnected;
```
jplafonta, line 495: >websocket->m_state = State::Disconnected; [](start = 8, length = 41)

should this only be in the bottom 2 cases?

jplafonta: I think we shouldn't be transitioning states if we get an unexpected provider callback

---
In reply to: [499870395](https://github.com/microsoft/libHttpClient/pull/545#discussion_r499870395) [](ancestors = 499870395)

rakhad: That was the existing behavior. Moving it would result in a breaking change.
[https://github.com/microsoft/libHttpClient/blob/master/Source/WebSocket/hcwebsocket.cpp#L474](https://github.com/microsoft/libHttpClient/blob/master/Source/WebSocket/hcwebsocket.cpp#L474)

---
In reply to: [499871609](https://github.com/microsoft/libHttpClient/pull/545#discussion_r499871609) [](ancestors = 499871609,499870395)

PR #556, opened 2021-01-06
Add iOS targets for building without websockets/SSL

sashaweiss:
This PR adds the ability to build libHttpClient for iOS without websockets/SSL support, to match functionality available on Windows/Android.

Currently:

- The `SSL` target does not produce output, but runs the `openssl.bash` script to produce a `libssl.a` and `libcrypto.a`.
- The `libHttpClient_[iOS|macOS]` static-lib targets link against the `libssl.a` and `libcrypto.a` produced in by the `SSL` target.
- The `libHttpClientFramework_[iOS|macOS]` framework targets link against the `libHttpClient_[iOS|macOS]` targets.

This PR introduces the following changes:

- Splits the `SSL` target into `SSL_iOS` and `SSL_macOS`, each of which links together the `libssl.a` and `libcrypto.a` produced by `openssl.bash` into a `libSSL.a` output (using one dummy source file).
- Makes the `libHttpClient_[iOS|macOS]` targets link against the appropriate `libSSL.a` outputs, rather than directly against the `libssl.a` and `libcrypto.a` from `openssl.bash`.
- Adds two new static-lib targets, `libHttpClient_NOWEBSOCKETS_iOS` and `libHttpClient_NOWEBSOCKETS_macOS`, which do *not* link against `libSSL.a` and both have `-DHC_NOWEBSOCKETS` as a C++ flag.
- Adds two new framework targets, `libHttpClientFramework_NOWEBSOCKETS_iOS` and `libHttpClientFramework_NOWEBSOCKETS_macOS`, which link against the new static-lib targets above.
- Adds an `exports_NOWEBSOCKETS.exp` exported-symbols file for the new framework targets, which does not include the websockets-related symbols.

This PR also explicitly skips `arm64` architecture for Mac, which is causing a build failure in OpenSSL. This problem likely exists on `master`, but showed up when I updated the CI pipeline to build Mac during this PR process.


MSFT-Heba (approved):

sashaweiss:
Relatedly, @MSFT-Heba @jasonsandlin I noticed that the `libHttpClient_iOS_Full_Build` pipeline only builds the `libHttpClient_iOS` static-lib target. Thoughts on making it build `_macOS`, as well as the frameworks, and the `_NOWEBSOCKETS` versions of those too?

jasonsandlin (approved):

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
```diff
@@ -106,6 +180,27 @@
 			remoteGlobalIDString = 67EE817D2362599C00A5F3DE;
 			remoteInfo = SSL;
```
matt-vanderkolk, line 181: >SSL [](start = 16, length = 3)

Looks like you didn't rename the original SSL to SSL_iOS. Intentional?

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
```diff
@@ -127,8 +231,8 @@
...
 		2C872C7E221C93800054F791 /* libc++.tbd */ = {isa = PBXFileReference; lastKnownFileType = "sourcecode.text-based-dylib-definition"; name = "libc++.tbd"; path = "usr/lib/libc++.tbd"; sourceTree = SDKROOT; };
 		2CE97718225EA543001F5E22 /* openssl.bash */ = {isa = PBXFileReference; lastKnownFileType = text.script.sh; path = openssl.bash; sourceTree = "<group>"; };
-		2CE97719225EB368001F5E22 /* libcrypto.a */ = {isa = PBXFileReference; lastKnownFileType = archive.ar; name = libcrypto.a; path = ../../Binaries/mac/lib/libcrypto.a; sourceTree = "<group>"; };
-		2CE9771A225EB368001F5E22 /* libssl.a */ = {isa = PBXFileReference; lastKnownFileType = archive.ar; name = libssl.a; path = ../../Binaries/mac/lib/libssl.a; sourceTree = "<group>"; };
+		2CE97719225EB368001F5E22 /* libcrypto.a */ = {isa = PBXFileReference; lastKnownFileType = archive.ar; name = libcrypto.a; path = openssl/lib/libcrypto.a; sourceTree = BUILT_PRODUCTS_DIR; };
+		2CE9771A225EB368001F5E22 /* libssl.a */ = {isa = PBXFileReference; lastKnownFileType = archive.ar; name = libssl.a; path = openssl/lib/libssl.a; sourceTree = BUILT_PRODUCTS_DIR; };
```
matt-vanderkolk, line 235: Since these rely on a product path instead of a path that has its flavor (mac/ios) built into it, is the build smart enough for the following:
From a clean build, if you build iOS framework, then the Mac framework, will the Mac build job know it needs a different SSL product?

matt-vanderkolk:
> 				58A7E972209ADEB100CC6774 /* WebSocket */,

alphabetize

---
Refers to: Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj:491 in 641e4a5. [](commit_id = 641e4a58b0f0419637c3c9cde6402d7b1f7dc5d5, deletion_comment = False)

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
```diff
@@ -720,11 +1019,26 @@
...
+			files = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+		};
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
 		67EE8187236259D700A5F3DE /* Build openssl */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
```
matt-vanderkolk, line 1041: >alwaysOutOfDate = 1; [](start = 3, length = 20)

What's the effect of this? Does this always rebuild then?

Source/SSL/SSLDummy.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  SSLDummy.swift
+//  SSL
+//
+//  Created by Sasha Weiss on 1/4/21.
+//
```
matt-vanderkolk, line 6: Clean this up

matt-vanderkolk (changes requested):
:clock1:

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
sashaweiss: Yep! That's the big advantage of this. `BUILT_PRODUCTS_DIR` will be `DerivedData/.../Products/Debug` for Mac targets, `.../Products/Debug-iphoneos` for iPhone device targets, and `.../Products/Debug-iphonesimulator` for iOS simulators.

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
sashaweiss: It will always run the script, yeah. However, the script has logic in it to detect libs from a previous build and skip if they are present - so we want the script to run every time and bail if it detects it doesn't need to run.

Confusing, perhaps, but not sure I want to address this here beyond making the script run every time. Changed this because I had issues with the script not getting re-run for an incremental build when I needed it to be, plus the script is designed to handle this itself.

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
sashaweiss: That's so weird. It shows in the UI as `SSL_iOS`, so I'm pretty unclear what this refers to...

Source/SSL/SSLDummy.swift
sashaweiss: done

matt-vanderkolk (approved):
:shipit:

PR #558, opened 2021-01-15
Make iOS NOWEBSOCKETS targets have same name but be sub-directory-ed

sashaweiss:
iOS uses the name of an included framework as the prefix for a fully-qualified `#include` - for example, `#include <httpClient/async.h>` will resolve to `async.h` provided by a linked `HttpClient.framework`. Consequently, naming the products of the "no websockets" iOS targets to `HttpClient_NOWEBSOCKETS` means that any upstream includes would have to be `#include <httpClient_NOWEBSOCKETS/async.h>`, which breaks a lot of upstream code that expects to include the same way as it does now - plus it's ugly 😅 

This PR reverts the `PRODUCT_NAME` for the `libHttpClient_NOWEBSOCKETS_[iOS|macOS]` targets to `HttpClient`, which means those targets will produce an `HttpClient.framework` that works with existing upstream `#include`s. To prevent overlap/conflict between the "with- and without-websockets" frameworks (which now have the same name), this PR also adds the `WRAPPER_NAME` build setting (see Xcode docs [here](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html)) to make the `_NOWEBSOCKETS` targets place their resulting `HttpClient.framework` files in a subdirectory in the built products directory.

So when build side-by-side, the "with websockets" file will live at `DerivedData/.../Products/<config>/HttpClient.framework`, and the "no websockets" file will live at `DerivedData/.../Products/<config>/NOWEBSOCKETS/HttpClient.framework`. This means that upstream consumers of the `_NOWEBSOCKETS` targets can keep their `#include`s the same as long as they ensure the `_NOWEBSOCKETS` version of the framework is in their framework search path.

This PR also disables header maps for all the libHttpClient static lib targets, which was causing build issues in XAL upstream. This is safe, as header maps are simply a build optimization and the libHttpClient project already sets its header search paths correctly.

### DIFF
--- a/Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
+++ b/Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
@@ -1479,6 +1479,7 @@
 				OTHER_CPLUSPLUSFLAGS = "$(OTHER_CFLAGS)";
 				SDKROOT = iphoneos;
 				SKIP_INSTALL = YES;
+				USE_HEADERMAP = NO;
 			};
 			name = Debug;
 		};
@@ -1505,6 +1506,7 @@
 				OTHER_CPLUSPLUSFLAGS = "$(OTHER_CFLAGS)";
 				SDKROOT = iphoneos;
 				SKIP_INSTALL = YES;
+				USE_HEADERMAP = NO;
 			};
 			name = Release;
 		};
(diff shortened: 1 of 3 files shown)


lubeltra (approved):

MSFT-Heba (approved):

matt-vanderkolk (approved):
:shipit:

PR #559, opened 2021-01-19
Bump Android NDK version to r22

sashaweiss:
Per the GH issue here: https://github.com/actions/virtual-environments/issues/2420, the Mac images that ADO/GH-actions CI uses have removed Android NDK r21d (our currently pinned version) in favor of r22.

This PR bumps our version to r22.

### DIFF
--- a/Build/libHttpClient.Android.Workspace/common_android_config.gradle
+++ b/Build/libHttpClient.Android.Workspace/common_android_config.gradle
@@ -5,7 +5,7 @@ if (!pluginManager.hasPlugin("com.android.base")) {
 android {
     compileSdkVersion 27
     defaultConfig.targetSdkVersion 21
-    ndkVersion "21.3.6528147"
+    ndkVersion "22.0.7026061"
 
     flavorDimensions "default"
 


jasonsandlin (approved):

carstenhag:
Noticed that this was linked on the other issue. FYI, you should upgrade Android Gradle Plugin to 4.1.1, as per https://github.com/actions/virtual-environments/issues/2420#issuecomment-760988437 that includes a fix so the specified ndk version gets downloaded instead of throwing an error.

sashaweiss:
Thanks for the note @carstenhag! I'll throw that on my backlog. I was hesitant to let the plugin auto-resolve an NDK version since we've been burned in the past with some fairly obscure incompatibility issues between libraries compiled with different NDKs, but I'm also being burned by pinning so 🤷

carstenhag:
Ah I see. Yeah, we got burned by pinning the version too (but we also used AGP 4.0). We now use the latest AGP, and pinning should work fine now, even if the image doesnt have it

PR #567, opened 2021-03-02
Update path casing for Android Java files to align with package name

sashaweiss:
Java package names expect to follow the path name of files, including casing. Currently the Java code lives at `src/main/com/xbox/httpClient`, but the package name is `com.xbox.httpclient` (note the lowercase).

This PR changes the path of the Java files to be `src/main/com/xbox/httpclient`, following lowercasing. Prefer to change the path name than the package name, since it affects fewer things.


lubeltra:
I'm slightly worried about the change in case for the folders: on windows the new path and the old path are considered equal, so it will not change the folder names in the clone and then when more files are added they will pick up the old case for the folder and things will get messy

sashaweiss:
Per offline convo b/w me and @lubeltra, we were able to confirm that Windows will do the "good/correct" thing here and re-case the paths. See:

On `main` currently:
image

On `sasha/javaPathCasing`:
image

lubeltra (approved):

PR #570, opened 2021-03-03
Remove network check during Android HTTP request initialization

sashaweiss:
On Android, we currently query the OS while initializing an HTTP request to check if network "is available" before doing the initialization. However, the network-checking code uses deprecated APIs and given the logs from some recent bug reports in XAL we believe there is a subtle and rare bug in that code.

This PR removes the check during initialization and instead (mirroring iOS behavior) does a best-effort at detecting a "no-network" related failure when a request fails (by checking for a `java.net.UnknownHostException`), and passes that signal along in the error code on the call's handle.


jasonsandlin (approved):

matt-vanderkolk (approved):
:shipit:

PR #577, opened 2021-04-27
Read and write callbacks to support large file transfers

v-davwes:
This adds request body read and response body write callbacks that are invoked when the request body is being uploaded and the response body is downloaded, respectively. Adding this enables support for large file uploads and downloads that otherwise would cause out of memory errors, especially on more resource constrained devices.

There is no observable behavioral change when these new callbacks aren't used. So using `HCHttpCallRequest[Get|Set]RequestBodyBytes`, `HCHttpCallRequest[Get|Set]RequestBodyString`, `HCHttpCallResponse[Get|Set]ResponseBodyBytes`, and `HCHttpCallResponse[Get|Set]ResponseString` behave as they did before. Using the callbacks, however, will give the user control over how data is sent and received, and where it is stored.

These are the new callback prototypes:
```
typedef HRESULT
(CALLBACK* HCHttpCallRequestBodyReadFunction)(
    _In_ HCCallHandle call,
    _In_ size_t offset,
    _In_ size_t bytesAvailable,
    _Out_writes_bytes_to_(bytesAvailable, *bytesWritten) uint8_t* destination,
    _Out_ size_t* bytesWritten
    );
```
```
typedef HRESULT
(CALLBACK* HCHttpCallResponseBodyWriteFunction)(
    _In_ HCCallHandle call,
    _In_reads_bytes_(bytesAvailable) const uint8_t* source,
    _In_ size_t bytesAvailable,
    _Out_ size_t* bytesRead
    );
```

With setters and getters for both:
```
STDAPI HCHttpCallRequestSetRequestBodyReadFunction(
    _In_ HCCallHandle call,
    _In_ HCHttpCallRequestBodyReadFunction readFunction,
    _In_ uint32_t bodySize
    ) noexcept;

STDAPI HCHttpCallRequestGetRequestBodyReadFunction(
    _In_ HCCallHandle call,
    _Out_ HCHttpCallRequestBodyReadFunction* readFunction,
    _Out_ uint32_t* bodySize
    ) noexcept;
```
```
STDAPI HCHttpCallResponseSetResponseBodyWriteFunction(
    _In_ HCCallHandle call,
    _In_ HCHttpCallResponseBodyWriteFunction writeFunction
    ) noexcept;

STDAPI HCHttpCallResponseGetResponseBodyWriteFunction(
    _In_ HCCallHandle call,
    _Out_ HCHttpCallResponseBodyWriteFunction* writeFunction
    ) noexcept;
```


jasonsandlin (approved):
:shipit:

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpclient/HttpClientRequest.java
```diff
@@ -33,16 +35,17 @@ public void setHttpUrl(String url) {
...
-        if (body == null || body.length == 0) {
+    public void setHttpMethodAndBody(String method, long call, String contentType, long contentLength) {
+        RequestBody requestBody = null;
+        if (contentLength == 0) {
             if ("POST".equals(method) || "PUT".equals(method)) {
-                this.requestBuilder = this.requestBuilder.method(method, RequestBody.create(NO_BODY, null));
-            } else {
-                this.requestBuilder = this.requestBuilder.method(method, null);
+                MediaType mediaType = (contentType != null ? MediaType.parse(contentType) : null);
+                requestBody = RequestBody.create(mediaType, NO_BODY);
```
sashaweiss, line 43: Related to comment below, if you upgrade to OkHttp version 4.9.1 (I believe the one we upgraded to), the arguments to this method should be reversed as they've deprecated this one.

sashaweiss (changes requested):
Thanks for these changes, @v-davwes! Looking forward to them. Broadly looks good to me.

Left a number of style comments, mostly, with one or two questions and one bug I found when trying it out myself - but nothing complex or that would block the changes overall. I should have reviewed iOS before Android and then it would have been easier for me to parse 😛

Also worth noting that I only reviewed the iOS/Android/common code here, but I'm assuming @jasonsandlin has the Windows platforms covered.

Source/HTTP/httpcall_request.cpp
```diff
@@ -100,6 +130,29 @@ HCHttpCallRequestSetRequestBodyString(
     );
 }
 
+STDAPI
+HCHttpCallRequestSetRequestBodyReadFunction(
+    _In_ HCCallHandle call,
+    _In_ HCHttpCallRequestBodyReadFunction readFunction,
+    _In_ uint32_t bodySize
```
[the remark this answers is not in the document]
v-davwes: That's correct. I think this could be tweaked in the future to support chunked transfer encoding when the body size may not be known but shouldn't be the default because some endpoints don't seem to support or allow chunked requests. For example, while testing SISU was failing my requests with chunked transfer encoding but succeeded without it

Source/HTTP/httpcall_request.cpp
```diff
@@ -6,6 +6,35 @@
 
 using namespace xbox::httpclient;
 
+HRESULT CALLBACK DefaultRequestBodyReadFunction(
```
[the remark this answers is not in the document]
v-davwes: The request body read and response body write callbacks can be invoked repeatedly until all data has been sent or received. Each invocation reads/writes the next chunk to be processed and in the case of the read callback is tracked by the “offset” argument to the callback.

As for JNI round-trips, on average we shouldn’t have additional JNI calls since OkHttp sends data in 8KiB chunks and most requests are much smaller than that. But in the case of larger requests, we’ll incur the cost of additional JNI calls. However we reduce our memory usage by one half since we avoid duplicating the body’s data in the JVM and let OkHttp read our native buffers (or files) directly in 8KiB blocks

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpclient/HttpClientRequest.java
```diff
@@ -33,16 +35,17 @@ public void setHttpUrl(String url) {
...
             }
         } else {
-            this.requestBuilder = this.requestBuilder.method(method, RequestBody.create(body, MediaType.parse(contentType)));
+            requestBody = new HttpClientRequestBody(call, contentType, contentLength);
         }
+        this.requestBuilder.method(method, requestBody);
```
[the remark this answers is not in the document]
v-davwes: Thanks for digging into this! I'll spend some more time with it today

Include/httpClient/httpClient.h
```diff
@@ -496,6 +527,38 @@ STDAPI HCHttpCallRequestSetSSLValidation(
...
+/// <summary>
+/// The callback definition used by an HTTP call to write the response body.
+/// </summary>
+/// <param name="call">The handle of the HTTP call.</param>
+/// <param name="source">The source from which bytes may be read.</param>
+/// <param name="bytesAvailable">The maximum number of bytes that can be read from the source.</param>
+/// <param name="bytesRead>The number of bytes that were actually written to the source.</param>
+/// <returns>Result code for this callback. Possible values are S_OK, E_INVALIDARG, or E_FAIL.</returns>
+typedef HRESULT
+(CALLBACK* HCHttpCallResponseBodyWriteFunction)(
```
lubeltra, line 543: should probably be marked noexcept

Include/httpClient/httpClient.h
```diff
@@ -360,6 +360,37 @@ STDAPI HCHttpCallRequestSetRequestBodyString(
...
+/// The callback definition used by an HTTP call to read the request body.
+/// </summary>
+/// <param name="call">The handle of the HTTP call.</param>
+/// <param name="offset">The offset from the beginning of the request body.</param>
+/// <param name="bytesAvailable">The maximum number of bytes that can be written to the destination.</param>
+/// <param name="destination">The destination where data may be written to.</param>
+/// <param name="bytesWritten">The number of bytes that were actually written to destination.</param>
+/// <returns>Result code for this callback. Possible values are S_OK, E_INVALIDARG, or E_FAIL.</returns>
+typedef HRESULT
+(CALLBACK* HCHttpCallRequestBodyReadFunction)(
```
lubeltra, line 373: should probably be marked noexcept

Source/HTTP/httpcall.h
```diff
@@ -11,6 +11,21 @@ struct http_header_compare
 
 using http_header_map = http_internal_map<http_internal_string, http_internal_string, http_header_compare>;
 
+HRESULT CALLBACK DefaultRequestBodyReadFunction(
```
lubeltra, line 14: mark both default callbacks noexcept

Source/HTTP/XMLHttp/xmlhttp_http_task.cpp
```diff
@@ -131,12 +130,12 @@ void xmlhttp_http_task::perform_async(
             return;
         }
 
-        if (requestBodyBytes > 0 && requestBody != nullptr)
+        if (requestBodySize > 0 && requestBodyReadFunction != nullptr)
```
lubeltra, line 133: is it ever valid to have a body size > 0 but no read function?

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -339,19 +339,38 @@ void winhttp_http_task::_multiple_segment_write_data(_In_ winhttp_http_task* pRe
...
     uint64_t safeSize = std::min(pRequestContext->m_requestBodyRemainingToWrite, defaultChunkSize);
 
-    const BYTE* requestBody = nullptr;
-    uint32_t requestBodyBytes = 0;
-    if ((HCHttpCallRequestGetRequestBodyBytes(pRequestContext->m_call, &requestBody, &requestBodyBytes) != S_OK) || 
-        requestBody == nullptr)
+    HCHttpCallRequestBodyReadFunction readFunction = nullptr;
+    uint32_t bodySize = 0;
+
+    if ((HCHttpCallRequestGetRequestBodyReadFunction(pRequestContext->m_call, &readFunction, &bodySize) != S_OK) ||
```
lubeltra, line 345: please use the FAILED macro instead of comparing to S_OK

lubeltra, line 345: is it valid to have no readFunction if bodySize == 0?

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -339,19 +339,38 @@ void winhttp_http_task::_multiple_segment_write_data(_In_ winhttp_http_task* pRe
...
+        readFunction == nullptr)
+    {
+        pRequestContext->complete_task(E_FAIL, static_cast<uint32_t>(E_FAIL));
+        return;
+    }
+
+    pRequestContext->m_requestBuffer.resize(safeSize);
```
lubeltra, line 352: should this be in the try catch in case of out of memory?

lubeltra (changes requested):

Include/httpClient/httpClient.h
```diff
@@ -360,6 +360,37 @@ STDAPI HCHttpCallRequestSetRequestBodyString(
     _In_z_ const char* requestBodyString
     ) noexcept;
 
+/// <summary>
+/// The callback definition used by an HTTP call to read the request body.
```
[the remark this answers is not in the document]
v-davwes: Yeah, it'll be invoked on whatever thread http happens to be running on and different for each platform. I'll make that clearer in my comments

Include/httpClient/httpClient.h
```diff
@@ -360,6 +360,37 @@ STDAPI HCHttpCallRequestSetRequestBodyString(
...
+/// Sets a custom callback function that will be used to read the request body when the HTTP call is
+/// performed.
+/// </summary>
+/// <param name="call">The handle of the HTTP call.</param>
+/// <param name="readFunction">The request body read function this call should use.</param>
+/// <returns>Result code of this API operation. Possible values are S_OK or E_INVALIDARG.</returns>
+STDAPI HCHttpCallRequestSetRequestBodyReadFunction(
+    _In_ HCCallHandle call,
+    _In_ HCHttpCallRequestBodyReadFunction readFunction,
+    _In_ uint32_t bodySize
```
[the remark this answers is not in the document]
v-davwes: `uint32_t` seems to be the type used by libHttpClient to describe the body size. The callbacks are working with memory buffers that could be arbitrarily large (theoretically larger than UINT32_MAX) and conventionally use `size_t` to denote its size

Source/HTTP/XMLHttp/xmlhttp_http_task.cpp
v-davwes: No, we should always have a read and write function

Source/HTTP/WinHttp/winhttp_http_task.cpp
v-davwes: It won't be used in the case where bodySize == 0 but there should always be a read function

Include/httpClient/httpClient.h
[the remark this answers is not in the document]
lubeltra: I dislike the mix of the 2 types, mostly cause libHC should have used `size_t` for the body size all along.

Include/httpClient/httpClient.h
```diff
@@ -496,6 +527,38 @@ STDAPI HCHttpCallRequestSetSSLValidation(
...
+typedef HRESULT
+(CALLBACK* HCHttpCallResponseBodyWriteFunction)(
+    _In_ HCCallHandle call,
+    _In_reads_bytes_(bytesAvailable) const uint8_t* source,
+    _In_ size_t bytesAvailable,
+    _Out_ size_t* bytesRead
+    );
+
+/// <summary>
+/// Sets a custom callback function that will be used to write the response body when the HTTP call
```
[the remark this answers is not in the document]
v-davwes: Yes, this should be called before perform async. I need to add a check for it but it should probably return E_FAIL unless there is another precedent for this type of situation?

Include/httpClient/httpClient.h
v-davwes: I tried that initially but it turns out `noexcept` is not allowed in a typedef declaration. I think might be why the other callback types (HCMemAllocFunction, HCMemFreeFunction,  HCWebSocketMessageFunction, ...) don't have the `noexcept` qualifier either

Source/HTTP/httpcall_response.cpp
```diff
@@ -6,6 +6,63 @@
...
+{
+    if (call == nullptr || source == nullptr || bytesAvailable == 0 || bytesRead == nullptr)
+    {
+        return E_INVALIDARG;
+    }
+
+    HRESULT hr = HCHttpCallResponseAppendResponseBodyBytes(call, source, bytesAvailable);
+    if (SUCCEEDED(hr))
+    {
+        *bytesRead = bytesAvailable;
```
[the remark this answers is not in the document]
v-davwes: This basically says that the write function consumed all the bytes that were given to it, which is what we want

Include/httpClient/httpClient.h
```diff
@@ -496,6 +527,38 @@ STDAPI HCHttpCallRequestSetSSLValidation(
...
+/// <param name="source">The source from which bytes may be read.</param>
+/// <param name="bytesAvailable">The maximum number of bytes that can be read from the source.</param>
+/// <param name="bytesRead>The number of bytes that were actually written to the source.</param>
+/// <returns>Result code for this callback. Possible values are S_OK, E_INVALIDARG, or E_FAIL.</returns>
+typedef HRESULT
+(CALLBACK* HCHttpCallResponseBodyWriteFunction)(
+    _In_ HCCallHandle call,
+    _In_reads_bytes_(bytesAvailable) const uint8_t* source,
+    _In_ size_t bytesAvailable,
+    _Out_ size_t* bytesRead
```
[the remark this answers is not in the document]
v-davwes: That's the current behavior, I think it makes it more flexible but maybe more error prone. I could change it so the client needs to consume the whole chunk

Include/httpClient/httpClient.h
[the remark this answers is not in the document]
v-davwes: I see there's a `RETURN_IF_PERFORM_CALLED` macro, I'll use that

Source/HTTP/httpcall_request.cpp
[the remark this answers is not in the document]
sashaweiss: Sounds good. I think I had misunderstood something and was concerned that there would now be an additional JNI round-trip even if there was only one chunk, but I don't think that's necessarily the case. Regardless, if it's 8kb chunks, yeah not worried.

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpclient/HttpClientRequestBody.java
```diff
@@ -0,0 +1,92 @@
...
+        @Override
+        public long skip(long n) throws IOException {
+            this.offset += n;
+            return n;
+        }
+
+        private native int nativeRead(long callHandle, long srcOffset, byte[] dst, long dstOffset, long bytesAvailable) throws IOException;
+    }
+
+    long callHandle;
```
sashaweiss, line 69: Can we make these private as well, and potentially final if not modified?

Build/libHttpClient.141.Android.Java/app/src/main/java/com/xbox/httpclient/HttpClientRequestBody.java
```diff
@@ -0,0 +1,92 @@
...
+    MediaType contentType;
+    long contentLength;
+
+    public HttpClientRequestBody(long sourceCallHandle, String sourceContentType, long sourceContentLength) {
+        callHandle = sourceCallHandle;
```
sashaweiss, line 74: Prefer `this.` in the main class as well.

sashaweiss (approved):
Android and iOS look good to me. It's a big PR so possible I missed something 😛 but independently from @v-davwes I confirmed that the default read/write functions work in the XAL test app for iOS and Android.

Include/httpClient/httpClient.h
```diff
@@ -360,6 +360,37 @@ STDAPI HCHttpCallRequestSetRequestBodyString(
...
+(CALLBACK* HCHttpCallRequestBodyReadFunction)(
+    _In_ HCCallHandle call,
+    _In_ size_t offset,
+    _In_ size_t bytesAvailable,
+    _Out_writes_bytes_to_(bytesAvailable, *bytesWritten) uint8_t* destination,
+    _Out_ size_t* bytesWritten
+    );
+
+/// <summary>
+/// Sets a custom callback function that will be used to read the request body when the HTTP call is
```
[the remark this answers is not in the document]
lubeltra: still missing the note on set body bytes

does calling set body bytes after calling this return E_UNEXPECTED or reset the default callback?

lubeltra (changes requested):

Source/HTTP/httpcall_request.cpp
```diff
@@ -6,6 +6,35 @@
...
+{
+    if (call == nullptr || bytesAvailable == 0 || destination == nullptr || bytesWritten == nullptr)
+    {
+        return E_INVALIDARG;
+    }
+
+    uint8_t const* requestBody = nullptr;
+    uint32_t requestBodySize = 0;
+    HRESULT hr = HCHttpCallRequestGetRequestBodyBytes(call, &requestBody, &requestBodySize);
+    if (FAILED(hr) || (requestBody == nullptr && requestBodySize != 0))
```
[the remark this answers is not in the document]
lubeltra: offset == requestBodySize is invalid too

Include/httpClient/httpClient.h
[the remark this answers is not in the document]
v-davwes: I think it should reset the default callback, at least that's what I would expect as a user

Include/httpClient/httpClient.h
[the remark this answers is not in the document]
v-davwes: Sounds good. I'll remove the option to only read part of the data

Include/httpClient/httpClient.h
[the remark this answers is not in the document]
lubeltra: either way we should document it in that function

Source/HTTP/WinHttp/winhttp_http_task.cpp
v-davwes: I'm not a big fan of using assert, mainly because they disappear in release builds. I'll pass the failed result on to the caller

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -704,20 +709,22 @@ void winhttp_http_task::callback_status_data_available(
...
+        }
     }
 
     // Read new data into buffer
     if (newBytesAvailable > 0)
     {
-        size_t oldSize = pRequestContext->m_responseBuffer.size();
-        size_t newSize = oldSize + newBytesAvailable;
-        pRequestContext->m_responseBuffer.resize(newSize);
+        pRequestContext->m_responseBuffer.resize(newBytesAvailable);
```
lubeltra, line 722: you probably want a try/catch here for the allocation

Include/httpClient/httpClient.h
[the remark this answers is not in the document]
lubeltra: we should also make read body bytes fail if this has been called

Include/httpClient/httpClient.h
```diff
@@ -496,6 +535,38 @@ STDAPI HCHttpCallRequestSetSSLValidation(
...
+
+/// <summary>
+/// Sets a custom callback function that will be used to write the response body when the HTTP call
+/// is performed.
+/// </summary>
+/// <param name="call">The handle of the HTTP call.</param>
+/// <param name="writeFunction">The response body write function this call should use.</param>
+/// <returns>Result code of this API operation. Possible values are S_OK or E_INVALIDARG.</returns>
+/// <remarks>This must be called prior to calling HCHttpCallPerformAsync</remarks>
+STDAPI HCHttpCallResponseSetResponseBodyWriteFunction(
```
lubeltra, line 565: should take a context pointer

Include/httpClient/httpClient.h
```diff
@@ -360,6 +364,41 @@ STDAPI HCHttpCallRequestSetRequestBodyString(
...
+/// Sets a custom callback function that will be used to read the request body when the HTTP call is
+/// performed. Note that if a custom read callback is used, any request body data previously set by
+/// HCHttpCallRequestSetRequestBodyBytes or HCHttpCallRequestSetRequestBodyString is ignored making
+/// these API operations mutually exclusive.
+/// </summary>
+/// <param name="call">The handle of the HTTP call.</param>
+/// <param name="readFunction">The request body read function this call should use.</param>
+/// <returns>Result code of this API operation. Possible values are S_OK or E_INVALIDARG.</returns>
+/// <remarks>This must be called prior to calling HCHttpCallPerformAsync.</remarks>
+STDAPI HCHttpCallRequestSetRequestBodyReadFunction(
```
lubeltra, line 396: should take a context pointer

lubeltra (changes requested):

Source/HTTP/WinHttp/winhttp_http_task.cpp
v-davwes: Looks like these will be caught by the calling function: https://github.com/microsoft/libHttpClient/blob/main/Source/HTTP/WinHttp/winhttp_http_task.cpp#L891

lubeltra (approved):

PR #580, opened 2021-05-13
Avoid holding locks in winhttp_http_task completion callback handling when calling into websocket event functions

natiskan:
Avoid holding locks in winhttp_http_task when calling into externally provided event handler functions on completion callback.


Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -298,6 +298,14 @@ void winhttp_http_task::complete_task(_In_ HRESULT translatedHR)
 
 void winhttp_http_task::complete_task(_In_ HRESULT translatedHR, uint32_t platformSpecificError)
 {
+    win32_cs_autolock autoCriticalSection(&m_lock);
+	
+	// Exit early if error happened and it was removed from cache to avoid calling XAsyncComplete() multiple times
```
jasonsandlin, line 303: nit: tabs

jasonsandlin (approved):

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -495,6 +505,8 @@ void winhttp_http_task::callback_status_request_error(
         HRESULT hr = HCHttpCallRequestGetUrl(pRequestContext->m_call, &method, &url);
         if (SUCCEEDED(hr))
         {
+            win32_cs_autolock autoCriticalSection(&pRequestContext->m_lock);
```
jplafonta, line 508: Should this be release prior to calling complete_task? Does that synchronously call back into client's code?

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -701,6 +722,8 @@ void winhttp_http_task::callback_status_data_available(
     _In_ winhttp_http_task* pRequestContext,
     _In_ void* statusInfo)
 {
+    win32_cs_autolock autoCriticalSection(&pRequestContext->m_lock);
```
jplafonta, line 725: Same question as above

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -298,6 +298,14 @@ void winhttp_http_task::complete_task(_In_ HRESULT translatedHR)
 
 void winhttp_http_task::complete_task(_In_ HRESULT translatedHR, uint32_t platformSpecificError)
 {
+    win32_cs_autolock autoCriticalSection(&m_lock);
```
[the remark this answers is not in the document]
jasonsandlin: critical section on win32 is better/faster than mutex.  more detail here:

https://docs.microsoft.com/en-us/windows/win32/sync/critical-section-objects

Source/HTTP/WinHttp/winhttp_http_task.cpp
natiskan: It looks like it can, but XSAPI doesn't currently use this. I've updated the PR to not hold locks for complete_task.

Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -1719,22 +1765,26 @@ void winhttp_http_task::callback_websocket_status_headers_available(
     {
         DWORD dwError = GetLastError();
         HC_TRACE_ERROR(HTTPCLIENT, "HCHttpCallPerform [ID %llu] [TID %ul] WinHttpWebSocketCompleteUpgrade errorcode %d", TO_ULL(HCHttpCallGetId(pRequestContext->m_call)), GetCurrentThreadId(), dwError);
+        pRequestContext->m_lock.unlock();
         pRequestContext->complete_task(E_FAIL, HRESULT_FROM_WIN32(dwError));
+        return;
```
natiskan, line 1770: Can someone verify that this is the intended behavior?

jasonsandlin: if pRequestContext->m_hRequest == NULL then you can't do the code below if that's the question...

---
In reply to: [632151443](https://github.com/microsoft/libHttpClient/pull/580#discussion_r632151443) [](ancestors = 632151443)

jplafonta (approved):

PR #582, opened 2021-05-20
Change BasicAsciiLowercase implementation to avoid allocation

jplafonta:


Source/Common/utils.cpp
```diff
@@ -56,16 +56,14 @@ bool StringToUint4(char const* begin, char const* end, uint64_t& v, int32_t base
 
 void BasicAsciiLowercase(String& s)
 {
-    static std::locale const classicLocale = std::locale::classic();
-    std::transform(s.begin(), s.end(), s.begin(), [](char c)
+    std::transform(s.begin(), s.end(), s.begin(), [](char c) -> char
     {
-        if ((c & 0x7F) == c)
+        if (c <= 'Z' && c >= 'A')
```
lubeltra, line 61: I would leave the old check for c being in the basic ascii range and only if that pass do the check for A-Z

if we ever pass strings with UTF-8 in them we want to know cause this method is just not doing the right thing

jplafonta: added back the assert

lubeltra (approved):

PR #586, opened 2021-06-03
Release locks before calling into HC_WEBSOCKET client funcions

natiskan:
Additional places where locks are being held while calling into client functions, this time in HC_WEBSOCKET.


rakhad (approved):
:shipit:

PR #588, opened 2021-06-18
Task 33321557: Streamline XSAPI thunk dll generation and ship it in the GDK

tculotta:


rakhad (approved):

Include/httpClient/config.h
```diff
@@ -35,6 +35,15 @@
         #define HC_PLATFORM HC_PLATFORM_GDK
     #elif !defined(WINAPI_FAMILY) || (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP)
         #define HC_PLATFORM HC_PLATFORM_WIN32
+        #include <windef.h>
```
jasonsandlin, line 38: isn't this the XAL win32 section?  seems odd to need to put it here

tculotta:
Also changed to check against SDK version (this code was introduced in vibranium), so the #include is no longer needed which was problematic anyway.

Include/httpClient/config.h
tculotta: moved to correct block

jasonsandlin (approved):
:shipit:

PR #598, opened 2021-07-30
Move the Android build to a Windows agent

sashaweiss-msft:
Moves the Android build from a MacOS to Windows agent (which has increased capacity in Azure, so should reduce agent wait time). Also adds a debug/release split for the Android build.


jasonsandlin (approved):
:shipit:

jasonsandlin (approved):

PR #601, opened 2021-08-10
Short-circuit request body read if no more data

sashaweiss-msft:
On Win32, during request body reading we tell the OS how large the request body will be. The OS then does not ask for more bytes when the `offset` has reached the `requestBodySize`.

On iOS and Android, however, we aren't able to tell the OS the request body size ahead of time, and so it will call us even when the `offset` is equal to the `requestBodySize`.

This PR adds a short-circuit check for iOS and Android to not call the "request body read" callback if we know that we've already provided the entire request body, and to instead signal the OS that we are done.

### DIFF
--- a/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequestBody.java
+++ b/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequestBody.java
@@ -49,6 +49,7 @@ public final class HttpClientRequestBody extends RequestBody
 
             int bytesRead = nativeRead(this.callHandle, this.offset, destination, destinationOffset, length);
             if (bytesRead == -1) {
+                // Tell the OS that we are done reading
                 return -1;
             }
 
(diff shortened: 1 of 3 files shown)


lubeltra (approved):

v-davwes (approved):

jasonsandlin (approved):

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequestBody.java
```diff
@@ -48,7 +48,8 @@ public int read(byte[] destination, int destinationOffset, int length) throws IO
             }
 
             int bytesRead = nativeRead(this.callHandle, this.offset, destination, destinationOffset, length);
-            if (bytesRead == 0) {
```
jasonsandlin, line 51: should it be 0 or -1 ?

jasonsandlin (approved):

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequestBody.java
sashaweiss-msft: if we return 0 bytes read that probably represents a bug, and i lean towards making this listen explicitly for -1 so that if there is that bug we get into trouble around it rather than a potentially more opaque kind of trouble by telling the OS prematurely that we're done reading. Open to changing it, though.

jasonsandlin: alright, that's fine

PR #605, opened 2021-08-19
Swapping HTTP provider for GDK from WinHttp to XCurl

jplafonta:


Source/HTTP/XCurl/XCurlEasyRequest.cpp
```diff
@@ -0,0 +1,343 @@
...
+    UNREFERENCED_PARAMETER(hr);
+
+    // Always complete with XAsync success; http/network errors returned via HCCallHandle
+    XAsyncComplete(m_asyncBlock, S_OK, 0);
+}
+
+void XCurlEasyRequest::Fail(HRESULT hr)
+{
+    HC_TRACE_ERROR_HR(HTTPCLIENT, hr, "XCurlEasyRequest::Fail");
+    XAsyncComplete(m_asyncBlock, S_OK, 0);
```
jplafonta, line 135: >S_OK [](http://example.com/codeflow?start=33&length=4)

hr

Source/HTTP/XCurl/XCurlMulti.cpp
```diff
@@ -0,0 +1,153 @@
...
+            multi->FailAllRequests(hr);
+        }
+    }
+}
+
+HRESULT XCurlMulti::Perform() noexcept
+{
+    //int numFds{ 0 };
+    //CURLMcode res = curl_multi_poll(m_curlMultiHandle, nullptr, 0, 1000, &numFds);
+    //RETURN_IF_FAILED(HrFromCurlm(res));
```
jplafonta, line 83: remove

Source/HTTP/XCurl/XCurlMulti.h
```diff
@@ -0,0 +1,38 @@
...
+    XCurlMulti() = default;
+
+    static void CALLBACK TaskQueueCallback(_In_opt_ void* context, _In_ bool canceled) noexcept;
+    HRESULT Perform() noexcept;
+
+    // Fail all active requests due to unexpected CURLM or platform error
+    void FailAllRequests(HRESULT hr) noexcept;
+
+    CURLM* m_curlMultiHandle{ nullptr };
+    XTaskQueueHandle m_queue;
```
jplafonta, line 32: initialize

Source/HTTP/XCurl/XCurlEasyRequest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    if (!curlEasyHandle)
+    {
+        HC_TRACE_ERROR(HTTPCLIENT, "XCurlEasyRequest::Initialize:: curl_easy_init failed");
+        return E_FAIL;
+    }
+
+    http_stl_allocator<XCurlEasyRequest> a{};
+    HC_UNIQUE_PTR<XCurlEasyRequest> easyRequest{ new (a.allocate(1)) XCurlEasyRequest{ curlEasyHandle, hcCall, async } };
+
+    // body (first so we can override things curl "helpfully" sets for us)
```
lubeltra, line 35: the snark is familiar :P

Source/HTTP/XCurl/XCurlEasyRequest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    // url & method
+    char const* url = nullptr;
+    char const* method = nullptr;
+    RETURN_IF_FAILED(HCHttpCallRequestGetUrl(hcCall, &method, &url));
+    RETURN_IF_FAILED(easyRequest->SetOpt(CURLOPT_URL, url));
+
+    CURLoption opt = CURLOPT_HTTPGET;
+    RETURN_IF_FAILED(MethodStringToOpt(method, opt));
```
lubeltra, line 59: are you sure you don't need to support `DELETE`?

Source/HTTP/XCurl/XCurlEasyRequest.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+HRESULT XCurlEasyRequest::AddHeader(char const* name, char const* value) noexcept
+{
+    int required = std::snprintf(nullptr, 0, "%s: %s", name, value);
+    assert(required > 0);
+
+    m_headersBuffer.emplace_back();
+    auto& header = m_headersBuffer.back();
+
+    header.resize(static_cast<size_t>(required) + 1, '\0');
```
lubeltra, line 147: the +1/-1 dance is not necessary: std::string guarantees that there is space for the null terminator and I've been taking advantage of that in newer code

Source/HTTP/XCurl/XCurlEasyRequest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    int written = std::snprintf(&header[0], header.size(), "%s: %s", name, value);
+    assert(written == required);
+    (void)written;
+
+    header.resize(header.size() - 1); // drop null terminator
+
+    curl_slist*& list = m_headers;
+    list = curl_slist_append(list, header.c_str());
```
lubeltra, line 155: you could probably just use `m_headers` directly, I think the switch code is detritus from earlier iterations

Source/HTTP/XCurl/XCurlMulti.h
```diff
@@ -0,0 +1,38 @@
...
+namespace xbox
+{
+namespace http_client
+{
+
+class XCurlMulti
+{
+public:
+    static Result<HC_UNIQUE_PTR<XCurlMulti>> Initialize(XTaskQueuePortHandle workPort);
+    XCurlMulti(const XCurlMulti&) = delete;
```
lubeltra, line 16: same here

Source/HTTP/XCurl/XCurlMulti.cpp
```diff
@@ -0,0 +1,154 @@
...
+            }
+            break;
+            }
+        }
+    }
+
+    if (runningRequests)
+    {
+        // Reschedule Perform if there are still running requests
+        RETURN_IF_FAILED(XTaskQueueSubmitCallback(m_queue, XTaskQueuePort::Work, this, XCurlMulti::TaskQueueCallback));
```
lubeltra, line 131: is this going to be a "busy loop" as long as a request is outstanding?

what happens if a game tries to empty the work queue while we're making an http request that takes 2s to complete?

Source/HTTP/XCurl/XCurlProvider.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    performEnv = initResult.ExtractPayload();
+
+    return S_OK;
+}
+
+void Internal_CleanupHttpPlatform(HC_PERFORM_ENV* performEnv) noexcept
+{
+    // HC_PERFORM_ENV created with custom deleter - ~HC_PERFORM_ENV needs to be explicitly invoked.
+    performEnv->~HC_PERFORM_ENV();
```
lubeltra, line 23: do we need to actually delete the memory?

lubeltra (changes requested):

Source/HTTP/XCurl/XCurlProvider.cpp
jplafonta: good question.  Will look at other implementations and test

Source/HTTP/XCurl/XCurlMulti.cpp
jplafonta: It will get continually scheduled, yes. A couple thoughts/questsions:

1. can a client empty/terminate a TaskQueue port?  XCurlMulti is using a composite TaskQueue that it creates.  My understanding is clients can terminate their queues without this being affected.

2. Should we be rescheduling the next Perform with a delay?  I originally intended to do that but XCurl doesn't support curl_multi_timeout ([https://curl.se/libcurl/c/curl_multi_timeout.html](https://curl.se/libcurl/c/curl_multi_timeout.html)).  We could just use a small fixed, delay though

lubeltra: so consider a game that does this:
```
// inside some game loop (main or worker thread)
while (XAsyncDoWork(myQueue, false)) {}
do_other_important_per_frame_work();
```

Your approach will stall this thread until the http request is completed

Ideally we would have a way to only schedule work if there's something to be done (a callback version of `curl_multi_timeout`) since we don't have that, I think scheduling with a delay is the right way to go

jplafonta: Ya, from what I can tell, there isn't a way to get a callback from curl when there is perform work to be done. I will add a small delay between each Perform

Source/HTTP/XCurl/XCurlEasyRequest.cpp
jplafonta: we should support it.  Will add.

Source/HTTP/XCurl/XCurlEasyRequest.h
```diff
@@ -0,0 +1,67 @@
...
+    size_t m_bodyCopied{ 0 };
+    char m_errorBuffer[CURL_ERROR_SIZE]{ 0 };
+};
+
+//------------------------------------------------------------------------------
+// Template implementations
+//------------------------------------------------------------------------------
+
+template<typename T>
+HRESULT XCurlEasyRequest::SetOpt(CURLoption option, T v) noexcept
```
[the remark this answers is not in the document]
jplafonta: Changed this to require explicit template instantiation.

Source/HTTP/XCurl/XCurlProvider.cpp
jplafonta: added explicit deallocation

Utilities/CMake/CMakeLists.txt
```diff
@@ -130,6 +130,15 @@ set(XMLHttp_HTTP_Source_Files
     ../../../Source/HTTP/XMLHttp/xmlhttp_http_task.h
     )
 
+set(XCurl_Source_Files
```
jplafonta, line 133: >X [](http://example.com/codeflow?start=4&length=1)

Per @jasonsandlin recommendation, dropping 'X' from paths, filenames, and types as this provider could be used with other Curl implementations as well

Source/HTTP/Curl/CurlEasyRequest.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
+
+#if HC_PLATFORM == HC_PLATFORM_GDK
+#include <XCurl.h>
+#else
+// This path is untested, but this http providered should work with other Curl implementations as well
```
jasonsandlin, line 6: typo

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -0,0 +1,345 @@
...
+    if (getInfoRes != CURLE_OK)
+    {
+        result = getInfoRes;
+    }
+
+    if (result != CURLE_OK)
+    {
+        HC_TRACE_VERBOSE(HTTPCLIENT, "CurlEasyRequest::m_errorBuffer='%s'", m_errorBuffer);
+
+        HRESULT hr = HCHttpCallResponseSetNetworkErrorCode(m_hcCallHandle, E_FAIL, result);
```
IUseGoto2, line 125: You can get the underlying WinHTTP HRESULT error with CURLINFO_OS_ERRNO.  It's super useful for debugging issues at least from our end.

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -0,0 +1,345 @@
...
+    case CURLINFO_SSL_DATA_IN: event = "SSL IN"; break;
+    case CURLINFO_END: event = "END"; break;
+    }
+
+    if (type == CURLINFO_TEXT && data[size - 1] == '\n')
+    {
+        size -= 1;
+    }
+
+    HC_TRACE_IMPORTANT(HTTPCLIENT, "CURL %10s - %.*s", event, size, data);
```
IUseGoto2, line 317: This is very spammy and basically includes the entire request + response data.  "Important" might be the wrong classification?

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -0,0 +1,345 @@
...
+        assert(false);
+    }
+
+    HRESULT hr = HCHttpCallResponseSetHeaderWithLength(request->m_hcCallHandle, name, nameSize, value, valueSize);
+    assert(SUCCEEDED(hr));
+
+    return bufferSize;
+}
+
+size_t CurlEasyRequest::WriteDataCallback(char* buffer, size_t /*size*/, size_t nmemb, void* context) noexcept
```
IUseGoto2, line 283: This should also be size * count as in the other headers.  We only ever return "1" right now as documented by libcurl, but it's still best practice.

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -0,0 +1,157 @@
+#include "pch.h"
+#include "CurlMulti.h"
+#include "CurlProvider.h"
+
+namespace xbox
+{
+namespace http_client
+{
+
+// XCurl doesn't support curl_multi_timeout, so use a small, fixed delay between calls to curl_multi_perform
```
IUseGoto2, line 10: should we be supporting this?

Source/HTTP/Curl/CurlEasyRequest.cpp
IUseGoto2: Also, the one "special suspend" thing with XCurl is that there is a special error for "failed to due to suspend" that should be documented out of libhttpclient/xsapi/etc.

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+        curl_multi_cleanup(m_curlMultiHandle);
+    }
+}
+
+HRESULT CurlMulti::AddRequest(HC_UNIQUE_PTR<CurlEasyRequest>&& easyRequest)
+{
+    std::unique_lock<std::mutex> lock{ m_mutex };
+
+    auto result = curl_multi_add_handle(m_curlMultiHandle, easyRequest->Handle());
```
IUseGoto2, line 56: the lock probably solves any problem here, but would recommend doing this last after adding to your lists and such since you can start getting callbacks or in the worst case completion messages once this call is made when perform is called.

Source/HTTP/Curl/CurlProvider.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+HRESULT HrFromCurle(CURLcode c) noexcept
+{
+    if (c == CURLE_OK)
+    {
+        return S_OK;
+    }
+    else
+    {
+        return E_FAIL;
```
IUseGoto2, line 69: this is unfortunate; is there a way to expose the HRESULT from the OS_ERRNO when possible?

Source/HTTP/Curl/CurlProvider.cpp
```diff
@@ -19,8 +19,10 @@ HRESULT Internal_InitializeHttpPlatform(HCInitArgs* args, PerformEnv& performEnv
 
 void Internal_CleanupHttpPlatform(HC_PERFORM_ENV* performEnv) noexcept
 {
-    // HC_PERFORM_ENV created with custom deleter - ~HC_PERFORM_ENV needs to be explicitly invoked.
-    performEnv->~HC_PERFORM_ENV();
+    // HC_PERFORM_ENV created with custom deleter - HC_PERFORM_ENV needs to be destroyed and cleaned up explicitly.
+    http_stl_allocator<HC_PERFORM_ENV> a{};
+    std::allocator_traits<http_stl_allocator<HC_PERFORM_ENV>>::destroy(a, std::addressof(*performEnv));
```
lubeltra, line 24: why the deref / std::addressof dance?

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -0,0 +1,345 @@
...
+    if (bodySize > 0)
+    {
+        // we set both POSTFIELDSIZE and INFILESIZE because curl uses one or the
+        // other depending on method
+        RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_POSTFIELDSIZE, bodySize));
+        RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_INFILESIZE, bodySize));
+
+        // read callback
+        RETURN_IF_FAILED(easyRequest->SetOpt<curl_read_callback>(CURLOPT_READFUNCTION, &ReadCallback));
+        RETURN_IF_FAILED(easyRequest->SetOpt<void*>(CURLOPT_READDATA, easyRequest.get()));
```
IUseGoto2, line 49: Some options that maybe you want to set that I thought LibHttpClient supported with the direct WinHTTP implementation?
CURLOPT_USERAGENT
CURLOPT_ACCEPT_ENCODING

lubeltra (approved):
One minor question left

Source/HTTP/Curl/CurlEasyRequest.cpp
lubeltra: You are right this feels more like info

Source/HTTP/Curl/CurlProvider.cpp
jplafonta: looks unnecessary (pulled this from winHttp implementation of Cleanup)

Source/HTTP/Curl/CurlEasyRequest.cpp
lubeltra: from https://curl.se/libcurl/c/CURLOPT_WRITEFUNCTION.html
>  ptr points to the delivered data, and the size of that data is nmemb; size is always 1.
To me that says that we don't know what to do with `size` if it is not 1. Maybe we multiply `nmemb` or maybe `nmemb` is premultiplied, who knows? The spec doesn't say and anyone setting `size` to not 1 is breaking the spec, trying to guess how they are breaking it could result in just as many bugs as assuming they aren't breaking it.

Source/HTTP/Curl/CurlMulti.cpp
lubeltra: Yes and no.
It would be great to have an equivalent but `curl_multi_timeout` is very much designed to work with Unix style `select` calls which have the completely unacceptable (in my opinion) downside of *blocking* the thread until an event occurs.

I will oppose any HTTP implementation that causes a thread calling XTaskQueueDispatch to block until the HTTP request is completed

Source/HTTP/Curl/CurlEasyRequest.cpp
jplafonta: I do see that winHttp provider sets libHttpClient as the User-Agent if one wasn't provided by the caller.  If clients manually set User-Agent header and we set CURLOP_USERAGENT, how is that handled?

I don't see where accept encoding is set in the WinHttp implementation, but maybe I'm just missing it.

Source/HTTP/Curl/CurlMulti.cpp
jplafonta: with the lock, perform can't be called until the easy request is added to m_easyRequests and to the curl multi handle.  This order makes sense to me so that we don't have to worry about removing the request from m_easyRequests if curl_multi_add_handle_fails

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -0,0 +1,157 @@
...
+    if (result != CURLM_OK)
+    {
+        HC_TRACE_ERROR(HTTPCLIENT, "XCurlMulti::AddRequest: curl_multi_add_handle failed with CURLCode=%u", result);
+        return HrFromCurlm(result);
+    }
+
+    m_easyRequests.emplace(easyRequest->Handle(), std::move(easyRequest));
+
+    RETURN_IF_FAILED(XTaskQueueSubmitCallback(m_queue, XTaskQueuePort::Work, this, CurlMulti::TaskQueueCallback));
```
jplafonta, line 65: should release m_mutex before submitting callback in case m_queue is immediate dispatch

Source/HTTP/Curl/CurlEasyRequest.cpp
jplafonta: changed to HC_TRACE_INFORMATION

Source/HTTP/Curl/CurlMulti.cpp
jplafonta: It would be nice to have some mechanism to know when there is work to do in curl_multi_perform.  Unless I'm misunderstanding something, the only way the only way to implement parallel http requests in a non-blocking fashion is to add an arbitrary sleep.

Source/HTTP/Curl/CurlProvider.cpp
jplafonta: There probably is.  Will think about that and update accordingly

Source/HTTP/Curl/CurlEasyRequest.cpp
jplafonta: In theory we want to work with other curl implementations as well.  If the curl spec dictates its always 1 I think we should assume that is true.

Source/HTTP/Curl/CurlMulti.cpp
IUseGoto2: libcurl unfortunately doesn't provide any way to notify you.  You can poll with curl_multi_poll or curl_multi_wait with a zeroed-out or small timeout, but libcurl is designed with select (or those two functions on windows where select doesn't work on arbitrary file handles) in mind.  That's better than an arbitrary sleep, but it still burns a thread.  You don't realize the joys of overlapped IO until its missing :/

Source/HTTP/Curl/CurlEasyRequest.cpp
IUseGoto2: It's also true in all implementations for all the other callbacks, too.  I think it's not gauranteed for non-HTTP requests which don't matter here, but it's weird to single this one out.

Source/HTTP/Curl/CurlEasyRequest.cpp
IUseGoto2: I was pretty sure that WinHTTP will override any specified user agent header regardless of whether you're using xcurl or not.  If it doesn't, then using the existing header mechanism is fine.  I thought that the encoding was set to JSON somewhere for winhttp?

Source/HTTP/Curl/CurlMulti.cpp
lubeltra: Indeed, the solution here at least doesn't burn the thread since we're scheduling work in the future.

I would still love for XCurl to provide a GDK specific function that integrates curl polling into the XAsync framework nicely

jplafonta: Going to leave the sleep for now - really a priority to avoid burning the thread.

lubeltra: Just to be clear, there is no sleep in this code. `XTaskQueueSubmitDelayedCallback` submits the task to be executed in the future but does not send the thread to sleep.

Source/HTTP/Curl/CurlEasyRequest.cpp
jplafonta: I think XSAPI sets the accept encoding header, but its not done anywhere in libHttpClient.  I will add a libHttpClient User-Agent if one was not manually specified (we want to allow XSAPI to set its own User-Agent for example)

Source/HTTP/Curl/CurlMulti.cpp
IUseGoto2: immediate dispatch won't work at all here and will fail when you try to add a delayed callback?  Should that be checked up front?

Source/HTTP/Curl/CurlProvider.cpp
jplafonta: This is only getting used in cases where some curl API fails unexpectedly (ex curl_easy_setopt, curl_initialize)

Source/HTTP/Curl/CurlEasyRequest.cpp
lubeltra: How would libhc even know what encodings the client accepts?

Source/HTTP/Curl/CurlMulti.cpp
lubeltra: why did we ever make immediate mode, it adds so many headaches
we should probably just reject immediate queues when the request is being passed in

jplafonta: Ya fair point.  Should probably fail for all providers but will add a check here for now

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -9,6 +9,7 @@ namespace http_client
 
 // XCurl doesn't support curl_multi_timeout, so use a small, fixed delay between calls to curl_multi_perform
 #define PERFORM_DELAY_MS 50
+#define POLL_TIMEOUT_MS 10
```
IUseGoto2, line 12: I think with the new model 0 is a reasonable choice.  I don't mind this, but it's now a 10ms sleep every time the work is queued.

Source/HTTP/Curl/CurlMulti.cpp
jplafonta: Actually seems like we can't even find the dispatch mode of a queue after creation.  That HR should get bubbled up correctly so will just leave it.

Source/HTTP/Curl/CurlMulti.cpp
jplafonta: changed to 0

Source/HTTP/Curl/CurlEasyRequest.cpp
lubeltra: looking at the docs, the header callback is also documented to always have 1 for size but not read

https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html
https://curl.se/libcurl/c/CURLOPT_READFUNCTION.html

PR #609, opened 2021-09-10
Add missing header

jplafonta:


Source/Common/pch_common.h
```diff
@@ -52,6 +52,7 @@
 #include <codecvt>
 #include <iomanip>
 #include <functional>
+#include <list>
```
jgalvan, line 55: nit: should this be alphabetical?

jgalvan (approved):

Source/Common/pch_common.h
jplafonta: sure, looks like it mostly is.  Will fix the few that aren't

jgalvan (approved):

jasonsandlin (approved):

PR #614, opened 2021-09-20
Support the streaming callbacks in the mocks

sashaweiss-msft:
Currently, the mocks do not use the `readBody` and `writeBody` callbacks added in #577. This PR adds support for reading from/writing to `HC_CALL`s that have a non-default read/write callback configured. Mock calls will still only support reading/writing the entire response body at once.

### DIFF
--- a/Source/Mock/lhc_mock.cpp
+++ b/Source/Mock/lhc_mock.cpp
@@ -88,11 +88,20 @@ bool Mock_Internal_HCHttpCallPerformAsync(
         );
     }
 
+    // Read response from mock
     size_t byteBuf;
     HCHttpCallResponseGetResponseBodyBytesSize(mock, &byteBuf);
-    http_memory_buffer buffer(byteBuf);
-    HCHttpCallResponseGetResponseBodyBytes(mock, byteBuf, static_cast<uint8_t*>(buffer.get()), nullptr);
-    HCHttpCallResponseSetResponseBodyBytes(originalCall, static_cast<uint8_t*>(buffer.get()), byteBuf);
+    if (byteBuf > 0)
+    {
+        http_memory_buffer buffer(byteBuf);
+        HCHttpCallResponseGetResponseBodyBytes(mock, byteBuf, static_cast<uint8_t*>(buffer.get()), nullptr);
+
+        // Write response to original call
+        HCHttpCallResponseBodyWriteFunction writeFunction;
+        void* context;
+        HCHttpCallResponseGetResponseBodyWriteFunction(originalCall, &writeFunction, &context);
+        writeFunction(originalCall, static_cast<uint8_t*>(buffer.get()), byteBuf, context);
+    }
 
     uint32_t code;
     HCHttpCallResponseGetStatusCode(mock, &code);
--- a/Source/Mock/lhc_mock.cpp
+++ b/Source/Mock/lhc_mock.cpp
@@ -141,9 +141,14 @@ HRESULT Mock_Internal_ReadRequestBodyIntoMemory(
     while (offset < bodySize)
     {
         size_t bytesWritten = 0;
-        RETURN_IF_FAILED(
-            readFunction(originalCall, offset, bodySize - offset, context, tempBodyBytes.data() + offset, &bytesWritten)
-        );
+        RETURN_IF_FAILED(readFunction(
+            originalCall,
+            offset,
+            bodySize - offset,
+            context,
+            tempBodyBytes.data() + offset,
+            &bytesWritten
+        ));
 
         offset += bytesWritten;
     }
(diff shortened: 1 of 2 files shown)


Source/Mock/lhc_mock.cpp
```diff
@@ -108,3 +124,38 @@ bool Mock_Internal_HCHttpCallPerformAsync(
...
+
+HRESULT Mock_Internal_ReadRequestBodyIntoMemory(
+    _In_ HCCallHandle originalCall,
+    _Out_ http_internal_vector<uint8_t>* bodyBytes
+    )
+{
+    HCHttpCallRequestBodyReadFunction readFunction;
+    size_t bodySize;
+    void* context;
+    RETURN_IF_FAILED(
```
lubeltra, line 136: one line

Source/Mock/lhc_mock.cpp
```diff
@@ -108,3 +124,38 @@ bool Mock_Internal_HCHttpCallPerformAsync(
...
+    );
+
+    http_internal_vector<uint8_t> tempBodyBytes(bodySize);
+
+    size_t offset = 0;
+    size_t bytesWritten;
+    while (offset < bodySize)
+    {
+        RETURN_IF_FAILED(
+            readFunction(
```
lubeltra, line 147: don't break for the function name

Source/Mock/lhc_mock.cpp
```diff
@@ -108,3 +124,38 @@ bool Mock_Internal_HCHttpCallPerformAsync(
 
     return true;
 }
+
+HRESULT Mock_Internal_ReadRequestBodyIntoMemory(
+    _In_ HCCallHandle originalCall,
+    _Out_ http_internal_vector<uint8_t>* bodyBytes
+    )
+{
+    HCHttpCallRequestBodyReadFunction readFunction;
```
lubeltra, line 133: init all of these

Source/Mock/lhc_mock.cpp
```diff
@@ -108,3 +124,38 @@ bool Mock_Internal_HCHttpCallPerformAsync(
...
+    size_t bodySize;
+    void* context;
+    RETURN_IF_FAILED(
+        HCHttpCallRequestGetRequestBodyReadFunction(originalCall, &readFunction, &bodySize, &context)
+    );
+
+    http_internal_vector<uint8_t> tempBodyBytes(bodySize);
+
+    size_t offset = 0;
+    size_t bytesWritten;
```
lubeltra, line 143: init, also move inside the loop

Source/Mock/lhc_mock.cpp
```diff
@@ -108,3 +124,38 @@ bool Mock_Internal_HCHttpCallPerformAsync(
...
+    )
+{
+    HCHttpCallRequestBodyReadFunction readFunction;
+    size_t bodySize;
+    void* context;
+    RETURN_IF_FAILED(
+        HCHttpCallRequestGetRequestBodyReadFunction(originalCall, &readFunction, &bodySize, &context)
+    );
+
+    http_internal_vector<uint8_t> tempBodyBytes(bodySize);
```
lubeltra, line 140: is this code safe to throw exceptions in?

lubeltra (changes requested):

Source/Mock/lhc_mock.cpp
sashaweiss-msft: err, i'm not sure. are you asking because this might throw while allocating?

Source/Mock/lhc_mock.cpp
sashaweiss-msft: we return if they won't be set, is init important? i don't have a huge problem with it, but curious

Source/Mock/lhc_mock.cpp
sashaweiss-msft: the previous code used an `http_memory_buffer`, happy to change to doing that if it'll avoid this question lol

lubeltra: no that doesn't help, let's assume the old code was doing the right thing

Source/Mock/lhc_mock.cpp
```diff
@@ -130,29 +130,21 @@ HRESULT Mock_Internal_ReadRequestBodyIntoMemory(
...
         RETURN_IF_FAILED(
-            readFunction(
-                originalCall,
-                offset,
-                bodySize - offset,
-                context,
-                tempBodyBytes.data() + offset,
-                &bytesWritten
-            )
+            readFunction(originalCall, offset, bodySize - offset, context, tempBodyBytes.data() + offset, &bytesWritten)
```
lubeltra, line 145: that's the opposite of what I meant :P

lubeltra (changes requested):

Source/Mock/lhc_mock.cpp
sashaweiss-msft: lmfao so all one line?

lubeltra (approved):

PR #615, opened 2021-09-22
Update min iOS version to 11.0

sashaweiss-msft:
Main change:

* Updates the iOS deployment targets to 11.0 across all libHttpClient targets. This drops armv7 architectures, which are only supported below iOS 11.0.

Ancillary changes:

* Enables "base internationalization" to silence a warning. (No meaningful impact as libHttpClient includes no strings.)

### DIFF
--- a/Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
+++ b/Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
@@ -1004,6 +1004,7 @@
 			hasScannedForEncodings = 0;
 			knownRegions = (
 				en,
+				Base,
 			);
 			mainGroup = 58722D05209AD61900B071F7;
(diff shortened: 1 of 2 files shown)


SahilAshar (approved):

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
```diff
@@ -1004,6 +1004,7 @@
 			hasScannedForEncodings = 0;
 			knownRegions = (
 				en,
+				Base,
```
jgalvan, line 1007: what does Base stand for?

sashaweiss-msft: I'm not 100% sure, but there was a build warning for "enable base internationalization" that i added this to silence. It's ultimately irrelevant since libHttpClient (nor XAL) includes any strings, so all this does is silence a warning.

PR #616, opened 2021-09-22
Fix WinHttp WebSocket stack for GDK

jplafonta:
* HC_PERFORM_ENV is the context for both the HTTP and WebSocket providers.  On GDK, we want to user the Curl HTTP provider but the WinHttp WebSocket provider, so I've separated out WinHttpState from the WinHttp provider's HC_PERFORM_ENV so that it can be included in XCurl provider as well.
* Also added missing XCurl lib path to libHttpClient.props that was causing link issues
* Properly use client WriteResponseBody/ReadRequestBody callbacks in Curl HTTP provider


jasonsandlin (approved):
:shipit:

PR #619, opened 2021-09-30
Readding i386 for simulator support

MSFT-Heba:
Due to some unexpected behavior from XCode, i386 is still needed for building simulators on build machines

### DIFF
--- a/Utilities/XcodeBuildScripts/openssl.bash
+++ b/Utilities/XcodeBuildScripts/openssl.bash
@@ -92,7 +92,6 @@ for BUILD_ARCH in $BUILD_ARCHS; do
     #
     # - x86_64, i386: Mac
     # - arm64: Might be an M1 Mac or a physical iOS device
-    # - armv7: Very old phyiscal iOS device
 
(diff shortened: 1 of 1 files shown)


Utilities/XcodeBuildScripts/openssl.bash
```diff
@@ -95,6 +95,8 @@ for BUILD_ARCH in $BUILD_ARCHS; do
 
     if [ "$BUILD_ARCH" == "x86_64" ]; then
         ./Configure darwin64-x86_64-cc shared enable-ec_nistp_64_gcc_128 no-ssl2 no-ssl3 no-comp no-async --prefix="$OPENSSL_TMP/" --openssldir="$OPENSSL_TMP/"
+    elif [[ "$BUILDARCH" = *"i386"* ]]; then
```
sashaweiss-msft, line 98: Let's do `== "i386"` rather than `= *"i386"*`, since we know it should be that string exactly?

sashaweiss-msft (approved):

PR #620, opened 2021-10-05
Fix WinHttp SSL validation and other small fixes

jgalvan:
- When SSL validation was added, it was only enabled for specific platforms, however there was one piece of code that was left without the platform check. This would break any other platform that doesn't use it. Original PR for reference: #517
- Fixing C4701/C4703 for uninitialized variables


Source/HTTP/WinHttp/winhttp_http_task.cpp
```diff
@@ -1208,6 +1208,7 @@ HRESULT winhttp_http_task::send(
         return HRESULT_FROM_WIN32(dwError);
     }
 
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK
```
lubeltra, line 1211: As I mentioned offline, I think this code should be on for GRTS

lubeltra (changes requested):

Source/HTTP/WinHttp/winhttp_http_task.cpp
sashaweiss-msft: but there's no GRTS `HC_PLATFORM`, right?

sashaweiss-msft (approved):

lubeltra (approved):

jasonsandlin (approved):

boczar:
Hi folks,

Sorry for digging this up two months later, but how is this supposed to work?

#if HC_PLATFORM_IS_MICROSOFT && !HC_PLATFORM_UWP && !HC_PLAFTORM_XDK

HC_PLATFORM_UWP is defined as '2'. Always. So the above statement will always be false, regardless what platform you build on right? Or am I missing something?

Previous version was functionally correct:
Although it could be adapted to match the change like this:
#if HC_PLATFORM_IS_MICROSOFT && (HC_PLATFORM != HC_PLATFORM_UWP) && (HC_PLATFORM != HC_PLAFTORM_XDK)

jplafonta:
I agree, this looks incorrect.  I will PR a fix for this and do a quick scan for other spots where it might have been messed up

---
In reply to: [990140683](https://github.com/microsoft/libHttpClient/pull/620#issuecomment-990140683) [](http://example.com/codeflow?ancestors=990140683,990140683)

jplafonta:
@boczar Looks like this has since been fixed.

---
In reply to: [990227793](https://github.com/microsoft/libHttpClient/pull/620#issuecomment-990227793) [](http://example.com/codeflow?ancestors=990227793,990140683)

PR #621, opened 2021-10-13
Switch to GetByteArrayElements from GetPrimitiveArrayCritical to avoid deadlock issues in JNI

lubeltra:


jplafonta (approved):
:shipit:

Hamlin-Li:
I hope [Region Pinning in JDK](https://bugs.openjdk.java.net/browse/JDK-8276094) may help eliminate or alleviate this problem in the future.

PR #622, opened 2021-10-14
Set trace level to verbose/important based on configuration

sashaweiss-msft:
For debug builds set `HC_TRACE_BUILD_LEVEL` to `VERBOSE`, and for release builds set it to `IMPORTANT`.

We currently do this for Android, but not Xcode and Visual Studio builds.

### DIFF
--- a/Build/libHttpClient.internal.props
+++ b/Build/libHttpClient.internal.props
@@ -12,11 +12,14 @@
     <ClCompile>
       <PreprocessorDefinitions>
         $(libHttpClientDefine);
-        HC_TRACE_BUILD_LEVEL=HC_PRIVATE_TRACE_LEVEL_IMPORTANT;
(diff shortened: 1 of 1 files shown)


Build/libHttpClient.internal.props
```diff
@@ -12,6 +12,11 @@
     <ClCompile>
       <PreprocessorDefinitions>
         $(libHttpClientDefine);
+        HC_TRACE_BUILD_LEVEL=HC_PRIVATE_TRACE_LEVEL_IMPORTANT;
```
lubeltra, line 15: have `Condition="'$(Configuration)' != 'Debug'"` on the first branch
have `$(libHttpClientDefine);` included in the second branch

lubeltra (changes requested):

PR #625, opened 2021-10-15
Reworking WinHttp stack to support handling PLM events

jplafonta:


Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -0,0 +1,556 @@
...
+#include "Websocket/WinHTTP/winhttp_websocket.h"
+#endif
+
+NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
+
+Result<std::shared_ptr<WinHttpProvider>> WinHttpProvider::Initialize()
+{
+    http_stl_allocator<WinHttpProvider> a{};
+    auto provider = std::shared_ptr<WinHttpProvider>{ new (a.allocate(1)) WinHttpProvider };
+
```
jasonsandlin, line 17: isn't there some helper function like MakeShared that'd work here?

jasonsandlin (approved):
:shipit:

Source/HTTP/WinHttp/winhttp_provider.cpp
jplafonta: That doesn't work with a private constructor.  I've marked the constructor private so that its clear that this class should be created/initialized with this method so that any initialization errors get reported.

PR #626, opened 2021-10-16
Expose tracing init/cleanup separately from overall libHttpClient init

sashaweiss-msft:
Expose `HCTraceInit()` and `HCTraceCleanup()` as public functions separate from just an `HCInitialize()` call. This allows tracing to happen before/after libHttpClient itself has been initialized/cleaned up.

### DIFF
--- a/Include/httpClient/trace.h
+++ b/Include/httpClient/trace.h
@@ -145,6 +145,16 @@ enum class HCTraceLevel : uint32_t
 };
 
 /// <summary>
+/// Initialize tracing for the library.
+/// </summary>
+void HCTraceInit() noexcept;
+
+/// <summary>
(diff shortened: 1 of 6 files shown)


Include/httpClient/trace.h
```diff
@@ -144,6 +144,16 @@ enum class HCTraceLevel : uint32_t
     Verbose = HC_PRIVATE_TRACE_LEVEL_VERBOSE,
 };
 
+/// <summary>
+/// Initialize tracing for the library.
+/// </summary>
+void HCTraceInit() noexcept;
```
lubeltra, line 150: can you add a note that HCInit/Cleanup call these implicitly?

PR #632, opened 2021-11-09
Partial Revert of #608, Updating Min Target to Ensure 8.1 Backcompat

SahilAshar:
#608 Changed the `WindowsTargetPlatform` to hard target Windows 8.1, so that Psychonauts 2 would be able to run on Windows 7. However, this change breaks XAL build pipelines, as our Windows2019 images don't support this scenario. We can do a partial revert and add Windows 8.1 backwards compatibility by specifying a `MinTargetVersion`


SahilAshar:
Since I didn't create a new branch for this fix, it's re-adding the commits for updating the Android NDK. That was already merged in #631

PR #633, opened 2021-11-09
Partial Revert of #608, Updating Min Target to Ensure 8.1 Backcompat

SahilAshar:
Partial revert of #608. We can add Windows 8.1 backwards compatibility by specifying a MinTargetVersion


jasonsandlin (approved):

Utilities/CMake/template-libHttpClient.142.Win32.C.vcxproj
```diff
@@ -43,7 +43,8 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
     <PreferredToolArchitecture>x64</PreferredToolArchitecture>
-    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
sashaweiss-msft, line 46: which one was the janky win10 version? not this one I assume?

sashaweiss-msft (approved):

Utilities/CMake/template-libHttpClient.142.Win32.C.vcxproj
SahilAshar: it was this, but it's only janky with specific components on VS2017. This is what LibHC had previously, just reverting back to it.

jasonsandlin (approved):

PR #640, opened 2021-11-12
vs2022 projects should locate the v142 LHC projects

tculotta:


libHttpClient.props
```diff
@@ -36,6 +36,7 @@
     <HCLibToolset Condition="'$(PlatformToolset)'=='v140'">140</HCLibToolset>
     <HCLibToolset Condition="'$(PlatformToolset)'=='v141'">141</HCLibToolset>
     <HCLibToolset Condition="'$(PlatformToolset)'=='v142'">142</HCLibToolset>
+    <HCLibToolset Condition="'$(PlatformToolset)'=='v143'">142</HCLibToolset>
```
jplafonta, line 39: ><HCLibToolset Condition="'$(PlatformToolset)'=='v143'">142</HCLibToolset> [](http://example.com/codeflow?start=4&length=73)

Are we confident that clients using a different toolset won't have issues linking with 142 libs?  @lubeltra did you say we've seen issues with toolset mismatch in the past?

lubeltra: this seems like an extremely bad idea
why are we doin this?

lubeltra: given the way the libhc build system is set up if we want to support vs2022 for real, we should create new projects for it like we did for for 2017 and 2019

if we are just using the old projects with the old toolset hardcoded in them why are we bothering?

tculotta: The GDK instructions around VS2022 support specifically include getting the v141 and v142 toolsets into the install location for the GDK libs to use. And with the conversations with @jasonsandlin and Andy, we ended up deciding on moving forward with the 142 libs rather than create everything made on 143

lubeltra: what about all the non gdk consumers? what is the timeline for creating the v143 projects?

amccalib: Right now, we don't have a timeline for that. We're unlikely to take it on before the new year, as the scope of the current work was constrained to ensuring we're aligned with the needs of the GDK. If there are urgent needs for VS2022 support outside the GDK, we can figure out how to prioritize that work, but right now, I'm not aware of those needs.

tculotta: That's definitely a good question that I'm still not clear on. I'm fine going all the way with v143 btw, but it seems like there will end up being some conversation on this.

jplafonta: Maybe its worth making this new property conditional on HCLibPlatform = GDK? So that we don't unintentionally break someone that picks this up on other platforms with vs2022

tculotta: Done. but to be clear, if someone wants to use this props file building on the v143 toolset, they'll still need to modify it themselves for now anyway

jplafonta (approved):
:shipit:

PR #643, opened 2021-11-29
Make XCurl HTTP Provider cleanup fully asynchronous

jplafonta:
* Fixes deadlock that occurs if callers use one single threaded manually dispatched queue for both HTTP calls and HCCleanupAsync
* Layered such that other HTTP providers can support async cleanup in the future

### DIFF
--- a/Source/Global/perform_env.cpp
+++ b/Source/Global/perform_env.cpp
@@ -221,3 +221,56 @@ Result<HC_UNIQUE_PTR<HC_PERFORM_ENV>> HC_PERFORM_ENV::Initialize(HCInitArgs* arg
 
     return std::move(performEnv);
 }
+
+HRESULT HC_PERFORM_ENV::CleanupAsync(HC_UNIQUE_PTR<HC_PERFORM_ENV>&& env, XAsyncBlock* async) noexcept
+{
+    RETURN_IF_FAILED(XAsyncBegin(async, env.get(), __FUNCTION__, __FUNCTION__, CleanupAsyncProvider));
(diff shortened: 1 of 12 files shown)


natiskan (approved):

Source/Global/perform_env.cpp
```diff
@@ -221,3 +221,56 @@ Result<HC_UNIQUE_PTR<HC_PERFORM_ENV>> HC_PERFORM_ENV::Initialize(HCInitArgs* arg
...
+        {
+            HC_UNIQUE_PTR<XAsyncBlock> curlCleanupAsyncBlock{ async };
+            XAsyncBlock* envCleanupAsyncBlock = static_cast<XAsyncBlock*>(curlCleanupAsyncBlock->context);
+            
+            HRESULT cleanupResult = XAsyncGetStatus(curlCleanupAsyncBlock.get(), false);
+            curlCleanupAsyncBlock.reset();
+
+            // HC_PERFORM_ENV fully cleaned up at this point
+            XAsyncComplete(envCleanupAsyncBlock, cleanupResult, 0);
+        };
```
jasonsandlin, line 254: maybe pass S_OK regardless and log failure instead?

jasonsandlin (approved):
:shipit:

Source/Global/perform_env.cpp
jplafonta: I think either way is fine.  I've bubbled the error up to the top level and logged it there.  In any case cleanup will complete successfully.

PR #648, opened 2021-12-17
Fix for Android JNI link bug

SahilAshar:


Source/HTTP/Android/android_http_request.cpp
```diff
@@ -1,9 +1,16 @@
 #include "pch.h"
 
 #include "android_http_request.h"
+#include "http_android.h"
 #include <httpClient/httpClient.h>
 #include <vector>
 
+// The following symbols live in files that are not referenced by any other c
```
sashaweiss-msft, line 8: nit: capitalize `c` at the end here

Source/HTTP/Android/http_android.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
```
sashaweiss-msft, line 1: do we need a header for this? i guess we probably do if we want to reference in the other file, yeah?

Source/HTTP/Android/android_http_request.cpp
```diff
@@ -1,9 +1,16 @@
...
+// file and are only called by JNI callers. In order to force the compiler to
+// not ignore the file we make a token dummy reference to the file here.
+volatile auto dummyHCRequestOnRequestCompleted = &Java_com_xbox_httpclient_HttpClientRequest_OnRequestCompleted;
```
sashaweiss-msft, line 11: why only do symbols for two of them? woulda thought we'd do one or all

Source/HTTP/Android/http_android.h
SahilAshar: yeah, can't reference the JNI calls in the other file otherwise

Source/HTTP/Android/android_http_request.cpp
SahilAshar: That's what I thought too, but the other two functions are linking just fine, so I didn't bother

sashaweiss-msft: do we need both referenced here for both to link properly?

SahilAshar: yes, `OnRequestCompleted` and `OnRequestFailed` need to be referenced otherwise the link won't happen

Source/HTTP/Android/http_android.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include <jni.h>
+
+extern "C"
```
sashaweiss-msft, line 5: can we leave a comment in this file explaining that we're exposing these so we can reference them in another file to prevent them being optimized out?

sashaweiss-msft (approved):

jplafonta (approved):
:shipit:

PR #653, opened 2022-01-12
Proper handling of large incoming messages for WinHttp WebSockets 

jplafonta:
* Fixes bug where messages that exceeded the hardcoded receive buffer size were never delivered to clients
* Added public API to set a message fragment handler. When messages are broken down at the platform level and exceed the configured receive buffer size, they will be passed to clients in fragments.
* Adds public API to configure maximum WebSocket receive buffer size.  When receive buffer is full, partial messages are passed to client. 
* Adds public API to configure the maximum receive buffer size (WinHttp only).  If large messages are expected, this can be changed so that full messages are delivered all at once rather than in chunks.  Setting this buffer size to large values does mean that the entire message will be stored in LHC memory until it is passed along to clients so memory performance may be affected.

### DIFF
--- a/Include/httpClient/httpClient.h
+++ b/Include/httpClient/httpClient.h
@@ -989,6 +989,21 @@ STDAPI HCWebSocketDisconnect(
     _In_ HCWebsocketHandle websocket
     ) noexcept;
 
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK
+/// <summary>
+/// Configures how large the WebSocket receive buffer is allowed to grow before passing messages to clients. If a single message
+/// exceeds the maximum buffer size, the message will be broken down and passed to clients via multiple calls to the HCWebSocketMessageFunction.
+/// The default value is 20kb.
+/// </summary>
+/// <param name="websocket">The handle of the WebSocket</param>
+/// <param name="bufferSizeInBytes">Maximum size (in bytes) for the WebSocket receive buffer.</param>
+/// <returns>Result code for this API operation.  Possible values are S_OK, E_INVALIDARG, or E_FAIL.</returns>
+STDAPI HCWebSocketSetMaxReceiveBufferSize(
+    _In_ HCWebsocketHandle websocket,
+    _In_ size_t bufferSizeInBytes
+) noexcept;
+#endif
+
 /// <summary>
 /// Increments the reference count on the call object.
 /// </summary>
--- a/Include/httpClient/httpClient.h
+++ b/Include/httpClient/httpClient.h
@@ -774,14 +774,33 @@ typedef void
 /// A callback invoked every time a WebSocket receives an incoming binary message
 /// </summary>
 /// <param name="websocket">Handle to the WebSocket that this message was sent to</param>
-/// <param name="incomingBodyPayload"></param>
-/// <param name="incomingBodyPayloadSize"></param>
+/// <param name="incomingBodyPayload">Binary message payload.</param>
+/// <param name="incomingBodyPayloadSize">Size of the payload in bytes.</param>
 /// <param name="functionContext">Client context to pass to callback function.</param>
 typedef void
 (CALLBACK* HCWebSocketBinaryMessageFunction)(
     _In_ HCWebsocketHandle websocket,
     _In_reads_bytes_(payloadSize) const uint8_t* payloadBytes,
     _In_ uint32_t payloadSize,
+    _In_ void* functionContext
+    );
+
+/// <summary>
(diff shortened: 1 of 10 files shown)


Source/HTTP/WinHttp/winhttp_connection.cpp
```diff
@@ -1334,73 +1333,29 @@ void WinHttpConnection::callback_websocket_status_read_complete(
...
-        void* functionContext = nullptr;
-
+        bool fragmentReadComplete{ false };
         {
             win32_cs_autolock autoCriticalSection(&pRequestContext->m_lock);
 
-            pRequestContext->m_websocketResponseBuffer.FinishWriteData(wsStatus->dwBytesTransferred);
-            pRequestContext->m_websocketResponseBuffer.TransferBuffer(&responseBuffer);
-            HCWebSocketGetEventFunctions(pRequestContext->m_websocketHandle, &messageFunc, nullptr, nullptr, &functionContext);
+            // WinHttp doesn't report the chunk of a fragmented message as a fragment, but LHC will
```
jplafonta, line 1340: >chunk [](http://example.com/codeflow?start=42&length=5)

should say finalChunk. Will update

Source/WebSocket/hcwebsocket.h
```diff
@@ -88,11 +92,13 @@ typedef struct HC_WEBSOCKET : std::enable_shared_from_this<HC_WEBSOCKET>
     http_internal_string m_proxyUri;
     http_internal_string m_uri;
     http_internal_string m_subProtocol;
+    size_t m_maxReceiveBufferSize;
```
jasonsandlin, line 96: { 0 }

Include/httpClient/httpClient.h
```diff
@@ -882,6 +917,18 @@ STDAPI HCWebSocketGetEventFunctions(
...
     ) noexcept;
 
+/// <summary>
+/// Gets the WebSocket binary message fragment handler.
+/// </summary>
+/// <param name="websocket">The handle of the websocket.</param>
+/// <param name="binaryMessageFragmentFunc">Returned binaryMessageFragmentFunc.</param>
+/// <param name="functionContext">Client context to pass to callback function.</param>
+/// <returns>Result code for this API operation.  Possible values are S_OK, E_INVALIDARG, E_HC_NOT_INITIALISED, or E_FAIL.</returns>
+STDAPI HCWebSocketGetBinaryMessageFragmentEventFunction(
```
jasonsandlin, line 927: >HCWebSocketGetBinaryMessageFragmentEventFunction [](http://example.com/codeflow?start=7&length=48)

Include/httpClient/httpClient.h
```diff
@@ -824,6 +843,22 @@ STDAPI HCWebSocketCreate(
...
+/// </summary>
+/// <param name="websocket">The handle of the websocket.</param>
+/// <param name="binaryMessageFragmentFunc">A pointer to the binary message fragment handling callback to use, or a null pointer to remove.</param>
+/// <returns>Result code for this API operation.  Possible values are S_OK, E_INVALIDARG, or E_FAIL.</returns>
+/// <remarks>
+/// If this handler is not set, messages larger than the configured buffer size may still be broken down and passed to the
+/// HCWebSocketBinaryMessageFunction, but there will be no indication that they are partial messages. If large WebSocket messages are expected,
+/// it is recommended to either set this handler OR set a receive buffer large enough to hold the entire message.
+/// </remarks>
+STDAPI HCWebSocketSetBinaryMessageFragmentEventFunction(
```
jasonsandlin, line 857: >HCWebSocketSetBinaryMessageFragmentEventFunction [](http://example.com/codeflow?start=7&length=48)

Source/WebSocket/hcwebsocket.cpp
```diff
@@ -373,6 +388,12 @@ HRESULT HC_WEBSOCKET::SetHeader(
     return S_OK;
 }
 
+HRESULT HC_WEBSOCKET::SetMaxReceiveBufferSize(size_t maxReceiveBufferSizeBytes) noexcept
+{
+    m_maxReceiveBufferSize = maxReceiveBufferSizeBytes;
+    return S_OK;
```
jasonsandlin, line 394: is this used anywhere?

jasonsandlin (changes requested):
:clock1:

Source/WebSocket/hcwebsocket.cpp
jplafonta: its returned to clients via HCWebSocketSetMaxReceiveBufferSize. Currently this method can't fail but I think its fine to keep this signature in case that changes in the future

jasonsandlin (approved):

PR #656, opened 2022-01-24
WinHttp Cleanup race condition

jplafonta:
Fixes a race condition be HCCleanupAsync and a WinHttpClose call by adding a final "Closed" state to connection.

### DIFF
--- a/Source/HTTP/Curl/CurlMulti.cpp
+++ b/Source/HTTP/Curl/CurlMulti.cpp
@@ -87,17 +87,7 @@ HRESULT CALLBACK CurlMulti::CleanupAsyncProvider(XAsyncOp op, const XAsyncProvid
     case XAsyncOp::Begin:
     {
         CurlMulti* multi = static_cast<CurlMulti*>(data->context);
-        RETURN_IF_FAILED(XTaskQueueTerminate(multi->m_queue, false, multi, [](void* context)
-        {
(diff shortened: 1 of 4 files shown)


jasonsandlin (approved):

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -87,17 +87,7 @@ HRESULT CALLBACK CurlMulti::CleanupAsyncProvider(XAsyncOp op, const XAsyncProvid
...
-            HC_UNIQUE_PTR<CurlMulti> multi{ static_cast<CurlMulti*>(context) };
-
-            // Ensure CurlMulti is destroyed (and thus curl_multi_cleanup is called) before completing asyncBlock
-            XAsyncBlock* cleanupAsync = multi->m_cleanupAsyncBlock;
-            multi.reset();
-
-            XAsyncComplete(cleanupAsync, S_OK, 0);
-        }));
-
+        RETURN_IF_FAILED(XTaskQueueTerminate(multi->m_queue, false, multi, CurlMulti::TaskQueueTerminated));
```
jplafonta, line 90: >RETURN_IF_FAILED(XTaskQueueTerminate(multi->m_queue, false, multi, CurlMulti::TaskQueueTerminated)); [](http://example.com/codeflow?start=8&length=100)

This is not a functional change, but it makes debugging this callback with VS possible (for some reason breakpoints cannot be set in this inline lambda)

PR #657, opened 2022-01-24
Use a safe binary staging dir for iOS OpenSSL build

sashaweiss-msft:
The OpenSSL build script for iOS uses `$CONFIGURATION_TEMP_DIR` as a staging directory for artifacts, and then later cleans up that directory by deleting it wholesale. This is unsafe, as `$CONFIGURATION_TEMP_DIR` is shared across targets in a build, and deleting it may interfere with other parts of the build.

This PR updates the script to use an explicit "staging" directory that is safe to delete.


SahilAshar (approved):

matt-vanderkolk (approved):
:shipit:

PR #659, opened 2022-01-27
Prevent request body read to continue reading if no more data on XMLHttp

jgalvan:
On UWP, per [ISequentialStream::Read](https://docs.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-isequentialstream-read) documentation, the Read method should return S_FALSE if the value returned in pcbRead is less than the number of bytes requested in cb (end of the stream has been reached).

This PR also adds a couple of extra checks to return a more appropriate error code for some failure scenarios.

### DIFF
--- a/Source/HTTP/XMLHttp/http_request_stream.cpp
+++ b/Source/HTTP/XMLHttp/http_request_stream.cpp
@@ -51,6 +51,11 @@ HRESULT STDMETHODCALLTYPE http_request_stream::Read(
     _Out_ ULONG *pcbRead
     )
 {
+    if (pv == nullptr || pcbRead == nullptr)
+    {
+        return STG_E_INVALIDPOINTER;
+    }
+
     HCHttpCallRequestBodyReadFunction readFunction = nullptr;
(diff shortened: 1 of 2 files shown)


Source/HTTP/XMLHttp/http_request_stream.cpp
```diff
@@ -60,6 +65,13 @@ HRESULT STDMETHODCALLTYPE http_request_stream::Read(
         return STG_E_READFAULT;
     }
 
+    if (m_startIndex == bodySize)
+    {
+        // Reached end of buffer
+        *pcbRead = 0;
+        return S_FALSE;
```
lubeltra, line 72: is returning S_FALSE how we signal end of buffer?

lubeltra (approved):

Source/HTTP/XMLHttp/http_request_stream.cpp
jgalvan: It is

PR #662, opened 2022-02-03
Cleanup redesign part 2

jplafonta:
* WebSocket class reworked allowing multiple "observers" to independently register for events.  This allows both clients and HC_PERFORM_ENV to keep track of the connection state
* HCWebSocketConnectAsync Shim added to HC_PERFORM_ENV, tracking WebSocket connect attempts
  * During cleanup, WebSockets in the process of connecting will be allowed to finish connecting (currently no support for cancellation)
  * Connected WebSockets will be properly closed
* Validated E2E using local WebSocket echo server + API runner scripts (added in separate PR)
* Regression tested via API runner

### DIFF
--- a/Samples/Win32WebSocket/main.cpp
+++ b/Samples/Win32WebSocket/main.cpp
@@ -3,6 +3,7 @@
 
 #include "pch.h"
 #include <atomic>
+#include <assert.h>
 
 class win32_handle
 {
@@ -225,7 +226,9 @@ int main()
     asyncBlock->callback = [](XAsyncBlock* asyncBlock)
     {
         WebSocketCompletionResult result = {};
-        HCGetWebSocketConnectResult(asyncBlock, &result);
+        HRESULT hr = HCGetWebSocketConnectResult(asyncBlock, &result);
+        assert(SUCCEEDED(hr));
+        UNREFERENCED_PARAMETER(hr);
 
         printf_s("HCWebSocketConnect complete: %d, %d\n", result.errorCode, result.platformErrorCode);
         SetEvent(g_eventHandle);
@@ -252,7 +255,9 @@ int main()
         asyncBlock->callback = [](XAsyncBlock* asyncBlock)
         {
             WebSocketCompletionResult result = {};
-            HCGetWebSocketSendMessageResult(asyncBlock, &result);
+            HRESULT hr = HCGetWebSocketSendMessageResult(asyncBlock, &result);
+            assert(SUCCEEDED(hr));
+            UNREFERENCED_PARAMETER(hr);
 
             printf_s("HCWebSocketSendMessage complete: %d, %d\n", result.errorCode, result.platformErrorCode);
             SetEvent(g_eventHandle);
@@ -267,7 +272,9 @@ int main()
         asyncBlock->callback = [](XAsyncBlock* asyncBlock)
         {
             WebSocketCompletionResult result = {};
-            HCGetWebSocketSendMessageResult(asyncBlock, &result);
+            HRESULT hr = HCGetWebSocketSendMessageResult(asyncBlock, &result);
+            assert(SUCCEEDED(hr));
+            UNREFERENCED_PARAMETER(hr);
 
             printf_s("HCWebSocketSendBinaryMessageAsync complete: %d, %d\n", result.errorCode, result.platformErrorCode);
             SetEvent(g_eventHandle);
--- a/Samples/Win32WebSocket/main.cpp
+++ b/Samples/Win32WebSocket/main.cpp
@@ -218,80 +218,77 @@ int main()
     HRESULT hr = HCWebSocketCreate(&websocket, message_received, binary_message_received, websocket_closed, nullptr);
     HCWebSocketSetBinaryMessageFragmentEventFunction(websocket, binary_message_fragment_received);
 
-    for (int iConnectAttempt = 0; iConnectAttempt < 10; iConnectAttempt++)
-    {
-        g_numberMessagesReceieved = 0;
+    g_numberMessagesReceieved = 0;
 
-        XAsyncBlock* asyncBlock = new XAsyncBlock{};
+    XAsyncBlock* asyncBlock = new XAsyncBlock{};
+    asyncBlock->queue = g_queue;
+    asyncBlock->callback = [](XAsyncBlock* asyncBlock)
+    {
+        WebSocketCompletionResult result = {};
+        HCGetWebSocketConnectResult(asyncBlock, &result);
+
+        printf_s("HCWebSocketConnect complete: %d, %d\n", result.errorCode, result.platformErrorCode);
+        SetEvent(g_eventHandle);
+        delete asyncBlock;
+    };
+
+    printf_s("Calling HCWebSocketConnect...\n");
+    hr = HCWebSocketConnectAsync(url.data(), "", websocket, asyncBlock);
+    WaitForSingleObject(g_eventHandle, INFINITE);
+
+    uint32_t numberOfMessagesToSend = 1;
+    for (uint32_t i = 1; i <= numberOfMessagesToSend; i++)
+    {
+        char webMsg[150000];
+        for (uint32_t j = 0; j < sizeof(webMsg); ++j)
+        {
+            webMsg[j] = 'X';
+        }
+        webMsg[sizeof(webMsg) - 1] = '\0';
+        snprintf(webMsg, sizeof(webMsg), "Message #%d should be echoed!", i);
+
+        asyncBlock = new XAsyncBlock{};
         asyncBlock->queue = g_queue;
         asyncBlock->callback = [](XAsyncBlock* asyncBlock)
         {
             WebSocketCompletionResult result = {};
-            HCGetWebSocketConnectResult(asyncBlock, &result);
+            HCGetWebSocketSendMessageResult(asyncBlock, &result);
 
-            printf_s("HCWebSocketConnect complete: %d, %d\n", result.errorCode, result.platformErrorCode);
+            printf_s("HCWebSocketSendMessage complete: %d, %d\n", result.errorCode, result.platformErrorCode);
             SetEvent(g_eventHandle);
             delete asyncBlock;
         };
 
-        printf_s("Calling HCWebSocketConnect...\n");
-        hr = HCWebSocketConnectAsync(url.data(), "", websocket, asyncBlock);
-        WaitForSingleObject(g_eventHandle, INFINITE);
+        printf_s("Calling HCWebSocketSend with message \"%s\" and waiting for response...\n", webMsg);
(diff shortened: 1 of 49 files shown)


Samples/Win32WebSocket/main.cpp
```diff
@@ -218,80 +218,77 @@ int main()
...
+        snprintf(webMsg, sizeof(webMsg), "Message #%d should be echoed!", i);
 
-        XAsyncBlock* asyncBlock = new XAsyncBlock{};
+        asyncBlock = new XAsyncBlock{};
         asyncBlock->queue = g_queue;
         asyncBlock->callback = [](XAsyncBlock* asyncBlock)
         {
             WebSocketCompletionResult result = {};
-            HCGetWebSocketConnectResult(asyncBlock, &result);
+            HCGetWebSocketSendMessageResult(asyncBlock, &result);
```
jasonsandlin, line 255: >HCGetWebSocketSendMessageResult [](http://example.com/codeflow?start=12&length=31)

check hr here and below Result().  

in this file, i would add asserts for all the hrs just for extra checking

Source/Global/perform_env.cpp
```diff
@@ -294,8 +307,175 @@ HRESULT CALLBACK HC_PERFORM_ENV::HttpPerformAsyncShimProvider(XAsyncOp op, const
...
+        if (internalAsyncBlock.queue)
+        {
+            XTaskQueueCloseHandle(internalAsyncBlock.queue);
+        }
+    }
+
+    HC_PERFORM_ENV* const env{};
+    http_internal_string uri;
+    http_internal_string subprotocol;  
+    std::shared_ptr<WebSocket> websocket;
```
jasonsandlin, line 341: >std::shared_ptr<WebSocket> [](http://example.com/codeflow?start=4&length=26)

does this need to use mem hook type?

Source/Global/perform_env.cpp
```diff
@@ -294,8 +307,175 @@ HRESULT CALLBACK HC_PERFORM_ENV::HttpPerformAsyncShimProvider(XAsyncOp op, const
...
+    if (disconnect)
+    {
+        hr = context->websocket->Disconnect();
+        if (FAILED(hr))
+        {
+            HC_TRACE_ERROR_HR(HTTPCLIENT, hr, "WebSocket::Disconnect failed during HCCleanup");
+        }
+    }
+    else if (scheduleProviderCleanup)
+    {
```
jasonsandlin, line 444: i see the assert but this might be a little safer if this was just else {} instead of else if {}.

jasonsandlin (approved):
:shipit:

Source/Global/perform_env.cpp
jplafonta: the type doesn't need to change, but it is allocated with the mem-hooked allocated in WebSocket::Initialize.

Source/Global/perform_env.cpp
jplafonta: I actually think "else if" is safer. If somehow the assertion is false and both 'disconnect' and 'scheduleProviderCleanup' are true, ProviderCleanup would scheduled twice - once here and once when disconnect completes.

Source/Global/perform_env.cpp
```diff
@@ -262,6 +265,10 @@ HRESULT CALLBACK HC_PERFORM_ENV::HttpPerformAsyncShimProvider(XAsyncOp op, const
         RETURN_IF_FAILED(XTaskQueueGetPort(data->async->queue, XTaskQueuePort::Work, &workPort));
         RETURN_IF_FAILED(XTaskQueueCreateComposite(workPort, workPort, &performContext->internalAsyncBlock.queue));
 
+        std::unique_lock<std::mutex> lock{ env->m_mutex };
```
lubeltra, line 268: why not use a scope to unlock?

Source/Global/perform_env.cpp
```diff
@@ -217,7 +218,7 @@ Result<HC_UNIQUE_PTR<HC_PERFORM_ENV>> HC_PERFORM_ENV::Initialize(HCInitArgs* arg
     return std::move(performEnv);
 }
 
-struct HC_PERFORM_ENV::HttpPerformContext
+struct HC_PERFORM_ENV::HttpPerformContext : public std::enable_shared_from_this<HC_PERFORM_ENV::HttpPerformContext>
```
lubeltra, line 221: is this used anywhere?

Source/Global/perform_env.cpp
lubeltra: I don't follow that logic at all:
the assert says there's 3 valid paths through this code:
1. only `disconnect == true`
2. only `scheduleProviderCleanup == true`
3. neither is true

(ie at least one of `disconnect`, `scheduleProviderCleanup` must be false)

Changing the `else if` into and `if` collapses case 2 and 3 into going into the branch which is obviously a massive and incorrect behavior change.

But your comment implies that at least one of them must be true, which we are not asserting at all right now, so which is it?

Source/Global/perform_env.cpp
```diff
@@ -315,58 +499,111 @@ HRESULT CALLBACK HC_PERFORM_ENV::CleanupAsyncProvider(XAsyncOp op, const XAsyncP
...
+#else
+    UNREFERENCED_PARAMETER(async);
+    assert(false);
+#endif
+}
+
+bool HC_PERFORM_ENV::CanScheduleProviderCleanup()
+{
+    // mutex should always be held when calling this method
+    assert(!m_mutex.try_lock());
```
lubeltra, line 585: don't do this, it will crash on some platforms
or at least wrap it in
```c++
#if !HC_PLATFORM_IS_EXTERNAL
```

Samples/Win32WebSocket/main.cpp
```diff
@@ -218,80 +219,83 @@ int main()
     HRESULT hr = HCWebSocketCreate(&websocket, message_received, binary_message_received, websocket_closed, nullptr);
     HCWebSocketSetBinaryMessageFragmentEventFunction(websocket, binary_message_fragment_received);
 
-    for (int iConnectAttempt = 0; iConnectAttempt < 10; iConnectAttempt++)
+    g_numberMessagesReceieved = 0;
```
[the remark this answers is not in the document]
jplafonta: odd indeed, will update that

Samples/Win32WebSocket/main.cpp
```diff
@@ -218,80 +219,83 @@ int main()
...
-            WaitForSingleObject(g_eventHandle, INFINITE);
-        }
+        hr = HCWebSocketSendBinaryMessageAsync(websocket, (uint8_t*)webMsg, 100, asyncBlock);
+    }
 
-        printf_s("Calling HCWebSocketDisconnect...\n");
-        HCWebSocketDisconnect(websocket);
+    while (g_numberMessagesReceieved < numberOfMessagesToSend * 2)
+    {
         WaitForSingleObject(g_eventHandle, INFINITE);
```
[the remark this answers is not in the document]
jplafonta: At least here we are awaiting messages from the WebSocket echo server.  Its not actually an XAsync call.

Samples/Win32WebSocket/main.cpp
```diff
@@ -218,80 +219,83 @@ int main()
...
     }
 
+    printf_s("Calling HCWebSocketDisconnect...\n");
+    HCWebSocketDisconnect(websocket);
```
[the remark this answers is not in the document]
jplafonta: There are two versions: HCCleanup and HCCleaupAsync.  For a long time only the synchronous version existed.  When we added the async version, we changed HCCleanup to effectively be HCCleanupAsync + XAsyncGetStatus w/ wait. 

The CloseHandle call isn't waiting on anything, it just closing the event handle.

Source/Global/perform_env.cpp
jplafonta: I can change.  Is that better somehow or just style preference?

Source/Global/perform_env.cpp
jplafonta: line 247

Source/Global/perform_env.cpp
jplafonta: oh good to know. Will update

Samples/Win32WebSocket/main.cpp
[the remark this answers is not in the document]
jplafonta: realized its not a function local but in fact a global :P

Source/Global/perform_env.cpp
jplafonta: is there a safe xplat way to assert that the mutex is held?

Samples/Win32WebSocket/main.cpp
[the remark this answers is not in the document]
lubeltra: oh right, duh

Source/Global/perform_env.cpp
lubeltra: it's more obvious what is protected by the lock and what isn't

on the other hand I see a lot of code here using `.unlock()` so it may not be worthwhile to change

Samples/Win32WebSocket/main.cpp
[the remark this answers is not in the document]
lubeltra: There's ways to make this better but they involve a large scale redesign of the api so probably not worth it.

Source/Global/perform_env.cpp
lubeltra: no :(

lubeltra: > If try_lock is called by a thread that already owns the mutex, the behavior is undefined.

jplafonta: I'll just remove this altogether

Source/Global/perform_env.cpp
jplafonta: Maybe my comment was unclear.

As the assertion implies at least one of these must be false.  All WebSockets need to be disconnected before cleanup happens.  If 'disconnect' is true, scheduleProviderCleanup should be false.  

What I was trying to convey in my last comment was that if somehow both of these ended up true (due to a bug somewhere) keeping it as an 'else if' protects against both paths executing, leading to ProviderCleanup being schedule twice.

lubeltra: both `else` and `else if` do that
they will only differ is both flags are false

jplafonta: Doh....I read Jason's original comment wrong.  I thought he was suggesting changing the 'else if' to just 'if'.

'else' definitely doesn't make sense.  It is a perfectly valid scenario for them to both be false.

lubeltra: cool cool

Source/Global/perform_env.cpp
jplafonta: keeping as is for consistency

lubeltra (approved):

PR #667, opened 2022-02-17
Adding HC_NOWEBSOCKETS around more socket operations

tculotta:


Source/HTTP/WinHttp/winhttp_connection.cpp
```diff
@@ -1461,6 +1483,13 @@ HRESULT WinHttpConnection::WebSocketReadComplete(bool binaryMessage, bool isFrag
...
+#else
+    UNREFERENCED_PARAMETER(binaryMessage);
+    UNREFERENCED_PARAMETER(isFragment);
+    UNREFERENCED_PARAMETER(isFinalFragment);
+    assert(false);
+    return E_FAIL;
+#endif
 }
 
 void WinHttpConnection::callback_websocket_status_headers_available(
```
tculotta, line 1495: Do we want checks around this? This is used along with other websocket callbacks from L:1044 to L:1099 along with m_websocketHandle. What about the m_websocketHandle ref on L:466?

jplafonta: you could add a sanity check here (just assert false if HC_NOWEBSOCKETS like the other WebSocket callbacks.  I think line 466 is fine.  m_websocketHandle being null differentiates an HTTP call from a WebSocket call so I think that check is fine.

jplafonta (approved):
:shipit:

PR #668, opened 2022-02-28
WinHttp Cleanup race condition (#656)

natiskan:
Fixes a race condition be HCCleanupAsync and a WinHttpClose call by adding a final "Closed" state to connection.


jplafonta (approved):
:shipit:

PR #669, opened 2022-03-03
Fix race condition in CurlProvider::CleanupAsync

jplafonta:
During cleaning, CurlProvider loops over its CurlMultis and kicks off CurlMulti::CleanupAsync for each of them.  When the last of those sub-tasks completes, the Provider will be freed and then the CurlProvider::CleanupAsync task will be completed.  There is a race condition such that the final CurlMulti cleanup may complete _before_ CurlProvider::CleanupAsyncProvider's XAsyncOp::Begin.  This race exposes a couple of nasty bugs: 
* Due to a tricky detail of how c++ move semantics work, ownership of the CurlMulti isn't properly transferred to CurlMulti::CleanupAsync, leading to the CurlMulti object (and its curl_multi handle) being cleaned up twice: once as part of CurlMulti::CleanupAsync (expected), and then a second time when the final MultiCleanupComplete fires while the CurlMulti object is still in the CurlProvider's m_curlMultis map (unexpected).  The fix is to properly transfer ownership of the CurlMulti by passing it by value rather than r-value reference.
* Because CurlProvider is looping over its member m_curlMultis, we need to ensure that the provider isn't destroyed before that loop completes.  The fix for this is to gate cleanup of the CurlProvider on not only the CurlMulti cleanup operations completing, but also on CurlProvider::CleanupAsyncProvider having exited that cleanup loop.

I also fixed a couple of other spots where ownership of unique_ptrs wasn't properly transferred, even though they didn't directly lead to bugs.

### DIFF
--- a/Source/HTTP/Curl/CurlProvider.cpp
+++ b/Source/HTTP/Curl/CurlProvider.cpp
@@ -41,8 +41,8 @@ Result<HC_UNIQUE_PTR<CurlProvider>> CurlProvider::Initialize()
 CurlProvider::~CurlProvider()
 {
     // Either CleanupAsync was never called or CurlProvider shouldn't be destroyed until it completes.
-    assert(!m_pendingMultiCleanups);
-    
+    assert(!m_cleanupTasksRemaining);
+
     if (m_multiCleanupQueue)
     {
         XTaskQueueCloseHandle(m_multiCleanupQueue);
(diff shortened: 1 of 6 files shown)


Source/HTTP/Curl/CurlProvider.cpp
```diff
@@ -124,7 +124,7 @@ HRESULT CALLBACK CurlProvider::CleanupAsyncProvider(XAsyncOp op, const XAsyncPro
 
         {
             std::lock_guard<std::mutex> lock{ provider->m_mutex };
-            provider->m_pendingMultiCleanups = provider->m_curlMultis.size();
+            provider->m_cleanupTasksRemaining = 1 + provider->m_curlMultis.size();
```
ScottMunroMS, line 127: nit: it might be worth a comment here describing what the +1 is for

ScottMunroMS (approved):

Source/HTTP/Curl/CurlProvider.cpp
jplafonta: will do

natiskan (approved):

PR #670, opened 2022-03-04
Fix bad logic in WinHttp WebSocket fragment handling

jplafonta:
When WebSocket message fragments are received, they are collected in a buffer within LHC.  They are only forwarded to the client when that buffer is full or when the final fragment has been received.  When WinHttp has given us fragments, LHC incorrectly invokes the HCWebSocketBinaryMessageFragmentFunction even if we were able to accumulate the fragments in our buffer before passing them along to the client.

This also fixes a bug where the WebSocket's default m_maxReceiveBufferSize isn't correctly initialized to 20kb as documented in the header.

### DIFF
--- a/Samples/Win32WebSocket/main.cpp
+++ b/Samples/Win32WebSocket/main.cpp
@@ -4,6 +4,7 @@
 #include "pch.h"
 #include <atomic>
 #include <assert.h>
+#include <vector>
 
 class win32_handle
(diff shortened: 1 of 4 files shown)


jasonsandlin (approved):

jasonsandlin:
test or changes to existing test needed?

PR #680, opened 2022-03-30
Reintroducing websocket close behavior for Request_Error scenario

tculotta:


Source/HTTP/WinHttp/winhttp_connection.cpp
```diff
@@ -681,6 +681,21 @@ void WinHttpConnection::callback_status_request_error(
...
+            }
+
+            if (pRequestContext->m_asyncBlock == nullptr)
+            {
+                pRequestContext->on_websocket_disconnected(static_cast<USHORT>(errorCode));
+            }
+        }
+#endif
+
         pRequestContext->complete_task(E_FAIL, HRESULT_FROM_WIN32(errorCode));
```
jplafonta, line 699: It seems like complete_task and on_websocket_disconnected shouldn't both be called here.  complete_task completes the WebSocketConnect async operation, while on_websocket_disconnected invokes the client's disconnect handler.  I think we should only do one or the other depending on ConnectionState

jplafonta:
>         DWORD result = m_winHttpWebSocketExports.close(m_hRequest, static_cast<USHORT>(HCWebSocketCloseStatus::GoingAway), nullptr, 0);

I'm still concerned that we aren't handling failures anywhere for this call.  Were you able to check what's returned here when the network is disconnected?

---
Refers to: Source/HTTP/WinHttp/winhttp_connection.cpp:428 in 47b5804. [](commit_id = 47b58045a64397a459bb3e66b8b4caef305cf396, deletion_comment = False)

Source/HTTP/WinHttp/winhttp_connection.cpp
jplafonta: I'm also not sure about the transition to WebSocketClosing.  It seems like by the time we've gotten this error callback we're already closed and should just transition straight to WinHttpClosing (which I think is already done as part of on_websocket_disconnected)

tculotta: We're definitely already in the WebSocketClosing state in this scenario where the error is from a Close call, but couldn't we get this error from any other websocket function and in that case we'd be in some Connected state instead?

I can add that comment. It's definitely a somewhat strange flow, but I thought it was probably best to reintroduce this with minimal changes to be safe.

jplafonta (approved):
:shipit:

PR #682, opened 2022-04-07
Add network-detail logging around network failures

sashaweiss-msft:
For unknown reasons, libHttpClient on Android sometimes fails to perform HTTP requests (OkHttp fails with an instance of a `ConnectException`, `UnknownHostException`, or `SocketTimeoutException`) when it is known that other components in the app are able to make requests.

This PR adds significant logging of network details to try and identify in the wild, through logs via bug reports, any correlations between "libHttpClient failed to make a request" and details of what networks are available.

Specifically, on network failure libHttpClient will now synchronously collect details of the networks present on the device, such as if they have a proxy configured.

Additionally, libHttpClient will now also collect those details of networks as they change, via the `ConnectivityManager.registerNetworkCallback` asynchronous listener.

The goal of this logging is to identify why libHttpClient sometimes gets stuck in failing to make requests, and should be considered temporary in that once the bug is fixed this logging should be removed.

### DIFF
--- a/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequest.java
+++ b/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequest.java
@@ -30,7 +30,7 @@ public class HttpClientRequest {
     private static final OkHttpClient OK_CLIENT;
     private static final byte[] NO_BODY = new byte[0];
 
-    private Context appContext;
+    private final Context appContext;
     private Request.Builder requestBuilder;
 
     static {
@@ -148,6 +148,10 @@ public class HttpClientRequest {
 
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
             networkDetails.addSection("internetWasValidated", networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED));
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            networkDetails.addSection("isNotSuspended", networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_SUSPENDED));
         }
 
         return networkDetails.toString();
--- a/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequest.java
+++ b/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequest.java
@@ -2,12 +2,8 @@ package com.xbox.httpclient;
 
 import android.content.Context;
 import android.net.ConnectivityManager;
-import android.net.LinkProperties;
 import android.net.Network;
-import android.net.NetworkCapabilities;
 import android.os.Build;
-
-import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -15,8 +11,6 @@ import java.io.StringWriter;
 import java.net.ConnectException;
 import java.net.SocketTimeoutException;
 import java.net.UnknownHostException;
-import java.util.ArrayList;
-import java.util.List;
 
 import okhttp3.Call;
 import okhttp3.Callback;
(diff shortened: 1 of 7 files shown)


Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequest.java
```diff
@@ -74,6 +92,44 @@ public void onResponse(Call call, final Response response) {
         });
     }
 
+    private String GetAllNetworksInfo() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            return "API version too old - unable to get network info!";
+        }
+
+        ConnectivityManager cm = (ConnectivityManager)appContext.getSystemService(Context.CONNECTIVITY_SERVICE);
```
SahilAshar, line 100: is this to check if it's currently connected? Or what the current HTTP service is?

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/NetworkObserver.java
```diff
@@ -0,0 +1,145 @@
...
+
+public class NetworkObserver {
+    @NotNull private static String s_lastCapabilities = "";
+    @NotNull private static String s_lastLinkProperties = "";
+
+    @SuppressWarnings("unused")
+    public static void Initialize(Context appContext) {
+        ConnectivityManager cm = (ConnectivityManager) appContext.getSystemService(Context.CONNECTIVITY_SERVICE);
+
+        // Build a network request to query for all networks that "should" be
```
SahilAshar, line 25: Does this mean you're querying all available WiFi/cellular networks for the device and seeing which ones provide internet access?

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/NetworkObserver.java
```diff
@@ -0,0 +1,145 @@
...
+
+            return join(sections);
+        }
+
+        private static String join(Map<String, Boolean> sections) {
+            StringBuilder builder = new StringBuilder();
+
+            boolean firstEntry = true;
+            for (Map.Entry<String, Boolean> entry : sections.entrySet()) {
+                if (firstEntry) {
```
matt-vanderkolk, line 129: Can't you just test the string builder's length instead?

matt-vanderkolk (approved):

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientRequest.java
sashaweiss-msft: Yeah, the point of this PR is to gather and log as much info as we can about the state of the various networks the device is connected to. The `ConnectivityManager` is the Android entrypoint to that information, and this function is just designed to collect that info and marshal it into a string we can log.

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/NetworkObserver.java
sashaweiss-msft: Yup, more or less. We construct a `NetworkRequest` that has some requirements - the only one we're setting here is that the network theoretically be able to access the internet (the system won't check that the network *can* actually reach the internet, but will know if a network *should* be able to) - and then that acts essentially as a kind of filter/query that selects the networks we want to be notified about in our callback.

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/NetworkObserver.java
```diff
@@ -0,0 +1,145 @@
...
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                sections.put("isNotSuspended", capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_SUSPENDED));
+            }
+
+            return join(sections);
+        }
+
+        private static String checkLinkProperties(@NotNull LinkProperties properties) {
+            Map<String, Boolean> sections = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
sashaweiss-msft: Eh, largely redundant but I'd like to leave this as-is in case we eventually want to log other info out of the `LinkProperties`. I don't anticipate that we will, but this way it matches the method above in structure, and the overhead is pretty minimal.

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/NetworkObserver.java
sashaweiss-msft: Yeah, that would work too - that's a good thought!

PR #690, opened 2022-05-22
Fix deadlock in threadpool_stl

redbaron:
Short lived libHC instances were deadlocking because of insufficient checks upon entering condvar wait.

If large number of threads created, then threads could start execution with `m_terminate` being set to true already.
Waiting for on `m_wake` condvar without prior check of `m_terminate` causes deadlock


lubeltra (approved):

brianpepin (approved):

jasonsandlin (approved):

jasonsandlin:
Thanks for tracking this down and submitting the PR!

PR #692, opened 2022-05-25
Bug 39732497: LHC needs to handle null XAsyncBlock

tculotta:


natiskan (approved):

Source/Task/AsyncLib.cpp
```diff
@@ -430,6 +430,8 @@ static HRESULT AllocStateNoCompletion(_Inout_ XAsyncBlock* asyncBlock, _Inout_ A
 
 static HRESULT AllocState(_Inout_ XAsyncBlock* asyncBlock, _In_ size_t contextSize)
 {
+    RETURN_HR_IF(E_INVALIDARG, asyncBlock == nullptr);
```
brianpepin, line 433: In the GDK code guidelines we are asked to dereference null args instead of returning an error.  The idea is the resulting crash is much easier to diagnose than weeding down where an E_INVALIDARG came from.

jasonsandlin (changes requested):
Sounds like no change needed then

PR #693, opened 2022-05-27
Check allocated buffer size before copying response

redbaron:
Small safety check to ensure destination buffer has enough space.

### DIFF
--- a/Source/HTTP/httpcall_response.cpp
+++ b/Source/HTTP/httpcall_response.cpp
@@ -126,6 +126,11 @@ try
         return E_FAIL;
     }
 
+    if (call->responseBodyBytes.size() > bufferSize)
+    {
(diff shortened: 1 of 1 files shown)


jasonsandlin:
Can you put back memcpy_s on windows cause MS compilers yell about using secure version otherwise as I recall.  aka do:

    if (call->responseBodyBytes.size() > bufferSize)
    {
        return E_BOUNDS;
    }

#if HC_PLATFORM_IS_MICROSOFT
    memcpy_s(buffer, bufferSize, call->responseBodyBytes.data(), call->responseBodyBytes.size());
#else
    memcpy(buffer, call->responseBodyBytes.data(), call->responseBodyBytes.size());
#endif

redbaron:
@jasonsandlin , done

PR #699, opened 2022-06-08
Addressing several WebSocket bugs

jplafonta:
* WinHttp request error not correctly handled when a disconnect isn't acknowledged by the server
* Fixes race condition between WebSocket disconnect event handler invocation and unregistration when closing a WebSocket handle
* HCWebSocketConnectAsync will now pass back the client's WebSocket handle in the XAsync result rather than an internal handle


jasonsandlin (approved):

redbaron:
@jplafonta , many thanks for fixing all three so quickly. Will it be included in the next GDK release?

PR #700, opened 2022-06-12
Fixed hang in XblCleanupAsync

mr-efficient:
CurlProvider::PerformAsync can be called simultaneously from multiple threads, which leaves CurlProvider::m_curlMultis in a bad state.  That results in m_curlMultis.size() being incorrect, which results in m_cleanupTasksRemaining being incorrect, which results in XAsyncComplete never being called in CurlProvider cleanup.

I protect against this by protecting access to m_curlMultis with m_mutex.

### DIFF
--- a/Source/HTTP/Curl/CurlProvider.cpp
+++ b/Source/HTTP/Curl/CurlProvider.cpp
@@ -81,13 +81,21 @@ HRESULT CurlProvider::PerformAsync(HCCallHandle hcCall, XAsyncBlock* async) noex
     auto easyInitResult = CurlEasyRequest::Initialize(hcCall, async);
     RETURN_IF_FAILED(easyInitResult.hr);
 
-    auto iter = m_curlMultis.find(workPort);
-    if (iter == m_curlMultis.end())
-    {
-        auto multiInitResult = CurlMulti::Initialize(workPort);
-        RETURN_IF_FAILED(multiInitResult.hr);
+    http_internal_map<XTaskQueuePortHandle, HC_UNIQUE_PTR<xbox::httpclient::CurlMulti>>::iterator iter;
 
-        iter = m_curlMultis.emplace(workPort, multiInitResult.ExtractPayload()).first;
+	{
+        // CurlProvider::PerformAsync can be called simultaneously from multiple threads so we need to lock
+        // to prevent unsafe access to m_curlMultis
+		std::lock_guard<std::mutex> lock{ m_mutex };
+
+		iter = m_curlMultis.find(workPort);
+		if (iter == m_curlMultis.end())
+		{
+			auto multiInitResult = CurlMulti::Initialize(workPort);
+			RETURN_IF_FAILED(multiInitResult.hr);
+
+			iter = m_curlMultis.emplace(workPort, multiInitResult.ExtractPayload()).first;
+		}
     }
 
     auto& multi{ iter->second };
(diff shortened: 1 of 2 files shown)


mr-efficient:
One of the unit tests for our game hangs about 50% of the time on shutdown in XblCleanupAsync.  After a lot of breakpoint and printf debugging, I tracked the issue back to CurlProvider::m_curlMultis getting into a bad state:  The size returned by m_curlMultis.size() was different from the number of items returned by iterating from begin() to end().  I tracked _that_ back to CurlProvider::m_curlMultis being simultaneously accessed from multiple threads.

I'm not sure if protecting it with CurlProvider::m_mutex is the right fix, but it does seem to fix the hang.

Source/HTTP/Curl/CurlProvider.cpp
```diff
@@ -81,13 +81,21 @@ HRESULT CurlProvider::PerformAsync(HCCallHandle hcCall, XAsyncBlock* async) noex
...
     RETURN_IF_FAILED(easyInitResult.hr);
 
-    auto iter = m_curlMultis.find(workPort);
-    if (iter == m_curlMultis.end())
-    {
-        auto multiInitResult = CurlMulti::Initialize(workPort);
-        RETURN_IF_FAILED(multiInitResult.hr);
+    http_internal_map<XTaskQueuePortHandle, HC_UNIQUE_PTR<xbox::httpclient::CurlMulti>>::iterator iter;
+
+	{
```
jplafonta, line 86: >	 [](http://example.com/codeflow?start=0&length=1)

Please replace all tabs with spaces

Source/HTTP/Curl/CurlProvider.cpp
```diff
@@ -133,28 +141,26 @@ HRESULT CALLBACK CurlProvider::CleanupAsyncProvider(XAsyncOp op, const XAsyncPro
...
-            {
-                // If there are no pending pending multi cleanups, complete cleanup here
-                cleanupComplete = true;
-            }
-        }
+		{
+			std::lock_guard<std::mutex> lock{ provider->m_mutex };
+			for (auto& pair : provider->m_curlMultis)
+			{
+				HRESULT hr = CurlMulti::CleanupAsync(std::move(pair.second), &provider->m_multiCleanupAsyncBlocks[multiIndex++]);
```
jplafonta, line 148: >HRESULT hr = CurlMulti::CleanupAsync [](http://example.com/codeflow?start=4&length=36)

I believe it is not safe to call CurlMulti::CleanupAsync while holding m_mutex here.  CleanupAsync can complete synchronously in same cases, calling back into CurlProvider::MultiCleanupComplete which will try to grab the mutex.

CurlProvider::CleanupAsyncProvider has started, m_curlMultis should never be modified by another thread again - it should be safe to not lock here.

That said, to be safe we can grab the mutex, move m_curlMultis into a local, then release the lock and continue with cleanup as before but instead operating on the local curlMultis list

Source/HTTP/Curl/CurlProvider.cpp
mr-efficient: Done.

Source/HTTP/Curl/CurlProvider.cpp
mr-efficient: Okay, I've put up a revised version.  Thanks.

jplafonta (approved):
:shipit:

PR #706, opened 2022-07-20
Adding support for /Zc:strictStrings

jasonsandlin:
from https://docs.microsoft.com/en-us/cpp/build/reference/zc-strictstrings-disable-string-literal-type-conversion?view=msvc-170

If /Zc:strictStrings is specified, the compiler enforces the standard C++ const qualifications for string literals, as type 'array of const char' or 'array of const wchar_t', depending on the declaration. String literals are immutable, and an attempt to modify the contents of one results in an access violation error at run time. You must declare a string pointer as const to initialize it by using a string literal, or use an explicit const_cast to initialize a non-const pointer.


jplafonta (approved):
:shipit:

PR #714, opened 2022-09-10
Use okhttp for websockets on android

v-davwes:
On android, libHttpClient's websockets don't work in environments that require the use of proxies. To enable that, this switches the implementation for android from websocketpp to okhttp since it works well with the android OS proxy configuration and is already being used for regular http requests.

The changes are pretty big but there's 3 main components to them:

1. `okhttp_websocket_impl` implements the "perform" callbacks, maintains the java websocket's state on the native side, and schedules the needed asynchronous work
2. The `HttpClientWebSocket` java class is the glue code between the native interface and OkHttp and is mainly there to reduce the dependencies that would need to be resolved if OkHttp was used directly
3. The `HttpClientWebSocket` native class is a helper class that contains all the JNI related logic and provides an interface for `okhttp_websocket_impl` to make calls to the `HttpClientWebSocket` java class (and indirectly to OkHttp)

### DIFF
--- a/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientWebSocket.java
+++ b/Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientWebSocket.java
@@ -1,0 +1,94 @@
+package com.xbox.httpclient;
+
+import java.nio.ByteBuffer;
+
+import okhttp3.Headers;
+import okhttp3.OkHttpClient;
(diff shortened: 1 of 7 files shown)


jasonsandlin:
Thanks.  We'll review soon

SahilAshar:
Note to self: update the build.gradles and remove all OpenSSL references.

SahilAshar (approved):

PR #719, opened 2022-10-05
Fix Android WebSocket Binary Messaging

SahilAshar:


Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientWebSocket.java
```diff
@@ -73,7 +73,14 @@ public void onMessage(WebSocket webSocket, String text) {
 
     @Override
     public void onMessage(WebSocket webSocket, okio.ByteString bytes) {
-        onBinaryMessage(bytes.asByteBuffer());
+        // These needs to be a directly allocated ByteBuffer or
+        // native layer won't be able to access this.
+
+        ByteBuffer buffer = ByteBuffer.allocateDirect(bytes.size());
+        buffer.put(bytes.toByteArray());
+        buffer.position(0);
```
jgalvan, line 81: I think you can also use `buffer.rewind();`

jgalvan (approved):

Build/libHttpClient.Android/src/main/java/com/xbox/httpclient/HttpClientWebSocket.java
SahilAshar: `buffer.rewind()` resets the mark unfortunately, so don't wanna do that. https://developer.android.com/reference/java/nio/Buffer#rewind()

PR #720, opened 2022-10-11
Improving Retry-After API doc

jasonsandlin:


Include/httpClient/httpClient.h
```diff
@@ -484,7 +484,7 @@ STDAPI HCHttpCallRequestSetTimeout(
 /// <br />
 /// If the service returns an HTTP error with a "Retry-After" header, then all future calls to that API 
 /// will immediately fail with the original error without contacting the service until the "Retry-After" 
-/// time has been reached.<br />
+/// time has been reached if HCHttpCallRequestSetRetryCacheId() has been called prior to cache the Retry-After header for fast fail logic.<br />
```
ScottMunroMS, line 487: starting and ending a sentence with an if clause is sort of weird grammatically.

recommend frontloading both conditions at the start of the sentence for better readability

> if the service returns an HTTP error with a "Retry-After" header and the title had previously called HCHttpCallRequestSetRetryCacheId(), then all future calls to that API will immediately fail with the original error without contacting the service until the "Retry-After" time has been reached.<br />

Include/httpClient/httpClient.h
```diff
@@ -446,7 +446,7 @@ STDAPI HCHttpCallRequestSetRetryAllowed(
 /// ID number of this REST endpoint used to cache the Retry-After header for fast fail.
 /// </summary>
 /// <param name="call">The handle of the HTTP call.  Pass nullptr to set the default for future calls.</param>
-/// <param name="retryAfterCacheId">ID number of this REST endpoint used to cache the Retry-After header for fast fail.  1-1000 are reserved for XSAPI.</param>
+/// <param name="retryAfterCacheId">ID number of this REST endpoint used to cache the Retry-After header for fast fail. Must be non-0. 1-1000 are reserved for XSAPI.</param>
```
ScottMunroMS, line 449: nit: elsewhere these docs spell out zero as a word rather than a number. recommend "non-zero" for consistency (also I think it reads better)

ScottMunroMS (changes requested):
some small word choice feedback but otherwise LGTM

Include/httpClient/httpClient.h
jasonsandlin: thanks, fixed.  when i write my novel, i'm going to hire you as an editor

ScottMunroMS (approved):

PR #725, opened 2022-10-25
include stdarg.h in trace.h to fix ios build

natiskan:


jasonsandlin (approved):

Include/httpClient/trace.h
```diff
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <httpClient/pal.h>
+#include <stdarg.h>
```
lubeltra, line 10: it's generally better to include system headers before any "local" headers

lubeltra (approved):

PR #728, opened 2022-12-05
Fix include directories when linking to via CMake

lugehorsam:
The libhttpclient includes cannot be linked to from a CMake project because what were the `PUBLIC_SOURCE_FILES` were included via a PRIVATE scope in `target_include_directories`. This change makes them public.

### DIFF
--- a/Utilities/CMake/Android/libHttpClient/CMakeLists.txt
+++ b/Utilities/CMake/Android/libHttpClient/CMakeLists.txt
@@ -55,8 +55,6 @@ set(ANDROID_SOURCE_FILES
     "${PATH_TO_ROOT}/Source/Common"
     "${PATH_TO_ROOT}/Source/HTTP"
     "${PATH_TO_ROOT}/Source/Logger"
-    "${PATH_TO_ROOT}/Include"
-    "${PATH_TO_ROOT}/Include/httpClient"
     "${PATH_TO_ROOT}/External/asio/asio/include"
     "${PATH_TO_ROOT}/External/openssl/include"
     "${PATH_TO_ROOT}/External/websocketpp"
@@ -79,6 +77,8 @@ add_library(
 
 target_include_directories(
     "${PROJECT_NAME}"
+    PUBLIC
+    "${PUBLIC_INCLUDE_DIRS}"
     PRIVATE
     "${COMMON_INCLUDE_DIRS}"
     "${ANDROID_INCLUDE_DIRS}"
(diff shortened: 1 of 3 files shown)


SahilAshar:
Howdy! Would you mind elaborating a bit more on why this change is relevant to y'all? Is there a specific build pattern y'all are currently using that this change fits better into?

lugehorsam:
Hey @SahilAshar 👋 

We use libhttpclient inside our networking library that serves as a thin wrapper for network calls to our core service.

We use CMake as our build system and vcpkg to bring in dependencies.

If a dependency cannot be brought in via vcpkg (e.g., there is no `install` target) then we use the next best option which is to `FetchContent()` on our dependency and essentially call `add_subdirectory` on it. I think this is common in the CMake community.

The problem in this particular case is, if you were to do this off master for `libhttpclient` you would get a "header not found" error for the `httpClient/httpclient`, for the reason I mentioned in the summary.

SahilAshar:
Thanks for the info @lugehorsam!

For some context, we currently recommend that Android LibHC consumers integrate LibHC one of two ways:
1. Take the LibHC Android Workspace, generate the relevant `libHttpClient.a` binary and `.aar` files, and link those into your CMake and Gradle directly. 
2. Add LibHC repo as a submodule to your repo, modify the `settings.gradle` for your Android project to add the `libHttpClient/Build/libHttpClient.Android` dir as a local module project, and integrate that local module as a dependency in your Android project's `build.gradle` file. Additionally, making sure you're linking the appropriate libs in your CMake, and you're including the header path in your CMake.

Making LibHC's CMake installable is not something we've considered in past; we've preferred customers take one of those build paths instead. That being said, I think your suggested changes make sense for this use case, and will help with other future customers following a similar pattern. I'm hesitant to take this change immediately without doing some further digging into other customer use cases and making sure this doesn't break anything else. I'll keep you updated, and please let me know if you have any questions in the meantime.

lugehorsam:
Okay thanks @SahilAshar, I'd definitely consider it as CMake will give you an abstraction that allows the same integration steps to be taken regardless of the host and target platform, so you won't need to maintain separate build instructions anymore.

jasonsandlin:
closing this PR as we've since we've since added Linux cmake build files.  

more info found here:
https://github.com/microsoft/libHttpClient/blob/main/Build/libHttpClient.Linux/README.md

when you get a chance can you see if this fixes the issue you were hitting

PR #729, opened 2022-12-13
Fix CMake builds for Apple and Linux

lugehorsam:
Based on #728 

- Namespaces the ios wait timer to `OS` like the other wait timers.
- Removes header prefixing for dynamic library Apple builds. We actually build libhttpclient as a CMake "object library" -- if this change breaks dynamic builds I can revert this change and find a better preprocessor directive.
- Wraps some MSVC-specific flags in a conditional for that compiler.
- Removes the force setting of debug builds for multi-config CMake generators like Xcode.
- Fixes source file paths.
- Adds some compiler flags into CMake that exist in the project files already.

I know CMake isn't the primary way you guys build the library but since the CMake file exists, we figured it may as well build. Happy to further discuss our use case and why CMake support is necessary for us.


jasonsandlin:
thanks for the PR however we've since added official support for these platforms, so i'm going to close this PR.  if you need to build Apple with CMake still (or anything else was missed), feel free to raise issues or send another PR as usual

PR #732, opened 2023-03-08
LibHttpClient Linux: Add Linux build files and configure files to support LibHC Linux.

SebastianPD:
This update is to add LibHC support to Linux. The following code consists of build files, modifications to files, and a new submodule.

This has been tested with WSL and built using Visual Studio 2019.


SahilAshar:
@lubeltra @jplafonta I've already reviewed this PR privately and it LGTM. Adding y'all in case you want to quickly peruse.

ScottMunroMS:
> @lubeltra @jplafonta I've already reviewed this PR privately and it LGTM. Adding y'all in case you want to quickly peruse.

@SahilAshar @SebastianPD 
It's my understanding that deb's team doesn't own LHC (even if y'all have contributed to it) so I'd like us to hold off on checking in until we get some kind of signoff from their team (even if it's purely a ceremonial rubber stamp)

SahilAshar:
> > @lubeltra @jplafonta I've already reviewed this PR privately and it LGTM. Adding y'all in case you want to quickly peruse.
> 
> @SahilAshar @SebastianPD It's my understanding that deb's team doesn't own LHC (even if y'all have contributed to it) so I'd like us to hold off on checking in until we get some kind of signoff from their team (even if it's purely a ceremonial rubber stamp)

Agreed - I told Seb we'd hold off merging until we got signoffs from other primary contributors.

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
```diff
@@ -49,6 +49,10 @@
 #include <wincrypt.h>
 #endif
 
+#if HC_PLATFORM == HC_PLATFORM_LINUX
```
lubeltra, line 52: We should avoid platform specific code. Does including this break other platforms?

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
```diff
@@ -94,8 +98,49 @@ bool verify_cert_chain_platform_specific(asio::ssl::verify_context &verifyCtx, c
...
+    int ret = X509_STORE_set_default_paths(store);
+    if (ret != 1)
+    {
+        return verify_result;
+    }
+
+    ret = X509_STORE_load_locations(store, NULL, "/etc/ssl/certs");
+    if (ret != 1)
+    {
+        printf("\n FAILED TO LOAD CERTS \n");
```
lubeltra, line 122: no printf
if you need to log something use the tracing macros

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
```diff
@@ -94,8 +98,49 @@ bool verify_cert_chain_platform_specific(asio::ssl::verify_context &verifyCtx, c
         certChain.push_back(std::move(certData));
     }
 
-    auto verify_result = verify_X509_cert_chain(certChain, hostName);
+    auto verify_result = false;
+
+#if HC_PLATFORM == HC_PLATFORM_LINUX
+    X509_STORE *store = X509_STORE_new();
```
lubeltra, line 104: why isn't this another implementation of `verify_X509_cert_chain`?

Source/HTTP/Curl/CurlEasyRequest.h
```diff
@@ -7,7 +7,11 @@
 #else
 // This path is untested, but this http provider should work with other curl implementations as well.
 // The logic in CurlMulti::Perform is optimized for XCurl, but should work on any curl implementation.
-#include <curl.h>
+#include <curl/curl.h>
+//The extern structs are here for LibHC Linux. CurlEasyRequest, CurlProvider, and CurlMulti create a circular dependency.
```
lubeltra, line 11: 1) can you expand on what the issue is?
2) can these declarations be moved into cpp files?
3) you should use HC_DECLARE_TRACE_AREA and not name the globals directly

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -38,17 +39,14 @@ Result<HC_UNIQUE_PTR<CurlEasyRequest>> CurlEasyRequest::Initialize(HCCallHandle
...
-        // we set both POSTFIELDSIZE and INFILESIZE because curl uses one or the
-        // other depending on method
-        RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_POSTFIELDSIZE, static_cast<long>(bodySize)));
-        RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_INFILESIZE, static_cast<long>(bodySize)));
-
-        // read callback
-        RETURN_IF_FAILED(easyRequest->SetOpt<curl_read_callback>(CURLOPT_READFUNCTION, &ReadCallback));
-        RETURN_IF_FAILED(easyRequest->SetOpt<void*>(CURLOPT_READDATA, easyRequest.get()));
-    }
+    // we set both POSTFIELDSIZE and INFILESIZE because curl uses one or the
```
lubeltra, line 42: why?

Include/httpClient/config.h
```diff
@@ -57,6 +58,14 @@
     #else
         #define HC_DATAMODEL HC_DATAMODEL_ILP32
     #endif
+#elif defined(__linux__) && !defined(__ANDROID__)
```
lubeltra, line 61: I would rewrite this as
```c++
#elif defined(__linux__)
    #if defined(__ANDROID)
        #define HC_PLATFORM HC_PLATFORM_ANDROID
    #else
        #define HC_PLATFORM HC_PLATFORM_LINUX
    #endif

    #if defined(__LP64__)
        #define HC_DATAMODEL HC_DATAMODEL_LP64
    #else
        #define HC_DATAMODEL HC_DATAMODEL_ILP32
    #endif
#endif
```

Build/libHttpClient.Linux.C/libHttpClient.Linux.C.vcxproj
```diff
@@ -0,0 +1,187 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
lubeltra, line 1: is this generated from a template?

lubeltra (changes requested):

Build/libHttpClient.Linux.C/libHttpClient.Linux.C.vcxproj
SebastianPD: It was generated from creating empty projects in visual studio.

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
SebastianPD: I placed it there due to already having the cert stack vat to be verified with OpenSSL. My previous implementation had me convert it all back to a cert stack in verify_X509_cert_chain but to prevent errors and to not be redundant I leveraged what was already made. From my understanding of the code what's supposed to happen is that openssl should verify before ever reaching here to this file, but it does not, so we step in here at the very end.

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
SebastianPD: I can only test it against windows at the moment, and that works.

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
SebastianPD: this was for debugging. I was supposed to remove this.

Source/HTTP/Curl/CurlEasyRequest.h
SebastianPD: 1. John explained to me what the problem was a while back and it had to do with how the compilers handled the declaration of traces in the .h file. The Clang Compiler doesn't know what should be filled in for g_trace while MSBuild did. 
2. Tried putting them in the cpp files and that did not work.
3. I should do that.

Source/HTTP/Curl/CurlEasyRequest.cpp
SebastianPD: It was to resolve a bug I found with the curl library. This doesn't happen in xcurl. What happened was that when creating a WebSocket it would never complete and that was because MLP wanted a websocket it a fixed body. When sending the request for the handshake it would send no data and these setOpts would never be initialized. Removing the if statement solves the issue. Bodysize is unsigned so I we don't have to worry about a negative number.

ScottMunroMS: @SebastianPD do we know that this is actually a bug in curl? or could this just be curl's (perhaps poorly documented) intentional behavior? 

in general, the goal of xcurl is to mirror curl's behavior as closely as reasonably possible. If this isn't definitely a bug it's possible there's reasons xcurl should mirror this as well

> beacuse MLP wanted a websocket it a fixed body
can you elaborate on this? if the bodySize here is zero, what additional data is it now sending?

additionally, have we validated that the requests sent by xcurl haven't changed at all by removing this if-check? even if things still work it'd be interesting to know whether the requests are sent with any additional headers which might not have been there beforeor if the CURLOPT_READFUNCTION is now called in a code path in which it previously wasn't

Include/httpClient/config.h
```diff
@@ -57,8 +58,12 @@
     #else
         #define HC_DATAMODEL HC_DATAMODEL_ILP32
     #endif
-#elif defined(__ANDROID__)
-    #define HC_PLATFORM HC_PLATFORM_ANDROID
+#elif defined(__linux__)
+    #if defined(__ANDROID)
```
ScottMunroMS, line 62: shouldn't this be `__ANDROID__`?

SebastianPD: you're right I'll resolve this in my next push. Need to make the templates

Source/HTTP/Curl/CurlEasyRequest.cpp
ScottMunroMS: > because MLP wanted a webscoket it a fixed body

I also wonder whether this is a bug in MLP instead of libhttpclient. What requires the fixed body and could be address it at the MLP level?

(if discussing this in any meaningful way requires sharing internal details about MLP let's chat out-of-band, maybe on teams. we should avoid putting MLP internal code in this public repo even in comments)

SebastianPD: what would happen previously when doing the handshake was `Transfer-Encoding: chunked` was a header and removing this if statement and setting the set opt would make the header say it has a size of 0. The body would have nothing in it. Those were the only differences when looking at the fiddler trace.

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
lubeltra: we could pass the cert stack down into `verify_X509_cert_chain`
I really want to minimize the amount of code we ifdef for specific platforms as it really hurts readability

lubeltra (changes requested):

Source/WebSocket/Websocketpp/x509_cert_utilities.hpp
```diff
@@ -560,7 +526,61 @@ bool verify_X509_cert_chain(const http_internal_vector<http_internal_string> &ce
 }
 #endif
 
+#if defined(__linux__)
```
lubeltra, line 529: check on HC_PLATFORM

lubeltra (changes requested):

jasonsandlin (approved):
:shipit:

PR #733, opened 2023-03-16
Fix for XcURL and cURL behavior differences.

SebastianPD:
When setting up POSTEILDSZE and INFILESIZE cURL and XcURL behave differently when the fields are set to 0. This fix is to account for that.


Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -41,14 +41,27 @@ Result<HC_UNIQUE_PTR<CurlEasyRequest>> CurlEasyRequest::Initialize(HCCallHandle
...
+    #else
+    if (bodySize > 0) {
+        RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_POSTFIELDSIZE, static_cast<long>(bodySize)));
+        RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_INFILESIZE, static_cast<long>(bodySize)));
+
+        // read callback
+        RETURN_IF_FAILED(easyRequest->SetOpt<curl_read_callback>(CURLOPT_READFUNCTION, &ReadCallback));
+        RETURN_IF_FAILED(easyRequest->SetOpt<void*>(CURLOPT_READDATA, easyRequest.get()));
+    }
+    #endif
```
jplafonta, line 65: could you instead do:

// XCurl code
#else
// curl code
#endif

since HC_PLATFORM_GDK is where XCurl is used.  This is a work around for XCurl, not specific to linux

jplafonta (approved):
:shipit:

PR #735, opened 2023-04-11
Fix crash when last websocket handle is closed while sending a message

lubeltra:
`HCWebSocketSendMessageAsync` and similar were not ensuring that the `HC_WEBSOCKET_OBSERVER` was kept alive until they completed, but store a pointer to it for use in the completion callbacks.

This change switches to properly using reference counting to manage the observer lifetime


jplafonta (approved):
:shipit:

jasonsandlin (approved):

PR #737, opened 2023-04-17
User/sebastianpd/libHC linux CMake

SebastianPD:
Created CMake files for LibHC Linux and a bash script to compile them so users running native Linux can develop and build from their machines instead of using a windows PC to build from the VCXProj files.

The Bash script can compile either a Debug or Release mode of libHC Linux and when the script is done all library files will be located in the Binaries folder.


Utilities/LinuxBuildScripts/CMake.bash
SahilAshar, line 1: this file is pretty messy - would def like to see more usage of variables to hold paths and less reliance on relative paths like this.

Utilities/CMake/Linux/README.md
```diff
@@ -0,0 +1,4 @@
+# libHttpClient Linux CMake build system
```
lubeltra, line 1: Before I look too deep into the CMake files (which I'm by no means an expert on)
Doesn't libhc already have a cmake file? Why aren't we using it?

Utilities/LinuxBuildScripts/CMake.bash
SebastianPD: I realized that when looking over my script. Making it more like my curl script where I establish a variable that has the full path. and making a less clunky if statement.

Utilities/CMake/Linux/README.md
SebastianPD: Not really. There are CMake files for android but for the other platforms it does not make cmake files but instead VCXProj files. the CMake.cmd script will go and copy the needed dependencies for each project and generate a list of files then it would paste them into a vcxproj template.

Utilities/CMake/Linux/libHttpClient/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+    "${PATH_TO_ROOT}/External/asio/asio/include"
+    "${PATH_TO_ROOT}/External/openssl/include"
+    "${PATH_TO_ROOT}/External/curl/include"
+    "${PATH_TO_ROOT}/External/websocketpp"
+    )
+
+set(LINUX_INCLUDE_DIRS
+    "${PATH_TO_ROOT}/External/opensslGeneratedHeaders/linux"
+    "${PATH_TO_ROOT}../openssl/install/linux/openssl/include"
+    "${PATH_TO_ROOT}../openssl/generatedHeaders/linux/include"
```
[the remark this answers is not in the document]
SebastianPD: One is the openssl headers in libHC external, the other one is the openssl headers in MLP external. I remember needing to use them when I was porting but I will check and test again to make sure if what I am doing is correct or even necessary.

Utilities/CMake/Linux/libHttpClient/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+    "${PATH_TO_ROOT}/Include/httpClient"
+    "${PATH_TO_ROOT}/External/asio/asio/include"
+    "${PATH_TO_ROOT}/External/openssl/include"
+    "${PATH_TO_ROOT}/External/curl/include"
+    "${PATH_TO_ROOT}/External/websocketpp"
+    )
+
+set(LINUX_INCLUDE_DIRS
+    "${PATH_TO_ROOT}/External/opensslGeneratedHeaders/linux"
+    "${PATH_TO_ROOT}../openssl/install/linux/openssl/include"
```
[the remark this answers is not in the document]
SebastianPD: No "${PATH_TO_ROOT}/External/opensslGeneratedHeaders/linux" are the openssl headers located in external.

Utilities/CMake/Linux/README.md
lubeltra: We still have a CMakeList.txt file in Utilities/CMake/ is there a way we can use it as a source for the linux build?
The current setup is already painful enough to use without adding another project file nobody will remember to ever update or test.

Utilities/CMake/Linux/libHttpClient/CMakeLists.txt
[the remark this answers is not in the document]
SebastianPD: after testing I realized they are not needed. Removing all but "${PATH_TO_ROOT}/External/opensslGeneratedHeaders/linux"

Utilities/CMake/Linux/README.md
SahilAshar: @lubeltra I'm fairly certain that CMake exists for VS CMake integration, looking over it, integrating Linux into it may be more trouble than it's worth. I do get your point though - I do think there might be value in combining the Linux and Android CMakes into an integrated one. Might be out of scope for this PR though.

PR #738, opened 2023-04-27
User/sebastianpd/lib hc linux c make

SebastianPD:
Created CMake files for LibHC Linux and a bash script to compile them so users running native Linux can develop and build from their machines instead of using a windows PC to build from the VCXProj files.

The Bash script can compile either a Debug or Release mode of libHC Linux and when the script is done all library files will be located in the Binaries folder.

Removed WSL VCXProj files

Created new pipeline for libHC Linux. The first task is to install ninja and generate files with CMake. The second task is to build all the required libraries.


SahilAshar (approved):
:shipit:

PR #744, opened 2023-05-31
Added changes to build for PS

nassosterz-ms:
Adding a macro for Playstation 5, as well, as fixing, what I think is an issue when calling localtime_s.


Source/Logger/trace.cpp
```diff
@@ -73,8 +73,8 @@ void TraceMessageToDebugger(
     uint32_t     fractionMSec = static_cast<uint32_t>(timestamp % 1000);
     std::tm      fmtTime = {};
 
-#if HC_PLATFORM_IS_MICROSOFT
-    localtime_s(&fmtTime, &timeTInSec);
+#if HC_PLATFORM_IS_MICROSOFT || HC_PLATFORM_SONY_PLAYSTATION_4 || HC_PLATFORM_SONY_PLAYSTATION_5
+    localtime_s(&timeTInSec, &fmtTime);
```
jplafonta, line 77: >localtime_s(&timeTInSec, &fmtTime); [](http://example.com/codeflow?start=4&length=35)

doesn't this change the HC_PLATFORM_IS_MICROSOFT path? fmtTime and timeTInSec are flipped

nassosterz-ms: Per here: https://en.cppreference.com/w/c/chrono/localtime I feel that they should be flipped for HC_MICROSOFT. Alternatively, I can add an elif case for Playstation

jplafonta: See here: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-s-localtime32-s-localtime64-s?view=msvc-170

jplafonta (approved):

PR #746, opened 2023-06-01
Internal PAL improvements (Part 2)

jplafonta:
* Additional cleanup related to shift to new internal PAL
  * Renamed and cleaned up HC_PERFORM_ENV. It is no longer used as a global context object passed to provider functions and we no longer need to expose it publicly.
  * Restructured how HC_CALL and WebSocket objects are created so that they don't depend on accessing global state
* Fixes to Curl provider to support implementation on new platforms.
  * Added some ifdef to add support for different versions of libcurl
  * Made some private members of CurlProvider protected to allow inheriting from and extending that class with platform specific functionality
* Small fix to allow for HCInitArgs to be defined outside httpClient.h, allowing for platform specific definitions.
* General cleanup including addition of StdExtra::optional, small improvements to Result class, and the additional of the types.h header


tculotta (approved):
:shipit:

PR #747, opened 2023-06-02
Fix leak in WebSocket

jplafonta:


lubeltra (commented):
Looks good, will give final approval Monday when I validate this does not reintroduce the crash

lubeltra (approved):

PR #749, opened 2023-06-05
Build fixes for Switch

jplafonta:
* Changing some preprocessor guards to support building in a Win32 environment that is a non-Windows HC_PLATFORM. This will be used exclusively for testing purposes


jasonsandlin (approved):
:shipit:

PR #752, opened 2023-06-13
LibHC linux build curl using libssl in repo submodule

SebastianPD:
Have cURL build using the libhc submdule in repo instead of relying on libssl install in machine. This will guarantee that libHC will build properly and that there will be no undefined symbols when building projects. 

For context Ubuntu 22 uses Openssl 3 while Ubuntu 20 uses Openssl 1.1.1. Our submodule uses 1.1.1 so in our previous iteration of libHC we would have a successful build on Ubuntu 20 but ubdefined symbols on Ubuntu 22.


Utilities/LinuxBuildScripts/README.md
```diff
@@ -28,6 +28,12 @@ Running the build script with the `-c|--config` argument will generate  Debug or
 
 Running the build script with the `-nc|--nocurl` will **not** generate a binary of `libcurl.a`. Use this flag if you wish to bring your own version of cURL.
 
+```
+./libHttpClient_Linux.bash <-ns|--nossl>
+```
+
+Running the build script with the `-ns|--nossl` will **not** generate a binary of `libssl.a` and `libcrypto.a`. Use this flag if you wish to bring your own version of cURL.
```
SahilAshar, line 35: > your own version of cURL.

you own version of OpenSSL.

SahilAshar (approved):

PR #755, opened 2023-06-20
Update Android against new PAL

jplafonta:
* Added AndroidWebSocketProvider and AndroidHttpProvider replacing existing callbacks
* Implement PlatformInitialize for Android


Source/HTTP/Android/AndroidHttpProvider.cpp
```diff
@@ -0,0 +1,85 @@
...
+            m_platformComponents->GetHttpResponseClass()
+        )
+    };
+
+    HRESULT result = httpRequest->Initialize();
+
+    if (!SUCCEEDED(result))
+    {
+        HCHttpCallResponseSetNetworkErrorCode(call, result, 0);
+        XAsyncComplete(asyncBlock, result, 0);
```
jasonsandlin, line 35: better to do:
        HCHttpCallResponseSetNetworkErrorCode(call, result, result);

cause 0 = S_OK

jplafonta: changed

SahilAshar (approved):
:shipit:

PR #760, opened 2023-06-28
Update linux against new internal PAL

jplafonta:
* Add WebSocketppProvider and PlatformIntialize for Linux
* Update Linux build files
* Cleanup CMakeLists since its no longer used for Windows builds


SebastianPD:
Built and tested this version of libHC using Ubuntu 22.04. All Tests Passed. 
Forcing c++ 17 is not problem since both Ubuntu 20 and Ubuntu 22 have it.

SebastianPD (approved):

jasonsandlin (approved):
:shipit:

PR #764, opened 2023-07-12
Naster/ps trace definitions

nassosterz-ms:
* Adding HC_PLATFORM_IS_PLAYSTATION
* Moving definition of trace areas from pch.cpp to global.cpp as it being in pch.cpp causes linker errors!


jasonsandlin (approved):

jplafonta (approved):
:shipit:

PR #772, opened 2023-08-03
Task 45828333: Move existing ETW tracing into LHC

tculotta:


Source/Logger/trace.cpp
```diff
@@ -46,6 +50,35 @@ int vsprintf_s(char(&buffer)[SIZE], _Printf_format_string_ char const* format, v
 #endif
 #endif
 
+#if HC_PLATFORM_IS_MICROSOFT
+TRACELOGGING_DEFINE_PROVIDER(
+    g_hTraceLoggingProvider,
+    "libHttpClient",
+    (0x2b6274b2, 0x159f, 0x4e6f, 0xb8, 0x0, 0xd, 0x17, 0x0, 0x9e, 0x14, 0x7a) // {2B6274B2-159F-4E6F-B800-0D17009E147A}
```
tculotta, line 57: This is the same existing GUID PF currently uses since that should already be registered in the wherever-that-is. Unless the name there needs to match what's defined here, I assume nothing needs to be changed service-side.

jplafonta: That work hasn't been done yet anyhow AFAIK so I don't think that's a concern

jplafonta:
>     TraceMessageToClient(area->Name, level, threadId, timestamp, message);

I'd probably add a new method (i.e. TraceMessageETW) rather than include this in TraceMessageToClient

---
Refers to: Source/Logger/trace.cpp:304 in 121f8d0. [](commit_id = 121f8d000b5db760ccedc1a117358213b1709bf8, deletion_comment = False)

Source/Logger/trace.cpp
```diff
@@ -221,6 +319,8 @@ void TraceState::Init() noexcept
     if (previousCount == 0)
     {
         m_initTime = std::chrono::high_resolution_clock::now();
+
+        RegisterProvider();
```
jplafonta, line 323: >RegisterProvider(); [](http://example.com/codeflow?start=8&length=19)

Curious about the best time to call this.  I think ideally we wouldn't register the provider if the client hasn't registered ETW.  Maybe move to HCTraceSetEtwEnabled?

tculotta: Ah that's a better idea, yeah

jplafonta (approved):
:shipit:

PR #774, opened 2023-08-03
Task 45629821: Add support for GZIP Compression

rgomez391:
Enabling GZIP compression for LibHttpClient using ZLib for Win32 and GDK Platforms initially. This work will expand to other platforms down the road.


Build/libHttpClient.Zlib/libHttpClient.Zlib.vcxitems
rgomez391, line 1: In this file I had to disable some specific warnings related to ZLib source code, otherwise library will not compile since those warnings are treated as errors.

Any issue on this?

Source/HTTP/httpcall.cpp
```diff
@@ -76,6 +77,40 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
     {
     case XAsyncOp::Begin:
     {
+#if !HC_NOZLIB
+
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK
+        // Compress body before call if applicable
+        if (call->compressionAlgorithm != xbox::httpclient::HCCompressionAlgorithm::None)
```
rgomez391, line 84: @jplafonta would you consider this is the right place to do body compression?

Build/libHttpClient.Zlib/libHttpClient.Zlib.vcxitems
jplafonta: No that's fine, we've done similar in the past when external code has warnings

libHttpClient.props
```diff
@@ -46,6 +46,7 @@
     <HCRoot Condition="'$(HCRoot)'==''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.root))\</HCRoot>
     <HCBuildRoot>$(HCRoot)\Build</HCBuildRoot>
     <libHttpClientInclude>$(HCRoot)\Include\</libHttpClientInclude>
+    <zlibInclude>$(HCRoot)\External\zlib\</zlibInclude>
```
jplafonta, line 49: This will add Zlib headers to client's include path. I think that's not what you want.  This probably belongs in libHttpClient.Zlib.vcxitems

Build/libHttpClient.GDK.props
```diff
@@ -96,6 +96,15 @@
...
   </ItemDefinitionGroup>
 
+  <PropertyGroup>
+    <HCNoZlib>false</HCNoZlib>
+  </PropertyGroup>
+  <ItemDefinitionGroup Condition="'$(HCNoZlib)'=='true'">
+    <ClCompile>
+      <PreprocessorDefinitions>%(PreprocessorDefinitions);HC_NOZLIB</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
```
jplafonta, line 106: This also feels like the wrong spot for this.  This props files defines the build config for all LHC projects, not just libHttpClient.lib (and is actually used by PlayFab and XSAPI as well).  This also probably belongs in libHttpClient.Zlib.vcxitems.

Source/HTTP/httpcall.h
```diff
@@ -18,6 +18,18 @@ struct HeaderCompare
...
 
+#if !HC_NOZLIB
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK
+enum class HCCompressionAlgorithm : uint32_t
+{
+    None = 0,
+    Gzip,
+    Deflate,
+    Brotli
+};
```
jplafonta, line 29: Seems like only GZip and None are used here.  Any reason why you didn't just do something like 'bool m_enableGzipCompression'?

jplafonta:
>     HCHttpCallRequestBodyReadFunction requestBodyReadFunction{ HC_CALL::ReadRequestBody };

Is it supported to set a RequestBodyReadFunction and have gzip compression enabled?

---
Refers to: Source/HTTP/httpcall.h:60 in 36e12bc. [](commit_id = 36e12bc0089a7347a9a7eb60f0c14d3858ecf9d2, deletion_comment = False)

Source/HTTP/httpcall.cpp
jplafonta: Yes and no. I think it should be done as the first step of PerformAsync, but if you do it as part of XAsyncBegin it will run synchronously, which is likely bad.  Probably what you will need to do is queue an async task to do the compression with XTaskQueueSubmitDelayedCallback.  When that step is complete, it will continue with the first perform iteration

jplafonta: The compression should also probably happen after the Fast fail check

Source/HTTP/httpcall.h
jasonsandlin: Internally it doesn't much matter what we do here.  I wouldn't add options that aren't implemented however, so None and GZip are fine.   On the public surface area, in httpClient.h I would do something like:

STDAPI HCHttpCallRequestEnableCompression(
    _In_ HCCallHandle call,
    HCCompressionType type
) noexcept;
#endif

where HCCompressionType is None, Low, Medium, High.

Build/libHttpClient.GDK.props
rgomez391: I think these couldn't live in libHttpClient.Zlib.vcxitems because this property group and item definition group are the ones that define if the libHttpClient.Zlib.vcxitems file will be included in the solution. HC_NOZLIB should be defined somewhere outside libHttpClient.Zlib.vcxitems for the condition below:

https://github.com/microsoft/libHttpClient/blob/36e12bc0089a7347a9a7eb60f0c14d3858ecf9d2/Build/libHttpClient.143.Win32.C/libHttpClient.143.Win32.C.vcxproj#L13C3-L13C136

libHttpClient.props
rgomez391: Fixed, I moved it to libHttpClient.Zlib.vcxitems

Source/HTTP/httpcall.h
rgomez391: Fixed

Source/HTTP/httpcall.cpp
rgomez391: I just implemented these changes; can you please take a quick look at it?

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "files.associations": {
+        "xiosbase": "cpp"
+    }
+}
```
jplafonta, line 5: I think this shouldn't be checked in

Source/HTTP/httpcall.cpp
```diff
@@ -137,6 +138,91 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
...
+    if (call->traceCall)
+    {
+        HC_TRACE_INFORMATION(HTTPCLIENT, "HC_CALL::CompressRequestBody [ID %llu] Iteration %d", TO_ULL(call->id), call->m_iterationNumber);
+    }
+
+#if !HC_NOZLIB
+
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK
+    // Compress body before call if applicable
+    if (call->enableGzipCompression && call->compressionLevel != HCCompressionLevel::None)
```
jplafonta, line 161: >call->enableGzipCompression && call->compressionLevel != HCCompressionLevel::None [](http://example.com/codeflow?start=8&length=81)

Minor, but I would move this check into the provider above so that CompressRequestBody doesn't even get scheduled if compression is disabled

Source/HTTP/httpcall.cpp
```diff
@@ -137,6 +138,91 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
...
+            return;
+        }
+        
+        size_t uncompressedRequestyBodySize = 0;
+        const size_t bufferSize = 128 * 1024;
+        uint8_t uncompressedRequestyBodyBuffer[bufferSize];
+
+        try
+        {
+            hr = clientRequestBodyReadCallback(call, 0, bufferSize, clientRequestBodyReadCallbackContext, uncompressedRequestyBodyBuffer, &uncompressedRequestyBodySize);
```
jplafonta, line 180: >hr = clientRequestBodyReadCallback(call, 0, bufferSize, clientRequestBodyReadCallbackContext, uncompressedRequestyBodyBuffer, &uncompressedRequestyBodySize); [](http://example.com/codeflow?start=12&length=157)

What happens if the buffer isn't large enough here? I believe you'll need some logic to resize the buffer and call the client callback again with an offset

Source/HTTP/httpcall.cpp
```diff
@@ -137,6 +138,91 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
...
+        auto requestCompressedBodyBytes = compressedRequestBodyBuffer.data();
+        auto requestCompressedBodySize = (uint32_t)compressedRequestBodyBuffer.size();
+
+        // Setting back to default read request body callback to be invoked by the Platform-specific code
+        call->requestBodyReadFunction = HC_CALL::ReadRequestBody;
+        call->requestBodyReadFunctionContext = nullptr;
+
+        // Directly setting compressed body bytes to HCCall
+        call->requestBodySize = requestCompressedBodySize;
+        call->requestBodyBytes.assign(requestCompressedBodyBytes, requestCompressedBodyBytes + requestCompressedBodySize);
```
jplafonta, line 206: >call->requestBodyBytes.assign(requestCompressedBodyBytes, requestCompressedBodyBytes + requestCompressedBodySize); [](http://example.com/codeflow?start=8&length=114)

I believe this copies the request body buffer. You should be able to avoid that with std::move(compressedRequestBodyBuffer)

Source/HTTP/httpcall.cpp
rgomez391: I had a question about what the purpose of the offset was and this is actually answers that... so maybe adding like a loop where we check that we already got all the request bytes?

.vscode/settings.json
rgomez391: Fixed

Source/HTTP/httpcall.cpp
rgomez391: Fixed

Source/HTTP/httpcall.cpp
rgomez391: Fixed

Source/HTTP/httpcall.cpp
```diff
@@ -137,6 +162,105 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
...
+
+    if (FAILED(hr) || !clientRequestBodyReadCallback)
+    {
+        HC_TRACE_ERROR(HTTPCLIENT, "HC_CALL::CompressRequestBody: Unable to get client's RequestBodyRead callback");
+        return;
+    }
+
+    size_t uncompressedRequestyBodySize = 0;
+    const size_t bufferSize = 500;
+    uint8_t* uncompressedRequestyBodyBuffer = new uint8_t[requestBodySize];
```
rgomez391, line 198: @jplafonta is there a better way to define the buffer we'll be reading from the client? 

Since this is a dynamically defined value, not sure how we can achieve this on the stack, so I've allocated it on the heap and deleting it after compression function.

Source/HTTP/httpcall.cpp
rgomez391: Fixed, can you take a look at my solution?

Source/HTTP/httpcall.cpp
```diff
@@ -100,8 +101,32 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
...
+                RETURN_IF_FAILED(XTaskQueueSubmitDelayedCallback(context->workQueue, XTaskQueuePort::Work, performDelay, context, HC_CALL::PerformSingleRequest));
+
+#if !HC_NOZLIB
+
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK
+
+            }
+#endif
+
+#endif // !HC_NOZLIB
```
jplafonta, line 128: nit: for readability, could you break this up into a single if HC_NOZLIB preprocessor block? You'll need to duplicate one line, but I think it will be much more readable

Build/libHttpClient.GDK.props
jplafonta: Understood. Since there's actually separate things being defined here, I think we should split this discussion up for clarity.

The 'HCNoZlib' build property shouldn't be defined here.  I don't think you'll ever need to set that.  That should only be set by clients who are building LHC by source. LHC projects should look for that property being set to 'true' and react accordingly.

I'd like to restrict defining preprocessor macro HC_NOZLIB to the LHC lib projects (I believe thats the only place it's needed?) Could that be moved to libHttpClient.Common.vcxitems?

Source/Common/compression.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+}
+
+void Compression::CompressToGzip(uint8_t* inData, uInt inDataSize, HCCompressionLevel compressionLevel, http_internal_vector<uint8_t>& outData)
+{
+    http_internal_vector<uint8_t> buffer;
+    uint32_t compressionLevelValue = static_cast<std::underlying_type<HCCompressionLevel>::type>(compressionLevel);
+
+    const size_t BUFSIZE = 128 * 1024;
+    uint8_t temp_buffer[BUFSIZE];
```
jplafonta, line 26: It still feels like you might be doing more copying than necessary here.  Could you initialize 'buffer' to be BUFSIZE, pass a pointer directly to that and the resize as needed for subsequent calls to deflate?

Source/HTTP/httpcall.cpp
```diff
@@ -137,6 +162,105 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
     }
 }
 
+#if !HC_NOZLIB
```
jplafonta, line 165: nit: minor style preference - no newline between these preprocessor guards

Source/HTTP/httpcall.cpp
jplafonta: You can't avoid using a dynamic allocation since we don't know the body size ahead of time. I'd use a Vector for a few reasons:

1) Its easier to resize on the fly
2) It automatically uses the client's allocators - you're using 'new' here which won't
3) you don't have to worry about freeing the allocation manually

Source/HTTP/httpcall.cpp
jplafonta: similar to my comment on the 'compress' method - I don't think you really need a second temp buffer here.  Use a vector + vector::resize as needed.  The reallocation and copying will be handled automatically that way

Source/Common/compression.cpp
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation
```
jplafonta, line 1: Is this used for anything besides compressing HTTP request bodies? would it make sense to move if from Source/Common to Source/HTTP?

Build/libHttpClient.GDK.props
rgomez391: Fixed, I removed setting the MSBuild property and moved setting the preprocessor definition to libHttpClient.Common.vcxitems file

Source/Common/compression.cpp
rgomez391: We can totally move it if it makes more sense! Working on it

Source/Common/compression.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include "utils.h"
+#include <zlib.h>
+
+NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
+
+class Compression
+{
+public:
+    static void Compression::CompressToGzip(uint8_t* inData, uInt inDataSize, HCCompressionLevel compressionLevel, http_internal_vector<uint8_t>& outData);
```
jplafonta, line 15: >Compression:: [](http://example.com/codeflow?start=16&length=13)

intentional?

rgomez391: oh no... that was a copy paste mistake from the source file! Let me correct it

rgomez391: Fixed

Source/Common/compression.cpp
rgomez391: Already moved both files.

Source/HTTP/httpcall.cpp
rgomez391: Fixed

Source/HTTP/httpcall.cpp
rgomez391: Fixed

jplafonta: I totally missed that we could query the full request body size up front with HCHttpCallRequestGetRequestBodyReadFunction.  In that case you can just create a vector large enough to hold the whole thing and avoid this loop altogether.  Sorry for making this more complicated than it needed to be!

rgomez391: Fixed 🙂

jplafonta (approved):
:shipit:

PR #775, opened 2023-08-09
Fixes to the GDK props file

jplafonta:
* Import GDK bwoi props file prior to using VCTargetsPath as that gets changed when 'GDKUseBWOI' is set
* Define _DEBUG macro for debug builds to avoid linker errors when building XSAPI


tculotta (approved):
:shipit:

AlonsoMondal (approved):

PR #776, opened 2023-08-10
GDK build fix

jplafonta:
* Allow override of SpectreMitigation property since enabling it breaks XSAPI's GDK BWOI builds


tculotta (approved):
:shipit:

PR #778, opened 2023-09-27
Expanding compression to Switch platform

rgomez391:


jasonsandlin (approved):

Source/HTTP/compression.cpp
```diff
@@ -3,7 +3,7 @@
 #include "pch.h"
 
 #if !HC_NOZLIB
-#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK || HC_PLATFORM == HC_PLATFORM_NINTENDO_SWITCH
```
jplafonta, line 6: Will this break Switch builds until zlib build material has been added to the LHC.Switch repo?

rgomez391: It should not break the Switch build; this will only expose the Compression Functionality to Switch, the actual calls to all of this are in the PlayFab.C PR. We should be fine for now.

PR #783, opened 2023-10-10
libcurl build update

SebastianPD:
When running the pipeline libcurl doesn't actually build. Modifying this script will allow up to build libcurl since the script can now find the appropriate LDFLAG


jasonsandlin:
file path moved so you'll have to merge with latest

jasonsandlin (approved):

SahilAshar (approved):

PR #785, opened 2023-10-25
Adding Compresion API to exp files

rgomez391:


jasonsandlin (approved):

rgomez391:
Abandoning PR for the time being since compression support for iOS hasn't been added and header is not even being compiled in ARM64 architecture.

PR #786, opened 2023-11-09
GZIP Compression support for Apple

rgomez391:


Include/httpClient/httpClient.h
```diff
@@ -381,7 +381,7 @@ STDAPI HCHttpCallRequestSetRequestBodyString(
     ) noexcept;
 
 #if !HC_NOZLIB
-#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK || HC_PLATFORM == HC_PLATFORM_NINTENDO_SWITCH
+#if HC_PLATFORM == HC_PLATFORM_WIN32 || HC_PLATFORM == HC_PLATFORM_GDK || HC_PLATFORM == HC_PLATFORM_NINTENDO_SWITCH || HC_PLATFORM == HC_PLATFORM_IOS || HC_PLATFORM == HC_PLATFORM_MAC
```
nassosterz-ms, line 384: could you replace with: HC_PLATFORM_IS_APPLE? Same applies to other places.

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
```diff
@@ -1606,7 +1788,7 @@
 					"HC_LINK_STATIC=1",
 					"HC_TRACE_BUILD_LEVEL=HC_PRIVATE_TRACE_LEVEL_VERBOSE",
 				);
-				GCC_TREAT_WARNINGS_AS_ERRORS = YES;
+				GCC_TREAT_WARNINGS_AS_ERRORS = NO;
```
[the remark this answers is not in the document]
rgomez391: Since we're building Zlib from source and not having a separate static lib, there are several warnings that are treated as errors from Zlib source files, we had to do the same on the Win32 solution, however, after some reading it seems that XCode does not provide a way to only do that for specific files. 

Do you know if there's a way we can do it only for some of them?

Include/httpClient/httpClient.h
rgomez391: sure, that makes sense!

rgomez391: Fixed

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
nassosterz-ms: What are the warnings that are treated as errors? There might a way to specifically not treat these warnings as errors I think. Like we can exclude them 1 by 1.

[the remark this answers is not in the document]
rgomez391: these are a couple of examples:

Screenshot 2023-11-09 at 12 57 54

image

If we can ignore those errors for specific files, I'd prefer that over doing it in the whole project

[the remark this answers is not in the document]
rgomez391: Fixed! Just to explain what I ended up doing just in case someone needs this in the future... I applied specific compiler flags to treat specific errors as warnings only for Zlib files.

In XCode to apply a compiler flag to a specific file navigate to the Target, Build Phases -> Compile Sources. Select the file you want and specify the flags in the Compiler Flags column.

These are the flags I used:

-Wno-error=implicit-int-conversion
-Wno-error=shorten-64-to-32
-Wno-error=macro-redefined
-Wno-error=comma

Taken from:

https://clang.llvm.org/docs/UsersManual.html#cmdoption-Wno-error

Full list of warnings and errors in Clang:

https://clang.llvm.org/docs/DiagnosticsReference.html

Build/libHttpClient.Apple.C/libHttpClient.xcodeproj/project.pbxproj
```diff
@@ -2077,7 +2261,7 @@
 					"HC_LINK_STATIC=1",
 					"HC_TRACE_BUILD_LEVEL=HC_PRIVATE_TRACE_LEVEL_IMPORTANT",
 				);
-				GCC_TREAT_WARNINGS_AS_ERRORS = YES;
+				GCC_TREAT_WARNINGS_AS_ERRORS = NO;
```
nassosterz-ms, line 2264: Should this still be NO?

rgomez391: Missed that one, I changed all of them to be YES now

nassosterz-ms (approved):

PR #787, opened 2023-11-10
fix crash when large buffers are sent through HCTrace

tculotta:
HCTraceImplMessage_v used to call vsnprintf, this changes the new safe call to the equivalent of that which allows truncating of large logs. There's an unfortunate difference between the two functions though, where vsnprintf would return the number of characters written even when the message was truncated, _vsnprintf_s for some reason returns -1 in that case. So now there's an additional check to avoid throwing valid logs away when the -1 is from truncating rather than an actual error.

### DIFF
--- a/Source/HTTP/Curl/CurlEasyRequest.cpp
+++ b/Source/HTTP/Curl/CurlEasyRequest.cpp
@@ -113,7 +113,7 @@ Result<HC_UNIQUE_PTR<CurlEasyRequest>> CurlEasyRequest::Initialize(HCCallHandle 
     RETURN_IF_FAILED(HCHttpCallRequestGetTimeout(hcCall, &timeoutSeconds));
     RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_TIMEOUT_MS, timeoutSeconds * 1000));
 
-    RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_VERBOSE, 1)); // verbose logging (0 off, 1 on)
+    RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_VERBOSE, 0)); // verbose logging (0 off, 1 on)
     RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_HEADER, 0)); // do not write headers to the write callback
     RETURN_IF_FAILED(easyRequest->SetOpt<char*>(CURLOPT_ERRORBUFFER, easyRequest->m_errorBuffer));
 
@@ -354,7 +354,7 @@ int CurlEasyRequest::DebugCallback(CURL* /*curlHandle*/, curl_infotype type, cha
         size -= 1;
     }
 
-    HC_TRACE_INFORMATION(HTTPCLIENT, "CURL %10s - %.*s", event, size, data);
+    HC_TRACE_VERBOSE(HTTPCLIENT, "CURL %10s - %.*s", event, size, data);
 
     return CURLE_OK;
 }
--- a/Source/Logger/trace.cpp
+++ b/Source/Logger/trace.cpp
@@ -280,9 +280,9 @@ STDAPI_(void) HCTraceImplMessage_v(
 
     char message[4096] = {};
 
-    auto result = vsprintf_s(message, format, varArgs);
+    auto result = _vsnprintf_s(message, sizeof(message), _TRUNCATE, format, varArgs);
 
-    if (result < 0)
+    if (result < 0 && strlen(message) != sizeof(message) - 1)
     {
         return;
     }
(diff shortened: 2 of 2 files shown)


Source/Logger/trace.cpp
```diff
@@ -280,9 +280,9 @@ STDAPI_(void) HCTraceImplMessage_v(
 
     char message[4096] = {};
 
-    auto result = vsprintf_s(message, format, varArgs);
+    auto result = _vsnprintf_s(message, sizeof(message), _TRUNCATE, format, varArgs);
 
-    if (result < 0)
+    if (result < 0 && strlen(message) != sizeof(message) - 1)
```
lubeltra, line 285: we should check errno instead

see here https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/vsnprintf-s-vsnprintf-s-vsnprintf-s-l-vsnwprintf-s-vsnwprintf-s-l?view=msvc-170

lubeltra: (basically fail if errno is not 0)

```cpp
errno = 0;
<print>

if (errno != 0)
{
    return;
}

**Review by lubeltra — CHANGES_REQUESTED** [review]

File: Source/Logger/trace.cpp (hunk 1 above)
Comment (lubeltra) [reply Source/Logger/trace.cpp:285]: and also use the template version of the function if at all possible

Comment (tculotta) [reply Source/Logger/trace.cpp]: done

File: Source/HTTP/Curl/CurlEasyRequest.cpp [hunk 2]
```diff
@@ -113,7 +113,7 @@ Result<HC_UNIQUE_PTR<CurlEasyRequest>> CurlEasyRequest::Initialize(HCCallHandle
     RETURN_IF_FAILED(HCHttpCallRequestGetTimeout(hcCall, &timeoutSeconds));
     RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_TIMEOUT_MS, timeoutSeconds * 1000));
 
-    RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_VERBOSE, 1)); // verbose logging (0 off, 1 on)
+    RETURN_IF_FAILED(easyRequest->SetOpt<long>(CURLOPT_VERBOSE, 0)); // verbose logging (0 off, 1 on)
```
tculotta, line 116: Meant to include this initially. This was unintentionally left enabled and lost in the big PAL refactor

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -354,7 +354,7 @@ int CurlEasyRequest::DebugCallback(CURL* /*curlHandle*/, curl_infotype type, cha
         size -= 1;
     }
 
-    HC_TRACE_INFORMATION(HTTPCLIENT, "CURL %10s - %.*s", event, size, data);
+    HC_TRACE_VERBOSE(HTTPCLIENT, "CURL %10s - %.*s", event, size, data);
```
tculotta, line 357: Moving the XCurl debug logging to a more appropriate level

jplafonta (approved):
:shipit:

lubeltra (approved):

tculotta:
Had to reroute through wrappers for platforms like mobile to work. vstprintf_s naming to align with previous wrapper names

Source/Logger/trace.cpp
```diff
@@ -43,13 +43,29 @@ int sprintf_s(char* buffer, size_t size, _Printf_format_string_ char const* form
...
 template<size_t SIZE>
-int vsprintf_s(char(&buffer)[SIZE], _Printf_format_string_ char const* format, va_list varArgs) noexcept
+int vstprintf_s(char(&buffer)[SIZE], _Printf_format_string_ char const* format, va_list varArgs) noexcept
 {
     return vsnprintf(buffer, SIZE, format, varArgs);
 }
 #endif
 #endif
 
+#if HC_PLATFORM_IS_MICROSOFT || HC_PLATFORM_IS_PLAYSTATION
```
lubeltra, line 53: why the `|| HC_PLATFORM_IS_PLAYSTATION`? I'm pretty sure you'd end up with a broken definition

lubeltra: and if you don't need the playstation bit, I'd move this to be in an if/else with the block above

tculotta: In the current version that redefines vsprintf_s, it only redefines that when not Microsoft and not Playstation (the same logic above). So this new redefinition needs to apply for both of those. Unless you're suggesting we should keep vsprintf_s for Playstation.

tculotta: Honestly we should probably just get @nassosterz-ms to weigh in here since he's actually in the ps stuff

lubeltra: oh right, I misread the if above

tculotta: Changed according to chat with Nassos

nassosterz-ms (approved):
Approving for the part that is relevant to PlayStation

PR #788, opened 2023-11-13
Fix crash caused by race condition in WinHttp teardown

jplafonta:
Upon receiving a suspend event, the WinHttp provider will automatically tear down any active WebSocket connections.  Depending on the current state of the connection, the suspend handler will either call WinHttpWebSocketClose followed by WinHttpCloseHandle, or just call WinHttpCloseHandle directly.  If we receive a WinHttp disconnect callback at the same time, the disconnect handler will call WinHttpCloseHandle.  This leads to race condition between the two handlers:

If the suspend handler runs first and determines that it needs to call WinHttpWebSocketClose (because at that point we believe the WebSocket is still connected) and then the disconnect handler runs and calls WinHttpCloseHandle BEFORE the suspend handler actually calls WinHttpWebSocketClose, the WinHttp handle may be in an invalid state when WebSocketClose finally gets called, leading to a crash.

The fix here involves a couple things. First, I updated the suspend handler to forego the call to WinHttpWebSocketClose altogether.  This means that during suspend the WebSocket won't be torn down gracefully, but it greatly simplifies the LHC teardown process.  Second, there is some additional logic needed to make sure that LHC still raises a disconnected event to clients during suspend, as the flow will change slightly with the removal of the WinHttpWebSocketClose call.

### DIFF
--- a/Source/HTTP/WinHttp/winhttp_connection.cpp
+++ b/Source/HTTP/WinHttp/winhttp_connection.cpp
@@ -1316,7 +1316,7 @@ void WinHttpConnection::StartWinHttpClose()
     {
         win32_cs_autolock autoCriticalSection(&m_lock);
 
-        if (m_state == ConnectionState::WinHttpClosing || m_state == ConnectionState::Closed)
+        if (m_state == ConnectionState::WebSocketClosing || m_state == ConnectionState::WinHttpClosing || m_state == ConnectionState::Closed)
         {
             HC_TRACE_VERBOSE(HTTPCLIENT, "WinHttpConnection::StartWinHttpClose called while already closing, ignored");
             return;
--- a/Source/HTTP/WinHttp/winhttp_connection.cpp
+++ b/Source/HTTP/WinHttp/winhttp_connection.cpp
@@ -1316,7 +1316,7 @@ void WinHttpConnection::StartWinHttpClose()
     {
         win32_cs_autolock autoCriticalSection(&m_lock);
 
-        if (m_state == ConnectionState::WebSocketClosing || m_state == ConnectionState::WinHttpClosing || m_state == ConnectionState::Closed)
+        if (m_state == ConnectionState::WinHttpClosing || m_state == ConnectionState::Closed)
         {
             HC_TRACE_VERBOSE(HTTPCLIENT, "WinHttpConnection::StartWinHttpClose called while already closing, ignored");
             return;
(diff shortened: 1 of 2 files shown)


jasonsandlin (approved):
:shipit:

Source/HTTP/WinHttp/winhttp_connection.cpp
```diff
@@ -430,13 +428,7 @@ HRESULT WinHttpConnection::Close(ConnectionClosedCallback callback)
         }
     }
 
-    if (doWebSocketClose)
-    {
-        assert(m_winHttpWebSocketExports.close);
-        DWORD result = m_winHttpWebSocketExports.close(m_hRequest, static_cast<USHORT>(HCWebSocketCloseStatus::GoingAway), nullptr, 0);
```
ScottMunroMS, line 436: I have some questions about this new behavior. I'll ping you on teams to follow up

Source/Logger/trace.cpp
```diff
@@ -129,7 +139,7 @@ void FormatTrace(
 #endif
 
     // [threadId][level][time][area] message
-    auto written = sprintf_s(outputBuffer, "[%04llX][%s][%02d:%02d:%02d.%03u][%s] %s",
+    auto written = _snprintf_s(outputBuffer, SIZE - 3, "[%04llX][%s][%02d:%02d:%02d.%03u][%s] %s",
```
jasonsandlin, line 142: why pass SIZE - 3 to a function that ignores the param.  count is commented out above

jplafonta: its used for the windows definition.  The definition above is for non-windows only

Source/Logger/trace.cpp
```diff
@@ -41,6 +41,16 @@ int sprintf_s(char* buffer, size_t size, _Printf_format_string_ char const* form
     return result;
 }
 
+template<size_t SIZE>
+int _snprintf_s(char(&buffer)[SIZE], size_t /*count*/, _Printf_format_string_ char const* format ...) noexcept
+{
+    va_list varArgs{};
+    va_start(varArgs, format);
+    auto result = vsnprintf(buffer, SIZE, format, varArgs);
```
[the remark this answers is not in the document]
ScottMunroMS: aren't the OS compliant *safe* variants the thing you just had to fix in the other tracing fix PR? those were causing crashes from calling fatal constraint handlers and what was really desirable was just getting the truncation?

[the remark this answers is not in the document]
jplafonta: This is a PAL version for non-windows only

jplafonta: this is for non-windows only

PR #791, opened 2024-01-11
Adding build material for windows dlls and updating samples

jplafonta:
* Add project files for libHttpClient.GDK.dll and libHttpClient.Win32.dll
* Add props file libHttpClient.import.props which adds references to the appropriate lhc projects.  By default, it will add a reference to the dll project, but this can be overridden by specifying the build property HCStaticLib=true prior to importing the props file.  The existing libHttpClient.props file will continue to add references to the old static lib projects as to not break existing clients.


Build/libHttpClient.GDK/libHttpClient.GDK.log
```diff
@@ -0,0 +1,25 @@
+﻿  Building against GXDK Edition 231001.
```
jasonsandlin, line 1: .gitignore log files

Build/libHttpClient.Win32/libHttpClient.Win32.log
```diff
@@ -0,0 +1,27 @@
+﻿  uri.cpp
```
jasonsandlin, line 1: remove

Include/XAsync.h
```diff
@@ -69,7 +69,7 @@ struct XAsyncBlock
 /// </summary>
 /// <param name='asyncBlock'>A pointer to the XAsyncBlock that was passed to the async call.</param>
 /// <param name='wait'>If true, XAsyncGetStatus waits until the async call either completes or is canceled.</param>
-STDAPI XAsyncGetStatus(
+HC_API XAsyncGetStatus(
```
jasonsandlin, line 72: XAsync* is shared source from Brian, so changing it might not be wise

jasonsandlin (approved):

Include/XAsync.h
jplafonta: Good point.  Another option would be to export the XAsync APIs (or even all the LHC APIs) with a .def file.  I find that to be slightly more annoying to maintain but maybe its worth it to leave the source untouched in this case.  Thoughts?

jasonsandlin: def file is better.  this api list doesn't change often. i'd rather be able to keep the source in sync.

PR #792, opened 2024-01-16
Properly honor the subprotocol specificied in HCWebSocketConnect

jplafonta:
WebSocket subprotocol parameter is ignored in WinHttp WebSocket stack.  If specified, this should set the "Sec-WebSocket-Protocol" header during the connect.


tculotta (approved):
:shipit:

PR #800, opened 2024-02-15
Switch to shared library for linux

jplafonta:
* Change LHC to be a shared library on Linux, exposing only methods specified in libHttpClientExports.txt
* Updates lib name and output paths to be consistent with other platforms


jasonsandlin (approved):

SebastianPD (approved):
My only nit is to go into the read me and change to the new names of the bach files. Instead of openssl.bash it's now openssl_linux.bash

SebastianPD (approved):

PR #809, opened 2024-03-05
Use less `#if` to handle `HC_NOZLIB` also fix `HC_NOWEBSOCKET`

lubeltra:


Build/libHttpClient.Common/libHttpClient.Common.vcxitems
```diff
@@ -27,6 +27,8 @@
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\Include\httpClient\mock.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\Include\httpClient\pal.h" />
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\Include\httpClient\trace.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\Include\httpClient\types_android.h" />
```
jplafonta, line 30: > <ClInclude Include="$(MSBuildThisFileDirectory)..\..\Include\httpClient\types_android.h" /> [](http://example.com/codeflow?start=3&length=92)

Should this be added here?

Include/httpClient/config.h
```diff
@@ -6,7 +6,6 @@
 #define HC_PLATFORM_UWP 2
 #define HC_PLATFORM_XDK 3
 #define HC_PLATFORM_GDK 4
-#define HC_PLATFORM_GRTS 5
```
jplafonta, line 9: >HC_PLATFORM_GRTS [](http://example.com/codeflow?start=8&length=16)

is this not needed for backward compat anymore?

Source/HTTP/compression.cpp
```diff
@@ -2,22 +2,24 @@
...
 #define CHUNK 16384
 #define WINDOWBITS 15
 #define GZIP_ENCODING 16
 
 NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
 
-Compression::Compression()
+bool Compression::Available() noexcept
 {
+    return false;
```
jplafonta, line 19: >false [](http://example.com/codeflow?start=11&length=5)

true?

Source/HTTP/httpcall_request.cpp
```diff
@@ -108,6 +107,11 @@ try
     }
     RETURN_IF_PERFORM_CALLED(call);
 
+    if (!Compression::Available() && level != HCCompressionLevel::None)
+    {
+        return E_NOT_SUPPORTED;
+    }
+
```
jplafonta, line 114: Any concern about this being a runtime failure as opposed to knowing its not supported at compile time?

Build/libHttpClient.Common/libHttpClient.Common.vcxitems
lubeltra: you are right, it's redundant

Include/httpClient/config.h
lubeltra: we only ever used this internally in grts itself and we do not use it anymore

Source/HTTP/compression.cpp
lubeltra: I am a derp

Source/HTTP/httpcall_request.cpp
lubeltra: As discussed offline, the client should be free to just ignore the error, in the end they still want the request to be made compression or not.

PR #815, opened 2024-04-12
Enable Gzip Response Decompression

NandoCT1014:


Include/httpClient/httpClient.h
```diff
@@ -398,6 +398,12 @@ STDAPI HCHttpCallRequestEnableGzipCompression(
     _In_ HCCompressionLevel level
 ) noexcept;
 
+// TODO: Add summary
+STDAPI HCHttpCallSetCompressedResponse(
```
jplafonta, line 402: The client calls this indicate that it wants the response to be compressed correct? If that's the case 'HCHttpCallSetCompressResponse' makes more sense to me.

Source/HTTP/httpcall.cpp
```diff
@@ -101,6 +101,12 @@ HRESULT CALLBACK HC_CALL::PerfomAsyncProvider(XAsyncOp op, XAsyncProviderData co
         }
         else
         {
+            // If Custom ReponseWriteFunction is specified and compressedResponse is specified reset to default response body write callback
```
jplafonta, line 104: Won't you need to store the client's custom callback still? So that you can call it when the request completes and you've decompressed the response?

Source/HTTP/httpcall.cpp
```diff
@@ -353,6 +363,20 @@ void HC_CALL::PerformSingleRequestComplete(XAsyncBlock* async)
...
+    if (Compression::Available() && call->compressedResponse == true)
+    {
+        http_internal_vector<uint8_t> uncompressedResponseBodyBuffer;
+
+        Compression::DecompressFromGzip(
+            call->responseBodyBytes.data(),
+            call->responseBodyBytes.size(),
+            uncompressedResponseBodyBuffer);
+
+        call->responseBodyBytes.resize(uncompressedResponseBodyBuffer.size());
```
jplafonta, line 376: I think here you'll need an extra check to see if the client had specified a response body write callback. If they did, you'd call that here.  If not, you copy the decompressed response to call->responseBodyBytes (like you've done) and they'll read it with HCHttpCallResponseGetResponseBodyBytes after the request completes.

jplafonta (changes requested):

Source/HTTP/httpcall.cpp
NandoCT1014: Added HCHttpCallResponseBodyWriteFunction temporaryResponseBodyWriteFunction and temporaryResponseBodyWriteFunctionContext fields to the HC_CALL struct to store the client's custom callback if these conditions are met

Build/libHttpClient.GDK/libHttpClient.GDK.def
```diff
@@ -16,6 +16,7 @@ EXPORTS
     HCHttpCallGetRequestUrl
     HCHttpCallPerformAsync
     HCHttpCallRequestEnableGzipCompression
+    HCHttpCallSetCompressResponse
```
jplafonta, line 19: Add new public APIs to the end of the .def file.  This allows hot swapping a new dll into a game built with an old import lib without things breaking.

Samples/Win32-Http/main.cpp
```diff
@@ -178,17 +178,33 @@ struct SampleHttpCallAsyncContext
...
 
-void DoHttpCall(std::string url, std::string requestBody, bool isJson, std::string filePath, bool enableGzipCompression)
+void DoHttpCall(std::string url, std::string requestBody, bool isJson, std::string filePath, bool enableGzipCompression, bool enableGzipResponseCompression)
 {
     std::string method = "GET";
     bool retryAllowed = true;
     std::vector<std::vector<std::string>> headers;
     std::vector< std::string > header;
 
+    if (enableGzipResponseCompression) {
```
jplafonta, line 188: nit: put curly brackets on separate lines

Include/httpClient/httpClient.h
```diff
@@ -641,6 +653,22 @@ STDAPI HCHttpCallResponseSetResponseBodyWriteFunction(
...
+/// Used to temporarily store a custom callback function that will be used to write the response body after the HTTP call
+/// has been completed. This should be called in the case where gzip compression has enabled for responses AND
+/// a custom response write callback function has been specified. 
+/// </summary>
+/// <param name="call">The handle of the HTTP call.</param>
+/// <param name="writeFunction">The custom response body write function this call should use.</param>
+/// <param name="context">The context to associate with this write function.</param>
+/// <returns>Result code of this API operation. Possible values are S_OK or E_INVALIDARG.</returns>
+/// <remarks>This must be called prior to calling HCHttpCallPerformAsync.</remarks>
+STDAPI HCHttpCallResponseSetTemporaryResponseBodyWriteFunction(
```
jplafonta, line 666: I don't think this needs to be a public API, correct? isn't this only something LHC should do internally?

Utilities/FrameworkResources/exports.exp
```diff
@@ -18,6 +18,7 @@ _HCHttpCallGetId
 _HCHttpCallSetTracing
 _HCHttpCallGetRequestUrl
 _HCHttpCallRequestEnableGzipCompression
+_HCHttpCallSetCompressResponse
```
jplafonta, line 21: Not sure if the same comment about the .def files applies for other platforms, but might be worth confirming

Source/HTTP/httpcall.h
```diff
@@ -66,6 +66,9 @@ struct HC_CALL
     xbox::httpclient::HttpHeaders responseHeaders{};
     HCHttpCallResponseBodyWriteFunction responseBodyWriteFunction{ HC_CALL::ResponseBodyWrite };
     void* responseBodyWriteFunctionContext{ nullptr };
+    HCHttpCallResponseBodyWriteFunction temporaryResponseBodyWriteFunction{ HC_CALL::ResponseBodyWrite };
```
jplafonta, line 69: Maybe the logic is a bit cleaner if you call this something like "clientResponseBodyWriteFunction"?  and I'd probably initialize to nullptr

jplafonta: either way, add a quick comment describing why this is needed

jplafonta (changes requested):

Source/HTTP/httpcall.cpp
NandoCT1014: Is a call to the custom write callback still needed here? Currently its missing....I'm not quite sure how we would call it I similar to the logic in [http_android.cpp](https://github.com/microsoft/libHttpClient/blob/565b98f779b13321372263b6addeab263cd518e0/Source/HTTP/Android/http_android.cpp#L254)?

Source/HTTP/httpcall.cpp
```diff
@@ -353,6 +368,32 @@ void HC_CALL::PerformSingleRequestComplete(XAsyncBlock* async)
...
+        call->responseBodyBytes = std::move(uncompressedResponseBodyBuffer);
+    }
+
+    // Check if we 'reset' the custom response write callback before decompressing the response
+    HCHttpCallResponseBodyWriteFunction temporaryWriteFunction = call->clientResponseBodyWriteFunction;
+
+    // call->temporaryResponseBodyWriteFunction should remain HC_CALL::ResponseBodyWrite if we did not 'reset' the custom response write callback
```
jplafonta, line 388: Update comment.  I also think this could be safely moved inside the above if statement, though I don't think it makes a big difference

jplafonta (approved):

Samples/Win32-Http/main.cpp
```diff
@@ -263,24 +279,27 @@ void DoHttpCall(std::string url, std::string requestBody, bool isJson, std::stri
...
+            if (str != nullptr) responseString = str;
+
+            if (!isJson)
+            {
+                size_t bufferSize = 0;
+                HCHttpCallResponseGetResponseBodyBytesSize(call, &bufferSize);
+                uint8_t* buffer = new uint8_t[bufferSize];
+                size_t bufferUsed = 0;
+                HCHttpCallResponseGetResponseBodyBytes(call, bufferSize, buffer, &bufferUsed);
+                HANDLE hFile = CreateFileA(filePath.c_str(), GENERIC_WRITE, 0, NULL, CREATE_NEW, FILE_ATTRIBUTE_NORMAL, NULL);
```
jplafonta, line 294: Not exactly sure what this write file bit doing, but I don't think that it happens for the customWriteUsed flow.

Samples/Win32-Http/main.cpp
```diff
@@ -176,9 +176,19 @@ struct SampleHttpCallAsyncContext
...
+    std::vector<uint8_t> response;
+    bool isCustom;
 };
 
-void DoHttpCall(std::string url, std::string requestBody, bool isJson, std::string filePath, bool enableGzipCompression, bool enableGzipResponseCompression)
+HRESULT CustomResponseBodyWrite(HCCallHandle call, const uint8_t* source, size_t bytesAvailable, void* context) 
+{
+    SampleHttpCallAsyncContext* customContext = static_cast<SampleHttpCallAsyncContext*> (context);
+    customContext->response.insert(customContext->response.end(), source, source + bytesAvailable);
+    customContext->response.push_back('\0');
```
jplafonta, line 187: I don't think you want to append a null character each time this method is called.  This may be called several times with a chunk of the response each time, so you'd be inserting some null characters into the middle of the response.  I'd append a null terminator in the HTTP request completion callback below.

jplafonta (approved):

Samples/Win32-Http/main.cpp
NandoCT1014: The file writing should be skipped when a custom write is used since its also encapsulated in the !customWriteUsed if statement.

PR #816, opened 2024-04-18
Fix subtle race in task queue

brianpepin:
The task queue has a narrow race: if a future callback is evaluated just when the task queue is terminated it may get skipped from termination processing. The result is that a call for the future won't be canceled immediately when the task queue terminates.  Instead it will be canceled when its due time occurs.

There is a second possible race as well.  If a new future callback is scheduled and the schedule code is interleaved with a terminate call on another thread, the same thing can occur.

This change fixes both cases. It also fixes up some incorrect macros in the test projects so they can build again.


jasonsandlin (approved):

jasonsandlin:
pr moved and completed

PR #820, opened 2024-04-29
Updating props file for XSAPI GDK Migration to LHC dll

rgomez391:


libHttpClient.props
```diff
@@ -57,15 +58,17 @@
 
   <ItemGroup>
     <ProjectReference Include="$(HCRoot)\Build\$(HCProjectName)\$(HCProjectName).vcxproj" Condition="'$(ConfigurationType)'!='StaticLibrary' And ('$(HCLibPlatformType)' == 'Win32' Or '$(HCLibPlatformType)' == 'GDK' Or '$(HCLibPlatformType)' == 'UWP')">
-      <Project Condition="'$(HCLibPlatformType)' == 'Win32' and '$(HCLibToolset)' == '141'">{BEDE280F-17F7-4A09-9F5D-2E02FBB1C490}</Project>
```
jplafonta, line 60: Rather than updating this props file, could you change XSAPI to add LHC references with \build\libHttpClient.import.props?

In the PR where I initially shifted to shared libs for Windows, I wanted to keep the libHttpClient.props file unchanged so that clients currently using LHC as a static lib wouldn't be impacted.  You can see how the samples were updated in that same PR here: https://github.com/microsoft/libHttpClient/pull/791

jplafonta (changes requested):

jplafonta (approved):

libHttpClient.props
rgomez391: After our offline conversation, this has been rolled back and I've just moved XSAPI to use the libHttpClient.import.props file

PR #822, opened 2024-05-02
Adding ZLIB source files to Android Workspace

rgomez391:


Build/libHttpClient.CMake/GetLibHCFlags.cmake
```diff
@@ -19,18 +18,20 @@ function(GET_LIBHC_FLAGS OUT_FLAGS OUT_FLAGS_DEBUG OUT_FLAGS_RELEASE)
...
     set("${OUT_FLAGS_DEBUG}"
         "-O0"
-        "-std=c++17"
         "-DHC_TRACE_BUILD_LEVEL=HC_PRIVATE_TRACE_LEVEL_VERBOSE"
         PARENT_SCOPE
         )
 
     set("${OUT_FLAGS_RELEASE}"
         "-Os"
-        "-std=c++17"
```
jplafonta, line 33: Why was this moved from here to both the linux and Android files?

jplafonta (approved):

Build/libHttpClient.CMake/GetLibHCFlags.cmake
rgomez391: Because if it's placed as part of the FLAGS, OUT_FLAGS_DEBUG or OUT_FLAGS_RELEASE variables the **-std=c++17** flag will be used to compile to all the C and C++ source files. This was causing compilation errors because ZLIB source files are C files and you can't specify **-std=c++17** flag for C files. What I had to do is to include the **-std=c++17** through **set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")** which applies the flag only to C++ files.

PR #831, opened 2024-06-19
Adding Progress report callbacks for Win32, GDK, Linux, Switch, Apple and Android

rgomez391:


Source/HTTP/httpcall_request.cpp
```diff
@@ -268,6 +302,33 @@ try
...
+{
+    if (call == nullptr || progressReportFunction == nullptr)
+    {
+        return E_INVALIDARG;
+    }
+
+    if (isUploadFunction)
+    {
+        *progressReportFunction = call->uploadProgressReportFunction;
+
```
jplafonta, line 321: nit: remove newline below

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -40,12 +40,30 @@ Result<HC_UNIQUE_PTR<CurlEasyRequest>> CurlEasyRequest::Initialize(HCCallHandle
     void* clientRequestBodyReadCallbackContext{};
     RETURN_IF_FAILED(HCHttpCallRequestGetRequestBodyReadFunction(hcCall, &clientRequestBodyReadCallback, &bodySize, &clientRequestBodyReadCallbackContext));
 
+// Specify libcurl progress callback and create libcurl progress callback for non-GDK platforms since XCurl doesn't support libcurl progress callback
+#if HC_PLATFORM != HC_PLATFORM_GDK
```
jplafonta, line 44: is the alternative implementation for XCurl already done somewhere that I'm missing or is that still pending?

jplafonta (approved):
:shipit:

Source/HTTP/httpcall_request.cpp
```diff
@@ -198,6 +198,40 @@ try
...
+{
+    if (call == nullptr || progressReportFunction == nullptr || minimumProgressReportInterval <= 0)
+    {
+        return E_INVALIDARG;
+    }
+    RETURN_IF_PERFORM_CALLED(call);
+
+    auto httpSingleton = get_http_singleton();
+    if (nullptr == httpSingleton)
+        return E_HC_NOT_INITIALISED;
```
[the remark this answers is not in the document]
rgomez391: noted, yeah, I've seen some legacy code that don't use brackets. I've corrected the ones I found in this file

Source/HTTP/Curl/CurlEasyRequest.cpp
rgomez391: It's already implemented in lines 259 and 397, instead of using the built-in progress callback from XCurl, we're reporting progress from Write and Read callbacks from WinHttp.

PR #832, opened 2024-06-24
Optimize Linux Builds by Checking for Previously Generated Bins

SahilAshar:
Similar to what we do for iOS OpenSSL builds, check for existing bins before rebuilding the entire project each time the build is invoked.

https://github.com/microsoft/libHttpClient/blob/main/Utilities/XcodeBuildScripts/openssl.bash#L50

### DIFF
--- a/Build/libHttpClient.Linux/openssl_Linux.bash
+++ b/Build/libHttpClient.Linux/openssl_Linux.bash
@@ -42,6 +42,13 @@ if [ ! -d /usr/local/ssl/include/openssl ] ; then
     exit 1
 fi
 
+if [ -f "$SCRIPT_DIR/../../Out/x64/$CONFIGURATION/libcrypto.Linux/libcrypto.a" ]; then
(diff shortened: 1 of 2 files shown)


nassosterz-ms (approved):

Build/libHttpClient.Linux/openssl_Linux.bash
```diff
@@ -42,6 +42,13 @@ if [ ! -d /usr/local/ssl/include/openssl ] ; then
     exit 1
 fi
 
+if [ -f "$SCRIPT_DIR/../../Out/x64/$CONFIGURATION/libcrypto.Linux/libcrypto.a" ]; then
```
SebastianPD, line 45: What about libssl or should we not worry about it since the expectation is that if crypto is built then so should ssl?

SebastianPD (approved):

Build/libHttpClient.Linux/openssl_Linux.bash
SahilAshar: The latter - since we're building OpenSSL via script and not source, the script will trigger the libcrypto build each time, prior to building libssl. I also wanted to match the behavior of the iOS build script.

PR #836, opened 2024-07-02
Remove magic statics for timerqueue

brianpepin:
waittimer uses a global timerqueue object.  But timerqueue's dtor destroys a thread, which has allocated resources. This can cause conflicts with other static / global destructors in the system. We should not be using "magic statics" to do complex things like freeing dynamic memory.

This changes the global timerqueue to a shared_ptr. In the destructor for waittimer, it checks the shared pointer usage count and is this is the last wait timer, it will free the global shared pointer. This causes the thread deallocation to happen when the last task queue is closed instead of happening post-main in crt cleanup.

Tested with a repro that exposed the magic static fragility (crash in custom global new operator). Verified all works and no leaks post change (thanks Sebastian Perez-Delgado).

### DIFF
--- a/Source/Task/WaitTimer_stl.cpp
+++ b/Source/Task/WaitTimer_stl.cpp
@@ -216,7 +216,7 @@ namespace OS
         {
             try
             {
-                auto queue = std::make_shared<TimerQueue>();
+                auto queue = http_allocate_shared<TimerQueue>();
                 if (!queue->Init())
                 {
                     return E_FAIL;
--- a/Source/Task/WaitTimer_stl.cpp
+++ b/Source/Task/WaitTimer_stl.cpp
@@ -5,9 +5,12 @@ using Deadline = std::chrono::high_resolution_clock::time_point;
 
 namespace OS
 {
+    class TimerQueue;
+
     class WaitTimerImpl
     {
     public:
+        ~WaitTimerImpl();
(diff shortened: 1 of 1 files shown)


Source/Task/WaitTimer_stl.cpp
```diff
@@ -186,27 +188,61 @@ namespace OS
...
 
-        if (!g_timerQueue.LazyInit())
+        std::lock_guard<std::mutex> lock{ g_timerQueueMutex };
+
+        if (g_timerQueue == nullptr)
         {
-            return E_FAIL;
+            try
+            {
+                auto queue = std::make_shared<TimerQueue>();
```
jasonsandlin, line 219: change to:
                auto queue = http_allocate_shared<TimerQueue>();
                
this will use the mem hooked alloc.  see global.cpp for example usage

jasonsandlin (approved):

SebastianPD (changes requested):
Everything works but I did have to make changes on my end to get it working. I received warning c4703 (Potentially uninitialized local pointer variable) which we treat as an error. I changed line 373 in httpcall.cpp to be initialized as a nullptr and it fixed everything. I did testing on the linux and windows platform and we are working as expected. 

I have been suggested by my team that we should get Luca's input when he comes back from break. Also, I haven't gotten confirmation that the solution worked for our partner yet.

SebastianPD (approved):

PR #837, opened 2024-07-08
Update httpcall.cpp

SebastianPD:
ensure encodingHeaderValue is initialized to not set off compiler warnings in Visual Studio


nassosterz-ms:
Adding @NandoCT1014 as original author of this code.

nassosterz-ms (approved):

PR #839, opened 2024-07-11
Updating WDK version for builds on Scarlett

AlonsoMondal:


Build/libHttpClient.GDK.props
```diff
@@ -24,7 +24,7 @@
     <TargetRuntime>Native</TargetRuntime>
     <CharacterSet>Unicode</CharacterSet>
     <GenerateManifest>false</GenerateManifest>
-    <WindowsTargetPlatformVersion>10.0.19041.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion Condition="'$(Platform)' == 'Gaming.Xbox.Scarlett.x64'">10.0.26100.0</WindowsTargetPlatformVersion>
```
jasonsandlin, line 27: previously this was set for Gaming.Desktop.x64 builds but this logic doesn't set it for that build flavor.  
seems like it should be.  aka add this line:

    <WindowsTargetPlatformVersion Condition="'$(Platform)' == 'Gaming.Desktop.x64'">10.0.26100.0</WindowsTargetPlatformVersion>

AlonsoMondal: Only Scarlett platform _requires_ 10.0.20000.0 or above. Would you still think that I should update that version when building on Desktop?

Also, to clarify, I should edit the platform condition to be equals to Desktop, and not add that additional line, right?
Is the removal of the original version setting still ok too?

jasonsandlin: Then add a new line that's something like this to keep existing logic in place:

<WindowsTargetPlatformVersion Condition="'$(Platform)' != ''Gaming.Xbox.Scarlett.x64'">10.0.19041.0</WindowsTargetPlatformVersion>

jasonsandlin (approved):

PR #841, opened 2024-08-01
Not use LHC when building merged SDK for apple platform

yuehanlin:


Include/httpClient/async.h
```diff
@@ -6,7 +6,7 @@
 #include <httpClient/config.h>
 #include <httpClient/pal.h>
 
-#if (!defined(HC_LINK_STATIC) || HC_LINK_STATIC == 0) && HC_PLATFORM_IS_APPLE
+#if (!defined(HC_LINK_STATIC) || HC_LINK_STATIC == 0) && HC_PLATFORM_IS_APPLE && (!defined(USE_PFCORE_SDK))
```
jplafonta, line 9: Seems strange that the include path here would have anything to do with PlayFab. I don't know enough about Apple builds to tell you why this conditional exists in the first place, but I suspect it has something to do with frameworks and/or how your iOS project is referencing libHttpClient.  Would setting HC_LINK_STATIC in your project do the trick?

jplafonta (changes requested):

Include/httpClient/async.h
rgomez391: I agree with John. I faced the same issue you're facing Yuehan... I was unable to understand why this condition exist... What I ended up doing is just leaving LHC the way it is and adjusting things on PlayFab Core/Services side, for example, when deploying the iOS and macOS SDK I moved the header files inside the httpClient folder. 

See: https://microsoft.visualstudio.com/Xbox/_git/PlayFab.C?path=/Pipelines/Tasks/ios-build.yml&version=GBfeature/gamesave&line=201&lineEnd=203&lineStartColumn=9&lineEndColumn=70&lineStyle=plain&_a=contents

@jplafonta or @nassosterz-ms any thoughts?

yuehanlin: I am thinking about the risks if I remove this condition check. I think this won't block PlayFab Core/Service build, but I don't know is there any other builds I need to aware of

yuehanlin: Loop @SahilAshar

rgomez391: I think it's risky to do it, I don't think we have knowledge of how many clients are consuming LHC, so we could potentially introduce a breaking change.

PR #844, opened 2024-10-18
Allow multiple trace client callbacks

jplafonta:


Source/Logger/trace.cpp
```diff
@@ -292,7 +302,17 @@ STDAPI_(void) HCTraceImplMessage_v(
...
+    for (size_t i = 0; i < _countof(traceState.clientCallbacks); ++i)
+    {
+        if (traceState.clientCallbacks[i])
+        {
+            haveClientCallback = true;
+            break;
+        }
+    }
+
+    if (haveClientCallback && !traceState.GetTraceToDebugger() && !traceState.GetEtwEnabled())
```
tculotta, line 315: Shouldn't this check !haveClientCallback?

Source/Logger/trace.cpp
```diff
@@ -194,10 +194,13 @@ void TraceMessageToClient(
...
 {
-    auto callback = GetTraceState().GetClientCallback();
-    if (callback)
+    TraceState& traceState{ GetTraceState() };
+    for (size_t i = 0; i < _countof(traceState.clientCallbacks); ++i)
     {
-        callback(areaName, level, threadId, timestamp, message);
+        if (traceState.clientCallbacks[i])
+        {
+            traceState.clientCallbacks[i](areaName, level, threadId, timestamp, message);
```
tculotta, line 202: Is this thread safe? What happens if titles remove a callback while this is happening?

tculotta (changes requested):

Source/Logger/trace.cpp
jplafonta: It should, good catch.

Source/Logger/trace.cpp
jplafonta: You're right. Its always technically been not thread safe in fact.  After thinking about it more, I think the best approach might just be to document that once set, a trace callback can't be removed and may be called until the library is cleaned up. I think grabbing mutexes each time we trace something is acceptable in terms of performance.

tculotta: Hm wait. The old way was getting a copy with "auto callback =...", right? Couldn't we keep that copy and check instead of checking clientCallbacks[i] directly?

jplafonta: I don't think that actually solves anything from the clients perspective.  Thinking about a sequence like this:

Thread 1:
TraceMessageToClient running, grabs non null from GetTraceState().GetClientCallback() and stores copy on stack.

Thread 2: 
Client calls HCTraceRemoveClientCallback, removes client callback and returns.  This should imply the client could client up any trace related state on their end.

Thread 1:
Only now call callback that we copied to stack.  Client may have cleaned up some state needed by that callback at this point, crashes possible

jplafonta: Obviously this can be solved with some locking, but I don't think thats a good approach given the overhead and frequency of trace callbacks

tculotta: > Client may have cleaned up some state needed by that callback at this point, crashes possible

Ah of course. Then yeah I agree with your solution in your first comment

PR #857, opened 2025-02-05
Fix cleanup race in CurlMulti that leads to a crash

jplafonta:
* Remove race condition between CurlMulti::CleanupAsyncProvider Begin releasing the CurlMulti mutex and CurlMulti::CleanupAsyncProvider DoWork deleting the CurlMulti.  The fix is to simply release the mutex in Begin prior to scheduling DoWork.
* Remove an extraneous case in CurlMulti::Perform where cleanup could be scheduled with a perform callback still pending. This should have been removed as part of [this](https://github.com/microsoft/libHttpClient/commit/e048d58ba951540a7f4e71e19e8c40f02fc77a8e) commit but was missed.  I believe that having that block may cause the cleanup provider to be scheduled multiple times, and could cause a double delete depending on the timing between scheduling and executing of the cleanup provider.

### DIFF
--- a/Source/HTTP/Curl/CurlMulti.cpp
+++ b/Source/HTTP/Curl/CurlMulti.cpp
@@ -94,16 +94,22 @@ HRESULT CALLBACK CurlMulti::CleanupAsyncProvider(XAsyncOp op, const XAsyncProvid
         assert(multi->m_cleanupAsyncBlock == nullptr);
         multi->m_cleanupAsyncBlock = data->async;
 
-        if (multi->m_taskQueueCallbacksPending == 0)
+        // If there are no pending task queue callbacks (and thus no pending HTTP requets), schedule cleanup immediately.
+        // If this condition is true, we're also guaranteed to be the final remaining reference to the CurlMulti.
+        bool cleanupImmediately = multi->m_taskQueueCallbacksPending == 0;
+        XTaskQueueHandle queue = multi->m_queue;
+
+        // Release the lock before going any further because the multi may be destroyed before Begin completes
+        lock.unlock();
+
+        if (cleanupImmediately)
         {
-            // No outstanding HTTP requests schedule cleanup immediately. This condition also ensures that we don't have
-            // a pending Perform call scheduled.
             RETURN_IF_FAILED(XAsyncSchedule(data->async, 0));
         }
         else
         {
             // Terminate the XTaskQueue. Cleanup will be completed after completing remaining HTTP requests
-            RETURN_IF_FAILED(XTaskQueueTerminate(multi->m_queue, false, nullptr, nullptr));
+            RETURN_IF_FAILED(XTaskQueueTerminate(queue, false, nullptr, nullptr));
         }
         return S_OK;
     }
@@ -162,9 +168,11 @@ void CALLBACK CurlMulti::TaskQueueCallback(_In_opt_ void* context, _In_ bool can
     }
 
     std::unique_lock<std::mutex> lock{ multi->m_mutex };
-    // If CurlMulti::CleanupAsync was called and there are no remaining task queue callbacks, schedule cleanup now
     if (--multi->m_taskQueueCallbacksPending == 0 && multi->m_cleanupAsyncBlock)
     {
+        // If CurlMulti::CleanupAsync was called and there are no remaining task queue callbacks, schedule cleanup now
+        // Release the lock before going any further because the multi may be destroyed before this method returns
+        lock.unlock();
         HRESULT hr = XAsyncSchedule(multi->m_cleanupAsyncBlock, 0);
         if (FAILED(hr))
         {
@@ -239,16 +247,6 @@ HRESULT CurlMulti::Perform() noexcept
 
         uint32_t delay = workAvailable ? 0 : PERFORM_DELAY_MS;
         ScheduleTaskQueueCallback(std::move(lock), delay);
-    }
-    else if (m_cleanupAsyncBlock)
-    {
-        // CleanupAsync was called and no remaining requests, schedule cleanup now
-        HRESULT hr = XAsyncSchedule(m_cleanupAsyncBlock, 0);
-        if (FAILED(hr))
-        {
-            HC_TRACE_ERROR_HR(HTTPCLIENT, hr, "CurlMulti: Failed to schedule CleanupAsyncProvider!");
-            assert(false);
-        }
     }
 
     return S_OK;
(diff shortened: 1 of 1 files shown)


ScottMunroMS:
>         HC_UNIQUE_PTR<CurlMulti> reclaim{ static_cast<CurlMulti*>(multi) };

something I noticed when I was skulking around the code on my own is that this static_cast is unnecessary. this is already the type of the `multi` var. maybe something that can be opportunistically cleaned up in this PR

---
Refers to: Source/HTTP/Curl/CurlMulti.cpp:119 in f848f72. [](commit_id = f848f72af32fb69bd0dfefd2954299ac8b4a2f16, deletion_comment = False)

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -94,16 +94,22 @@ HRESULT CALLBACK CurlMulti::CleanupAsyncProvider(XAsyncOp op, const XAsyncProvid
...
         multi->m_cleanupAsyncBlock = data->async;
 
-        if (multi->m_taskQueueCallbacksPending == 0)
+        // If there are no pending task queue callbacks (and thus no pending HTTP requets), schedule cleanup immediately.
+        // If this condition is true, we're also guaranteed to be the final remaining reference to the CurlMulti.
+        bool cleanupImmediately = multi->m_taskQueueCallbacksPending == 0;
+        XTaskQueueHandle queue = multi->m_queue;
+
+        // Release the lock before going any further because the multi may be destroyed before Begin completes
+        lock.unlock();
```
ScottMunroMS, line 103: it might help to clarify how/why within this comment

```
// Release the lock before scheduling any further async work. That async work may drive the next XAsyncOp::DoWork iteration of this loop before this XAsyncOp::Begin iteration completes. Because DoWork cleans up the "multi" object it becomes no longer safe to reference that multi object after the async work is scheduled.
```

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -94,16 +94,22 @@ HRESULT CALLBACK CurlMulti::CleanupAsyncProvider(XAsyncOp op, const XAsyncProvid
...
+
```
ScottMunroMS, line 104: nit: it might not be a bad idea to null out the `multi` temporary after this unlocks or after the async work is scehduled. Just as a guard against future programmer misuse to not reference the muli-object after scheduling the async work.

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -162,9 +168,11 @@ void CALLBACK CurlMulti::TaskQueueCallback(_In_opt_ void* context, _In_ bool can
     }
 
     std::unique_lock<std::mutex> lock{ multi->m_mutex };
-    // If CurlMulti::CleanupAsync was called and there are no remaining task queue callbacks, schedule cleanup now
     if (--multi->m_taskQueueCallbacksPending == 0 && multi->m_cleanupAsyncBlock)
     {
+        // If CurlMulti::CleanupAsync was called and there are no remaining task queue callbacks, schedule cleanup now
+        // Release the lock before going any further because the multi may be destroyed before this method returns
+        lock.unlock();
         HRESULT hr = XAsyncSchedule(multi->m_cleanupAsyncBlock, 0);
```
ScottMunroMS, line 176: Rather than an explicit unlock, you could use scoping to your advantage here

```
{
    auto multi = static_cast<CurlMulti*>(context);

    if (!canceled)
    {
        HRESULT hr = multi->Perform();
        if (FAILED(hr))
        {
            HC_TRACE_ERROR_HR(HTTPCLIENT, hr, "CurlMulti::Perform failed. Failing all active requests.");
            multi->FailAllRequests(hr);
        }
    }
    else
    {
        multi->FailAllRequests(E_ABORT);
    }

    std::unique_lock<std::mutex> lock{ multi->m_mutex };
    if (--multi->m_taskQueueCallbacksPending == 0 && multi->m_cleanupAsyncBlock)
    {
        // If CurlMulti::CleanupAsync was called and there are no remaining task queue callbacks, schedule cleanup now.
        // We *MUST* schedule the cleanup outside of holding the lock though. Scheduling cleanup may free the CurlMulti 
        // object memory before we're done using it here and we don't want that to happen while we're holding the lock 
        // or still otherwise referencing the CurlMulti object memory
        scheduledCleanupBlock = multi->m_cleanupAsyncBlock;
    }
}

if (scheduledCleanupBlock)
{
    // Must not reference CurlMulti object memory or hold CurlMulti object lock when scheduling this work.
    HRESULT hr = XAsyncSchedule(scheduledCleanupBlock, 0);
    if (FAILED(hr))
    {
        HC_TRACE_ERROR_HR(HTTPCLIENT, hr, "CurlMulti::TaskQueueCallback: Failed to schedule CleanupAsyncProvider!");
        assert(false);
    }
}
```

the downside of explicit unlocks is someone can always come along and write some buggy code that doesn't factor in the above statement that the lock has been released OR that it's generally unsafe to reference CurlMulti memory aftet XAsyncSchedule has been called.

the scoping makes it easier to force this xasyncschedule to be the very last thing we do in this function

ScottMunroMS (approved):
LGTM just left some feedback on ways I think can make the code just a little bit nicer and less likely to stray due to programmer error in the future.

feel free to disregard as codebase owners. I defer to your stylistic/convention preferences here

jplafonta:
>         HC_UNIQUE_PTR<CurlMulti> reclaim{ static_cast<CurlMulti*>(multi) };

sure, can remove that.

---
In reply to: [2638174455](https://github.com/microsoft/libHttpClient/pull/857#issuecomment-2638174455) [](http://example.com/codeflow?ancestors=2638174455)

---
Refers to: Source/HTTP/Curl/CurlMulti.cpp:119 in f848f72. [](commit_id = f848f72af32fb69bd0dfefd2954299ac8b4a2f16, deletion_comment = False)

Source/HTTP/Curl/CurlMulti.cpp
jplafonta: Reworded slightly but I made the comment more detailed.

Source/HTTP/Curl/CurlMulti.cpp
jplafonta: added.

Source/HTTP/Curl/CurlMulti.cpp
jplafonta: thats good feedback, I will update that.

ScottMunroMS (approved):
LGTM!

PR #860, opened 2025-03-03
Add websocketpp loggers

dawesten-msft:
This adds alevel and elevel loggers to the websocketpp implementation so those messages go through the normal tracing functions provided by clients. The alevel is disabled when the library is built without verbose logging support and the elevel logger when built without any tracing


dawesten-msft:
Closing in favor of #864

PR #862, opened 2025-03-07
address compiler warnings

jasonsandlin:
* Fixes unknown-pragma warnings for non-MSC compilers in `LocklessQueue.h`.
* Suppress 64-to-32-bit shortening compiler warning for newer versions of MSC (warning 4244) in `websocketpp_websocket.cpp`.

from https://github.com/microsoft/libHttpClient/pull/861


jasonsandlin:
just to enable PR gated build of #861

PR #864, opened 2025-03-08
Add websocketpp loggers

dawesten-msft:
This adds alevel and elevel loggers to the websocketpp implementation so those messages go through the normal tracing functions provided by clients. The alevel is disabled when the library is built without verbose logging support and the elevel logger when built without any tracing


jasonsandlin:
Note: this isn't going into main to avoid conflicts if they go to main the wrong order

jplafonta (approved):

jasonsandlin (approved):

PR #867, opened 2025-03-17
Add HCWebSocketSetPingInterval

dawesten-msft:
Not all platforms behave similarly or have similar capabilities but this is a best effort to at least expose some control over the ping behavior by adding HCWebSocketSetPingInterval and HCWebSocketGetPingInterval. If the client did not specify a ping interval the library behaves as before to avoid unexpected changes.

If the client does, Android and websocketpp now have a pong-timeout to help detect broken connections. Unfortunately, WinRT does not have any options for ping/pong frames and WinHttp can (as far as I could tell) only send unsolicited pongs.


Build/libHttpClient.GDK/libHttpClient.GDK.def
```diff
@@ -108,12 +108,14 @@ EXPORTS
     HCWebSocketGetHeader
     HCWebSocketGetHeaderAtIndex
     HCWebSocketGetNumHeaders
+    HCWebSocketGetPingInterval
```
jplafonta, line 111: >HCWebSocketGetPingInterval [](http://example.com/codeflow?start=4&length=26)

For both def files, add new methods to the end.  This maintains backwards compatibility for users linking with an old import lib.

dawesten-msft: @jplafonta done, please take another look. Thanks!

PR #877, opened 2025-04-21
Fix local reference table overflow for Android 32bit device

yuehanlin:
I am seeing the program crash at `JNI ERROR (app bug): local reference table overflow (max=512)` when running a bunch of tests on Arm7 Android devices, and I don't see this crash on 64 bit devices.

Copilot tells me it's because "On 32-bit Android, the local reference table size is limited to 512 entries by default, On 64-bit Android, the limit is much higher (sometimes ~5120)". I am not sure the architecture difference behind here, but DeleteLocalRef is the right fix after seeing this crash https://stackoverflow.com/questions/26685491/jni-error-app-bug-local-reference-table-overflow-max-512.

Verification: I don't see crash any more on x64 and Arm7

### DIFF
--- a/Source/WebSocket/Android/AndroidWebSocketProvider.cpp
+++ b/Source/WebSocket/Android/AndroidWebSocketProvider.cpp
@@ -85,10 +85,14 @@ struct HttpClientWebSocket
         const jstring headerValue = env->NewStringUTF(value);
         if (HadException(env) || !headerValue)
         {
+            env->DeleteLocalRef(headerName);
             return E_UNEXPECTED;
         }
 
         env->CallVoidMethod(m_webSocket, m_addHeader, headerName, headerValue);
+        env->DeleteLocalRef(headerName);
+        env->DeleteLocalRef(headerValue);
+
         if (HadException(env))
         {
             return E_UNEXPECTED;
@@ -119,10 +123,14 @@ struct HttpClientWebSocket
         const jstring javaSubProtocol = env->NewStringUTF(subProtocol.c_str());
         if (HadException(env) || !javaSubProtocol)
         {
+            env->DeleteLocalRef(javaUri);
             return E_UNEXPECTED;
         }
 
         env->CallVoidMethod(m_webSocket, m_connect, javaUri, javaSubProtocol);
+        env->DeleteLocalRef(javaUri);
+        env->DeleteLocalRef(javaSubProtocol);
+
         if (HadException(env))
         {
             return E_UNEXPECTED;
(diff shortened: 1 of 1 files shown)


jplafonta (commented):
I assume you've also validated that this doesn't regress WebSocket functionality in addition to alleviating the crash?

jplafonta (approved):
:shipit:

yuehanlin:
> I assume you've also validated that this doesn't regress WebSocket functionality in addition to alleviating the crash?

I've verified the Multiplayer tool test suites on x86 and arm7 after this fix, all test passed and no strange things happen

Source/WebSocket/Android/AndroidWebSocketProvider.cpp
```diff
@@ -85,10 +85,14 @@ struct HttpClientWebSocket
         const jstring headerValue = env->NewStringUTF(value);
         if (HadException(env) || !headerValue)
         {
+            env->DeleteLocalRef(headerName);
```
nassosterz-ms, line 88: This should be headerValue, right?

nassosterz-ms: Also line 82, should you delete local reference to headerName?

Source/WebSocket/Android/AndroidWebSocketProvider.cpp
```diff
@@ -119,10 +123,14 @@ struct HttpClientWebSocket
         const jstring javaSubProtocol = env->NewStringUTF(subProtocol.c_str());
         if (HadException(env) || !javaSubProtocol)
         {
+            env->DeleteLocalRef(javaUri);
```
nassosterz-ms, line 126: Line 120, delete javauri?

nassosterz-ms: Also, here, delete javaSubProtocol?

Source/WebSocket/Android/AndroidWebSocketProvider.cpp
```diff
@@ -149,6 +157,7 @@ struct HttpClientWebSocket
         {
             return E_UNEXPECTED;
         }
+        env->DeleteLocalRef(javaMessage);
```
nassosterz-ms, line 160: line 158, delete javaMessage?

Source/WebSocket/Android/AndroidWebSocketProvider.cpp
```diff
@@ -188,6 +197,7 @@ struct HttpClientWebSocket
         {
             return E_UNEXPECTED;
         }
+        env->DeleteLocalRef(buffer);
```
nassosterz-ms, line 200: Line 192, delete buffer?

Source/WebSocket/Android/AndroidWebSocketProvider.cpp
yuehanlin: yes I resolved all comments in latest iteration, I overthink this in the first iteration

nassosterz-ms (approved):

PR #879, opened 2025-04-24
Change spinlock implementation in TaskQueue to prevent CPU starvation

brianpepin:
The spinlocks in TaskQueue use a trivial implementation that can cause CPU starvation on single core machines. This change inserts processor hints that a spinlock is happening and implements backoff and a yield to the CPU if the spinlock spins too much.

### DIFF
--- a/Source/Task/SpinLock.h
+++ b/Source/Task/SpinLock.h
@@ -3,6 +3,12 @@
 #include <atomic>
 #include <thread>
 
+#if defined(_WIN32) || defined(__WINDOWS__) 
+#include <intrin.h>
+#elif (defined(_M_IX86) || defined(_M_X64))
+#include <x86intrin.h>
+#endif
+
 //
 // SpinLock: A spinlock implementation based on std::atomic_flag that
 // prevents CPU starvation. SpinLock can be used as a RAII wrapper around
--- a/Source/Task/SpinLock.h
+++ b/Source/Task/SpinLock.h
@@ -1,0 +1,66 @@
+#pragma once
+#include <algorithm>
+#include <atomic>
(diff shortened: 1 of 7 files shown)


jasonsandlin (approved):

Source/Task/SpinLock.h
```diff
@@ -0,0 +1,66 @@
...
+    std::atomic_flag& m_lock;
+
+    static inline void cpu_pause()
+    {
+#if defined(_M_IX86) || defined(_M_X64)
+        // x86/x64: Tells CPU we're spinning
+        // Reduces energy consumption
+        // Helps avoid memory order violations
+        // Maps to PAUSE instruction
+        _mm_pause();
```
PushpadantK, line 57: >_mm_pause(); [](http://example.com/codeflow?start=8&length=12)

One compiler issue I found is that this particular API is not supported on NDA console platform.

PushpadantK: need to add :
#if defined(_M_IX86) || defined(_M_X64)
#include <x86intrin.h>
#endif

brianpepin: Is there another #ifdef we can check here? x86intrin.h doesn't exist on windows platforms.  In Windows OS code I see this pattern.  Will this work for the NDA platform?

#if defined( _MSC_VER )
    #include <intrin.h>
#elif defined( __GNUC__ )
    #include <x86intrin.h>
#endif

brianpepin: Worked with Pushpadant offline and uploaded what we think will work for all platforms.

jasonsandlin (approved):

PR #880, opened 2025-05-07
Merging Gamesave feature into main

rgomez391:


Utilities/FrameworkResources/exports.exp
```diff
@@ -41,6 +41,7 @@ _HCHttpCallRequestGetRequestBodyReadFunction
 _HCHttpCallResponseSetResponseBodyWriteFunction
 _HCHttpCallResponseGetResponseBodyWriteFunction
 _HCHttpCallResponseSetGzipCompressed
+_HCHttpCallRequestSetProgressReportFunction
```
jasonsandlin, line 44: the two .exp files should have the same edits as the as the .def files in the \build folder

jasonsandlin (approved):

Utilities/FrameworkResources/exports.exp
rgomez391: that's right, otherwise we may see issues in iOS

rgomez391: Alright, I just added the missing calls

jasonsandlin (approved):

PR #881, opened 2025-05-12
Check undefined macros

tculotta:
Adds defined checks but keeps original conditions for consistency

### DIFF
--- a/Include/httpClient/httpClient.h
+++ b/Include/httpClient/httpClient.h
@@ -868,7 +868,7 @@ STDAPI HCHttpCallResponseGetHeaderAtIndex(
     _Outptr_result_z_ const char** headerValue
     ) noexcept;
 
-#if !HC_NOWEBSOCKETS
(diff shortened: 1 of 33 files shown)


Include/httpClient/httpClient.h
```diff
@@ -868,7 +868,7 @@ STDAPI HCHttpCallResponseGetHeaderAtIndex(
     _Outptr_result_z_ const char** headerValue
     ) noexcept;
 
-#if !HC_NOWEBSOCKETS
```
jplafonta, line 871: >HC_NOWEBSOCKETS [](http://example.com/codeflow?start=5&length=15)

is this not used elsewhere in LHC too? or do we just care about the public headers?

tculotta: Made the update everywhere. During the global search, I also found that we specifically doc this macro to be set as a build flag, so changing to ifdef and ifndef should be safe after all

jasonsandlin (approved):

PR #895, opened 2025-08-02
Adding support for HCHttpCallRequestSetMaxReceiveBufferSize and fixing previously broken unit tests

jasonsandlin:


ScottMunroMS (changes requested):
the change looks mostly good as-is. as mentioned out of band, the winhttp provider isn't yet supporting this new API correctly. We'll want that winhttp provider working in the near future but its not a strict blocker for this PR. 

marked as "requesting changes" since I had a few comments and curious to see where the resolution lands on addressing the winhttp behavior but otherwise LGTM

Tests/UnitTests/Tests/BufferSize/BufferSizeTests.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    // Create HTTP call
+    HCCallHandle call;
+    hr = HCHttpCallCreate(&call);
+    assert(SUCCEEDED(hr));
+
+    // Test setting and getting buffer size
+    const size_t testBufferSize = 64 * 1024; // 64KB
+    
+    // Set buffer size
```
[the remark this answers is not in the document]
jasonsandlin: Good catch on the API documentation. The current behavior is that GetMaxReceiveBufferSize returns 0 when 
no explicit size has been set, which means 'use provider default'. I should update the API documentation to 
clearly state this behavior. The test does validate this at the end, but it would be clearer to 
test it first as the baseline behavior.

Tests/UnitTests/Tests/BufferSize/BufferSizeTests.cpp
```diff
@@ -0,0 +1,63 @@
...
+    assert(SUCCEEDED(hr));
+    assert(retrievedBufferSize == testBufferSize);
+    printf("✓ Successfully retrieved buffer size: %zu bytes\n", retrievedBufferSize);
+
+    // Test invalid parameters
+    hr = HCHttpCallRequestSetMaxReceiveBufferSize(nullptr, testBufferSize);
+    assert(hr == E_INVALIDARG);
+    printf("✓ Correctly rejected null call handle\n");
+
+    hr = HCHttpCallRequestSetMaxReceiveBufferSize(call, 0);
```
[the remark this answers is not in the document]
jasonsandlin: That's a great suggestion for API symmetry. 
Currently we reject 0 as invalid, but allowing SetMaxReceiveBufferSize(call, 0) as a way to explicitly 
reset to provider default would be much cleaner. This would make the API more intuitive - you could 
set a custom size, then later reset to default if needed. I'll implement this change.

Source/HTTP/WinHttp/winhttp_connection.cpp
```diff
@@ -1045,13 +1030,37 @@ void WinHttpConnection::callback_status_data_available(
...
+            {
+                // Use the larger of the two, but cap at remaining bytes to read
+                size_t remainingBytes = pRequestContext->m_responseBodyRemainingToRead;
+                bytesToRead = static_cast<DWORD>(std::min({
+                    static_cast<size_t>(maxReceiveBufferSize),
+                    remainingBytes,
+                    static_cast<size_t>(MAXDWORD)
+                }));
+                
+                HC_TRACE_INFORMATION(HTTPCLIENT, "WinHttpConnection [ID %llu] Using maxReceiveBufferSize=%zu for known content-length, reading %d bytes",
```
ScottMunroMS, line 1052: I don't really have a strong opinion but is this really an INFORMATION level trace? this seems really verbose?

Source/HTTP/WinHttp/winhttp_connection.cpp
```diff
@@ -1045,13 +1030,37 @@ void WinHttpConnection::callback_status_data_available(
...
+        if (pRequestContext->m_responseBodySize > 0)
+        {
+            size_t maxReceiveBufferSize = 0;
+            HRESULT hr = HCHttpCallRequestGetMaxReceiveBufferSize(pRequestContext->m_call, &maxReceiveBufferSize);
+            if (SUCCEEDED(hr) && maxReceiveBufferSize > 0)
+            {
+                // Use the larger of the two, but cap at remaining bytes to read
+                size_t remainingBytes = pRequestContext->m_responseBodyRemainingToRead;
+                bytesToRead = static_cast<DWORD>(std::min({
+                    static_cast<size_t>(maxReceiveBufferSize),
```
ScottMunroMS, line 1047: is this cast necessary? maxReceiveBufferSize is already a size_t

Source/HTTP/WinHttp/winhttp_connection.cpp
```diff
@@ -1045,13 +1030,37 @@ void WinHttpConnection::callback_status_data_available(
     // Read new data into buffer
     if (newBytesAvailable > 0)
     {
-        pRequestContext->m_responseBuffer.resize(newBytesAvailable);
+        // Stage 1 performance optimization: For requests with known Content-Length,
```
ScottMunroMS, line 1033: nit: "Stage 1" feels like a weird comment given I don't think I see any other "stages" of optimization here?

Source/HTTP/WinHttp/winhttp_connection.cpp
jasonsandlin: Right, fixed

jasonsandlin: Right, fixed

Source/HTTP/WinHttp/winhttp_connection.cpp
jasonsandlin: Right, removed

ScottMunroMS (approved):
a few more comments but generally LGTM

ScottMunroMS (approved):
LGTM

jplafonta (approved):

PR #898, opened 2025-08-08
Making XCurl be delay-loaded

jasonsandlin:
Loading XCurl.dll at runtime (using new CurlDynamicLoader class)

### DIFF
--- a/Source/HTTP/Curl/CurlEasyRequest.cpp
+++ b/Source/HTTP/Curl/CurlEasyRequest.cpp
@@ -183,7 +183,7 @@ void CurlEasyRequest::Complete(CURLcode result)
         HC_TRACE_INFORMATION(HTTPCLIENT, "CurlEasyRequest::m_errorBuffer='%s'", m_errorBuffer);
 
         long platformError = 0;
-    auto curle = CURL_CALL(curl_easy_getinfo)(m_curlEasyHandle, CURLINFO_OS_ERRNO, &platformError);
+        auto curle = CURL_CALL(curl_easy_getinfo)(m_curlEasyHandle, CURLINFO_OS_ERRNO, &platformError);
         if (curle != CURLE_OK)
         {
             return Fail(HrFromCurle(curle));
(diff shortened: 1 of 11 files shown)


Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -18,13 +18,32 @@ CurlEasyRequest::CurlEasyRequest(CURL* curlEasyHandle, HCCallHandle hcCall, XAsy
 
 CurlEasyRequest::~CurlEasyRequest()
 {
+#if HC_PLATFORM == HC_PLATFORM_GDK
+    if (CurlDynamicLoader::GetInstance().IsLoaded())
+    {
+        CURL_CALL(curl_easy_cleanup)(m_curlEasyHandle);
```
jplafonta, line 24: >CURL_CALL [](http://example.com/codeflow?start=8&length=9)

To make code more Xplat couldn't you just define CURL_CALL to be basically a no-op on non GDK platforms? you'd need many less if-defs throughout the code then

jplafonta: You could also add something like

CONDITIONAL_CURL_CALL(...)

that first validates the library is loaded on GDK and does nothing on non-gdk

Source/HTTP/Curl/CurlEasyRequest.h
```diff
@@ -1,13 +1,15 @@
 #pragma once
-
 #if HC_PLATFORM == HC_PLATFORM_GDK
 // When developing titles for Xbox consoles, you must use WinHTTP or xCurl.
 // See https://docs.microsoft.com/en-us/gaming/gdk/_content/gc/networking/overviews/web-requests/http-networking for detail
 #include <XCurl.h>
+#include "CurlDynamicLoader.h"
+#define CURL_CALL(func_name) CurlDynamicLoader::GetInstance().func_name##_fn
```
jplafonta, line 7: >CURL_CALL [](http://example.com/codeflow?start=8&length=9)

I'd define these in CurlDynamicLoaded for clarity

jplafonta (changes requested):
:clock1:

Source/HTTP/Curl/CurlEasyRequest.cpp
```diff
@@ -175,7 +183,7 @@ void CurlEasyRequest::Complete(CURLcode result)
         HC_TRACE_INFORMATION(HTTPCLIENT, "CurlEasyRequest::m_errorBuffer='%s'", m_errorBuffer);
 
         long platformError = 0;
-        auto curle = curl_easy_getinfo(m_curlEasyHandle, CURLINFO_OS_ERRNO, &platformError);
+    auto curle = CURL_CALL(curl_easy_getinfo)(m_curlEasyHandle, CURLINFO_OS_ERRNO, &platformError);
```
jplafonta, line 186: >     [](http://example.com/codeflow?start=0&length=4)

nit: tabbing looks off here

jplafonta (dismissed):
:shipit:

jplafonta (approved):
:shipit:

PR #907, opened 2025-09-03
Changing WinHTTP flags: WINHTTP_FLAG_SECURE_DEFAULTS for HTTPS, WINHTTP_FLAG_ASYNC for HTTP

jasonsandlin:


Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -342,31 +358,40 @@ Result<HINTERNET> WinHttpProvider::GetHSession(uint32_t securityProtocolFlags)
...
+    DWORD openFlags;
+    if (isHttps)
+    {
+        // For HTTPS, use secure defaults which implies WINHTTP_FLAG_ASYNC
+        openFlags = WINHTTP_FLAG_SECURE_DEFAULTS;
+    }
+    else
+    {
+        // For HTTP, use async only (allow insecure connections)
+        openFlags = WINHTTP_FLAG_ASYNC;
```
ScottMunroMS, line 372: If I understood correctly, HTTP being broken was something you found incidentally during your own out-of-band testing. Should you add automated tests to make sure HTTP continues to work for all HTTP-supporting platforms?

Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -326,8 +327,23 @@ Result<XPlatSecurityInformation> WinHttpProvider::GetSecurityInformation(const c
...
+    {
+        return E_INVALIDARG;
+    }
+    
+    bool isHttps = uri.IsSecure();
+    
+    // Log warning for insecure HTTP requests
+    if (!isHttps)
+    {
+        HC_TRACE_WARNING(HTTPCLIENT, "WARNING: Insecure HTTP request \"%s\"", url);
```
ScottMunroMS, line 344: For XCurl we had the policy to warn about HTTP use for historical and console certification reasons. I think it's fine to say that LibHttpClient does not have the same policy and where not strictly required (i.e. most platforms) you won't generate noise about HTTP usage.

I'd recommend either removing this *or* putting it behind some GDK ifdef *or* some GDK-Console runtime flag

Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -375,20 +400,26 @@ Result<HINTERNET> WinHttpProvider::GetHSession(uint32_t securityProtocolFlags)
...
-        return hr;
+        auto result = WinHttpSetOption(
+            hSession,
+            WINHTTP_OPTION_SECURE_PROTOCOLS,
+            &securityProtocolFlags,
+            sizeof(securityProtocolFlags));
+        if (!result)
+        {
+            HRESULT hr = HRESULT_FROM_WIN32(GetLastError());
+            HC_TRACE_ERROR_HR(HTTPCLIENT, hr, "WinHttpProvider WinHttpSetOption");
```
ScottMunroMS, line 415: not strictly related to this PR but it seems like this error could be clearer about the optinon it failed to set?

ScottMunroMS (approved):
Approve w/ suggestions

Source/HTTP/WinHttp/winhttp_provider.cpp
jasonsandlin: fixed, test added

Source/HTTP/WinHttp/winhttp_provider.cpp
jasonsandlin: yes, fixed.

Tests/UnitTests/Tests/HttpTests.cpp
```diff
@@ -449,6 +453,71 @@ DEFINE_TEST_CLASS(HttpTests)
...
+        // being blocked by WinHTTP flag configuration issues.
+        
+        VERIFY_ARE_EQUAL(S_OK, HCInitialize(nullptr));
+
+        HCCallHandle call = nullptr;
+        VERIFY_ARE_EQUAL(S_OK, HCHttpCallCreate(&call));
+
+        // Use an HTTP URL (not HTTPS) to test the WinHTTP flags behavior
+        // Using a local/private IP that won't resolve to test flag configuration without external dependency
+        VERIFY_ARE_EQUAL(S_OK, HCHttpCallRequestSetUrl(call, "GET", "http://192.168.1.255/test"));
```
ScottMunroMS, line 478: don't you have other tests hitting example.com? would that be a way to hit a real http test endpoint and verify SUCCESS rather than hitting a failure and then trying to determine it wasn't an unexpected failure?

Tests/UnitTests/Tests/HttpTests.cpp
```diff
@@ -449,6 +453,71 @@ DEFINE_TEST_CLASS(HttpTests)
         HCHttpCallCloseHandle(call);
         HCCleanup();
     }
+
+    // Temporarily enabled for testing WinHTTP flag changes
```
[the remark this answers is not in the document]
jasonsandlin: fixed

Tests/UnitTests/Tests/HttpTests.cpp
```diff
@@ -454,69 +454,52 @@ DEFINE_TEST_CLASS(HttpTests)
...
-        // Use an HTTP URL (not HTTPS) to test the WinHTTP flags behavior
-        // Using a local/private IP that won't resolve to test flag configuration without external dependency
-        VERIFY_ARE_EQUAL(S_OK, HCHttpCallRequestSetUrl(call, "GET", "http://192.168.1.255/test"));
+        // Use HTTP protocol to verify it works correctly
+        VERIFY_ARE_EQUAL(S_OK, HCHttpCallRequestSetUrl(call, "GET", "http://example.com"));
+        VERIFY_ARE_EQUAL(S_OK, HCHttpCallRequestSetRetryAllowed(call, false));
         
-        // Set a short timeout since we expect this to fail with network error, not WinHTTP flag error
-        VERIFY_ARE_EQUAL(S_OK, HCHttpCallRequestSetTimeout(call, 1));
+        // Create a mock response for the HTTP call
```
ScottMunroMS, line 475: I'm not familiar with the HC Mocking stuff. if you use a mock is it still actually testing that WinHTTP is allowing HTTP protocol usage? I think part of my example.com feedback was that you could make a real http request not a mock one

jasonsandlin: unit tests are all against mocks and don't make real calls so I modified the GDK sample to verify its working e2e

ScottMunroMS (approved):
lgtm

PR #908, opened 2025-09-09
Ignore unknown pragmas on apple

btreadway-blizzard:
These two pragmas are unknown in XCode. While xcode can normally handle this and treat it as only a warning, systems that build with warnings as errors will fail to build. Add a check to ignore these two pragmas if the platform is Apple (macOS or iOS)

### DIFF
--- a/Include/httpClient/pal.h
+++ b/Include/httpClient/pal.h
@@ -5,7 +5,7 @@
 #endif
 
 #pragma once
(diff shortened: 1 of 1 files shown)


SebastianPD (approved):

btreadway-blizzard:
Going to close and re-create in main repo and not a branch per request of Jason

Issue #932, opened 2026-02-18
Flaky async/task-queue unit tests due to timing assumptions and use-after-free bugs


jhugard:
### Description
Multiple unit tests in the async and task-queue test suites exhibit intermittent failures due to:
1. Overly strict timing assumptions that don't account for thread pool scheduling variance
2. Race conditions between async operations completing and test verification
3. Use-after-free bugs in test fixtures that pass invalid/freed handles

### Symptoms
- **VerifyDistributedAsyncCall**: Fails intermittently when timing assumptions for iteration delays are not met due to system load
- **VerifyWaitForCompletion, VerifySimpleAsyncCall**: Race conditions where verification occurs before async callbacks complete
- **VerifyDuplicateQueueHandle**: Use-after-free when testing failure case with closed queue handle

### Root Causes

1. **Timing races**: Tests check operation counts/opcodes without waiting for all operations to complete
   - Threadpool scheduling can delay callback execution
   - Completion port drains happen asynchronously
   
2. **Tight timing windows**: Tests assume specific timing relationships (e.g., 100ms per iteration) without accounting for system variance

3. **Factorial worker lifetime**: Per-iteration callbacks can execute after the test verification phase, causing use-after-free of the work data

4. **Invalid handles in error path testing**: VerifyDuplicateQueueHandle tests error handling by passing a freed handle, which gets dereferenced

### Solution
- Add explicit wait loops before verification to ensure all async operations complete
- Drain task queues before verifying operation counts
- Hold per-call references in FactorialWorker to prevent UAF during async callbacks
- Replace freed handle with stack-allocated fake handle for error case testing

### Impact
- Unit tests become stable and reliable across varying system loads
- Reduces false negatives in CI/CD pipelines
- No impact on non-test code

PR #935, opened 2026-02-18
Stabilize unit tests (#932, #938)

jhugard:
## Problem (#932, #938)
Unit tests for async operations and task queues exhibit intermittent failures on win32 target (others untested) due to timing assumptions, races, and use-after-free in test fixtures (#932). In addition, `AsyncBlockTests::VerifyAsyncBlockReuse` can crash under stress when concurrent async lifecycle phases append to a shared `std::vector` (#938).

**Key failure modes:**
- **#932 UAF:** `VerifyDuplicateQueueHandle` tests an error path by passing a freed queue handle, which can be dereferenced in the duplicate-handle failure path.
- **#938 race:** cleanup of the first async call can run concurrently with begin of the second call, leading to simultaneous `opCodes.push_back()` and heap corruption during vector reallocation.

### Stabilize async/task-queue unit tests and harden factorial worker lifetime (#932)
- **Explicit synchronization**: Add wait loops in `VerifySimpleAsyncCall`, `VerifyWaitForCompletion`, and `VerifyDistributedAsyncCall` to ensure async operations complete before verification.
- **Queue drains**: Wait for both Work and Completion ports to be empty before verifying final state.
- **Relaxed timing**: Allow per-iteration slack window for `VerifyDistributedAsyncCall` timing check (5 iterations x 100ms with one-interval slack).
- **Lifetime hardening**: Hold per-call references in `FactorialWorkerSimple` and `FactorialWorkerDistributed` to prevent use-after-free during async callbacks.
- **Conditional waits**: In `_VerifyQueueTermination`, only wait for counts to settle for non-blocking termination cases.

### Fix use-after-free in test VerifyDuplicateQueueHandle (#932)
- Replace freed queue handle with stack-allocated fake handle (64-byte aligned buffer) for error testing.
- Prevents dereferencing of previously-freed memory in the `XTaskQueueDuplicateHandle` failure path.

### Fix data race in AsyncBlockTests opcode logging (#938)
- Replace `std::vector<XAsyncOp>` with a fixed-capacity lock-free log: `std::array<std::atomic<XAsyncOp>, 16>` + `std::atomic<size_t>`.
- Use `store(memory_order_release)` for writes and `load(memory_order_acquire)` for snapshots to ensure visibility on weakly-ordered architectures (ARM).
- Provide `RecordOp()` for lock-free append and `GetOpCodes()` for snapshots.
- Snapshot optimization for repeated verification checks.

## Impact
- Eliminates timing-dependent test failures and data races (#932, #938).
- Prevents use-after-free in test fixtures.
- Improves test reliability on high-load systems and CI.
- No impact on product code or API.

## Testing
- `VerifyAsyncBlockReuse`: 10/10 passes.
- AsyncBlockTests suite: 23/23 passes.
- Note: The original data-race crash required a long soak with page heap; extended stress testing is recommended for maximum confidence.

### DIFF
--- a/Tests/UnitTests/Tests/AsyncBlockTests.cpp
+++ b/Tests/UnitTests/Tests/AsyncBlockTests.cpp
@@ -7,6 +7,7 @@
 #include "XAsyncProviderPriv.h"
 #include "XTaskQueue.h"
 #include "XTaskQueuePriv.h"
+#include <array>
 
 #define TEST_CLASS_OWNER L"brianpe"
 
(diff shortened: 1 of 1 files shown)


jhugard:
Sneaking in another change to relax a time-based test, found while soaking under cdb with page-heap.  I've found that a loaded system sometimes needs >50ms epsilon to service delayed tasks.

jasonsandlin (approved):

Issue #936, opened 2026-02-18
Heap Corruption in Concurrent XTaskQueue Termination


jhugard:
## Summary

Concurrent termination operations on XTaskQueue cause heap corruption and access violations due to unsynchronized access to internal termination tracking lists.

## Environment

- **Platform**: Windows x64
- **Build**: Debug with Page Heap enabled
- **Test**: VerifyCompositeTerminationRaceRepro (64 threads, 100 iterations)

## Symptoms

- **Crash Type**: Access violation (0xc0000005) 
- **Location**: `AtomicVector<ITaskQueuePortContext*>::Remove` → corrupted vector iterator
- **Failure Bucket**: `INVALID_POINTER_WRITE_c0000005_ntdll.dll!RtlpWaitOnCriticalSection`
- **Trigger**: High-concurrency composite queue termination stress test

### Race Condition #1: Nested Terminate During SignalTerminations

**Scenario**: Thread A iterates `m_terminationList` in `SignalTerminations()`, invokes a termination callback, which calls `XTaskQueueTerminate()` on another queue sharing the same port, causing concurrent modification of the list being iterated.

**Code Path**:
```
SignalTerminations()
  → m_terminationList->remove_if([...])
    → entry->callback(callbackContext)
      → User callback calls XTaskQueueTerminate()
        → ScheduleTermination()
          → m_terminationList->push_back()  // RACE: concurrent modification
```

### Race Condition #2: Concurrent ScheduleTermination

**Scenario**: Multiple threads call `ScheduleTermination()` simultaneously (e.g., multiple delegate queues terminating concurrently), causing heap corruption in the lockless queue's push_back operation.

**Code Path**:
```
Thread A: ScheduleTermination() → m_terminationList->push_back(...)
Thread B: ScheduleTermination() → m_terminationList->push_back(...)
// RACE: concurrent push_back corrupts internal vector state
```

### Minimal Reproduction

```cpp
// Create root queue and get ports
XTaskQueueHandle root;
XTaskQueueCreate(XTaskQueueDispatchMode::ThreadPool,
                 XTaskQueueDispatchMode::ThreadPool, &root);
XTaskQueuePortHandle workPort, completionPort;
XTaskQueueGetPort(root, XTaskQueuePort::Work, &workPort);
XTaskQueueGetPort(root, XTaskQueuePort::Completion, &completionPort);

// 64 threads, each creating 100 delegate queues
// Terminate all concurrently → triggers heap corruption
[...]

PR #937, opened 2026-02-18
Fix Concurrent Termination Race Conditions in XTaskQueue #936

jhugard:
## Overview

This PR fixes critical heap corruption issues when multiple threads concurrently terminate task queues or invoke nested `XTaskQueueTerminate()` calls during termination callback processing.

## Problem Statement (#936)

The termination list (`m_terminationList`) and pending termination list (`m_pendingTerminationList`) were accessed concurrently without proper synchronization, leading to two race conditions that manifest as heap corruption and access violations under high concurrency.

### Race #1: Nested Terminate During Iteration

When `SignalTerminations()` iterates the termination list and invokes callbacks, those callbacks may call `XTaskQueueTerminate()` on other queues, causing concurrent modification of the list being iterated.

### Race #2: Concurrent ScheduleTermination  

Multiple threads calling `ScheduleTermination()` simultaneously corrupt the lockless queue's internal state during `push_back()`.

### Core Fix

- **Added `m_terminationLock`**: New mutex in `TaskQueuePortImpl` to serialize all termination list operations
- **Refactored `SignalTerminations()`**: Collects entries under lock, processes callbacks outside iteration to prevent nested modification races
- **Added UAF protection**: `AddRef`/`Release` around termination callbacks to prevent use-after-free if callback releases the queue

### Protected Operations  

All termination list accesses now use `std::lock_guard<std::mutex>`:
- `PrepareTermination` - lock when reserving nodes
- `CancelTermination` - lock when freeing nodes  
- `Terminate` - lock when adding to pending list
- `ScheduleTermination` - lock when pushing to active list
- `ResumeTermination` - lock during removal, schedule entries outside lock
- `ResumePort` - lock when manipulating termination list
- `TerminationListEmpty()` - new helper for thread-safe empty checks

### Documentation Enhancement

Enhanced `XTaskQueueTerminate()` API documentation to clarify:
- Queues terminate independently (composite queues don't auto-cascade)
- `wait=true` blocks only on **this** queue's termination, not delegates
- Proper pattern for coordinated multi-queue shutdown using termination callbacks

### Crash Reproduction ✅

- **Before Fix**: Stress test with page heap → access violation reproducible within 100 runs of the test
- **After Fix**: Clean execution, no heap corruption detected after 100 runs of the test

### Stack Trace (Before Fix)

```
ntdll!RtlpWaitOnCriticalSection+0xb3
  → AtomicVector<ITaskQueuePortContext*>::Remove
  → TaskQueuePortImpl::Detach  
  → TaskQueueImpl::RundownObject
```

### Test Results ✅

- **Full suite**: 81/81 tests passed  
- **Stress mode**: 256 threads, 100 iterations, no crashes
- **Page heap**: Clean (no corruption detected)

## Performance Considerations

The mutex adds minimal overhead as termination is an infrequent operation (once per queue lifecycle). The lock is held only during list manipulation, not during callback execution, preserving concurrency.

## Breaking Changes

None. This is a pure bug fix with no API changes.

## Related Issues

#936 Fixes heap corruption crash in high-concurrency composite queue termination scenarios.

### DIFF
--- a/Tests/UnitTests/Tests/TaskQueueTests.cpp
+++ b/Tests/UnitTests/Tests/TaskQueueTests.cpp
@@ -963,7 +963,7 @@ public:
 
         // Verify the default task queue can be terminated, but can come back if the
         // runtime is re-initialized.
-        VERIFY_SUCCEEDED(XTaskQueueTerminate(globalQueue, false, nullptr, nullptr));
+        VERIFY_SUCCEEDED(XTaskQueueTerminate(globalQueue, true, nullptr, nullptr));
         VERIFY_ARE_EQUAL(E_ABORT, XTaskQueueSubmitCallback(globalQueue, XTaskQueuePort::Work, nullptr, cbEmpty));
 
         globalQueue.Close();


brianpepin (approved):
I suspect this is a little over-constrained as the LocklessQueue class is the basis for m_terminationList and m_pendingTerminationList and is _supposed_ to always stay sane regardless of thread load.  However there are several places in the TaskQueue code that do perform multiple operations on the termination lists and is pretty hard to prove those do not need locks.  As you indicate, termination is a rare occurrence so locking should not affect performance here. I would rather be a bit over-constrained and have the code be clear that there are no threading issues.  Proving lock free code is safe can be very difficult especially for things like termination that involve coordination and callbacks. It is not worth it in a path that is not performance critical.  So approved. Thanks for the contribution.

PR #957, opened 2026-04-10
Protect HTTP cleanup callback handoff 

### LINKED ISSUE #956: Potential use-after-free during HTTP cleanup/completion overlap
## Summary

`HCCleanupAsync` can overlap completion of an in-flight HTTP request. If the HTTP completion callback frees its `XAsyncBlock`, cleanup can subsequently attempt to `XAsyncCancel` that same block, creating a potential use-after-free.  This hazard was introduced with #931.

## Impact

- Potential use-after-free on the client-owned `XAsyncBlock`
- Nondeterministic cleanup/completion ordering for HTTP performs
- Possible follow-on failures in later operations once the race is hit

## Expected behavior

Once the HTTP completion callback begins running, library cleanup should no longer treat that request as cancelable through the client-owned `XAsyncBlock`.

## Actual behavior

During cleanup, the request can still be observed as active and cancelable while the client completion callback is already executing.

## Repro

At a high level:

1. Initialize libHttpClient.
2. Create a mocked HTTP call and perform it asynchronously.
3. Let the completion callback run on the client-owned `XAsyncBlock`.
4. Observe that cleanup can still treat that request as cancelable from within the callback window.

## Note

A PR fix will be inbound after internal review.

jhugard:
Fixes #956

## Summary

This change tightens the handoff between HTTP cleanup and HTTP completion so cleanup no longer touches the client-owned `XAsyncBlock` after the completion path has claimed callback ownership.

## Problem

Cleanup and completion can overlap on the same HTTP perform. In that window, cleanup can still treat the request as cancelable through the client-owned `XAsyncBlock` even after the completion callback has started. The API documentation allows the callback to free that `XAsyncBlock`, which can turn a later `XAsyncCancel` into a use-after-free.

## What changed

- Track active HTTP performs by `HttpPerformContext` instead of raw client `XAsyncBlock` pointers.
- Add an explicit handoff state for the client-owned async block during cleanup and completion.
- Snapshot requests for cancellation under the `NetworkState` mutex and issue `XAsyncCancel` outside the lock.
- Keep the completion path from publishing callback ownership until cleanup has finished its cancel propagation for that request.

## Validation

- Added the repro regression in the first commit on this branch.
- Verified the repro state fails the Windows x64 Debug TAEF suite.
- Verified this fix restores the suite to green: `Summary: Total=83, Passed=83, Failed=0, Blocked=0, Not Run=0, Skipped=0`.

## Notes

This branch is intentionally split into two commits: the repro commit and the fix commit.

### DIFF
--- a/Source/Global/NetworkState.cpp
+++ b/Source/Global/NetworkState.cpp
@@ -68,6 +68,17 @@ Result<UniquePtr<HC_CALL>> NetworkState::HttpCallCreate() noexcept
     return call;
 }
 
+// Coordinates handoff of the client-owned XAsyncBlock between cleanup and completion.
+// This prevents CleanupAsyncProvider from canceling a request while HttpCallPerformComplete is
+// handing that same XAsyncBlock to XAsyncComplete, after which the client callback may delete it.
+// Once CallbackStarted is published, cleanup must never touch clientAsyncBlock again.
+enum class HttpPerformClientBlockState : uint8_t
+{
+    CleanupMayCancel,
+    CleanupCancelInFlight,
+    CallbackStarted
+};
+
 struct NetworkState::HttpPerformContext
 {
     HttpPerformContext(NetworkState& _state, HCCallHandle _callHandle, XAsyncBlock* _clientAsyncBlock) :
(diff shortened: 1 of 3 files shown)


jasonsandlin (approved):

jplafonta (approved):
:shipit:

PR #959, opened 2026-04-21
Use parallel make builds to speed up LibHttpClient.Linux build times

ScottMunroMS:
Fix some low hanging fruit to speed up linux builds.

Prior we were running all make compilations in sequence. Now make builds openssl, curl, and the core libhttpclient code with a parallelism setting equal to the number of cores on the machine doing the build.

Additionally, specify `no-tests` in the OpenSSL build to avoid building unused OpenSSL tests

In isolated libHttpClient linux builds I'm seeing an 80% reduction in clean build times (5 min -> ~45 secs) on native linux hosts and similar reductions on WSL/crossVM builds (10 min -> ~1:30m)


SebastianPD (approved):

ScottMunroMS:
Closing and replacing with a PR from the right topic branch
https://github.com/microsoft/libHttpClient/pull/960

PR #961, opened 2026-04-22
Handle spontaneous WebSocket disconnects during connect completion

PushpadantK:
Problem

There is a race in the WebSocket connect flow where the provider can complete the connect operation successfully, but the socket may be closed before ConnectComplete runs. In that case, the previous behavior could leave the socket in an inconsistent state and allow the public connect path to appear successful even though the connection had already been lost.

Fix

This change makes the connect/disconnect race deterministic in hcwebsocket.cpp.

CloseFunc now treats a close that arrives before m_providerContext is established as an early disconnect during connect rather than an unexpected callback to ignore. It marks the socket as Disconnected and returns, leaving final cleanup to ConnectComplete.

ConnectComplete now explicitly handles the case where the socket has already transitioned to Disconnected before the callback runs. If the provider-level connect result succeeded but the socket was already disconnected, the public connect result is converted to a failure. This keeps ConnectComplete as the single point that finalizes the connect outcome and state transition.

There is also a small cleanup in ConnectAsyncProvider so the websocket reference is resolved only inside the Begin path, after asserting the observer is valid.

Why This Approach

The important behavior here is not just detecting the race, but making one place responsible for the final result. By letting CloseFunc only record the early disconnect and letting ConnectComplete own the final success/failure decision, we avoid split cleanup logic and ensure callers do not observe a successful connect on a socket that has already closed.

Validation

I added a focused regression test, TestConnectFailsWhenDisconnectedDuringCompletion, in WebsocketTests.cpp.

The test uses a targeted test provider to simulate a disconnect during the connect-completion window and verifies that:

connect does not complete as a usable successful connection
retrieving the connect result fails
the socket cannot be used for subsequent send/disconnect operations
I also reran the focused WebSocket TE coverage locally:

TestGlobalCallbacks
TestCloseHandles
TestConnect
TestRequestHeaders
TestConnectFailsWhenDisconnectedDuringCompletion
Those tests passed in the local TE run.

### DIFF
--- a/Source/WebSocket/hcwebsocket.cpp
+++ b/Source/WebSocket/hcwebsocket.cpp
@@ -205,12 +205,13 @@ HRESULT WebSocket::ConnectAsync(
 HRESULT CALLBACK WebSocket::ConnectAsyncProvider(XAsyncOp op, XAsyncProviderData const* data)
 {
     ConnectContext* context{ static_cast<ConnectContext*>(data->context) };
-    auto& ws{ context->observer->websocket };
 
     switch (op)
     {
     case XAsyncOp::Begin:
     {
+        assert(context->observer);
+        auto& ws{ context->observer->websocket };
         std::unique_lock<std::mutex> lock{ ws->m_stateMutex };
 
         RETURN_HR_IF(E_UNEXPECTED, ws->m_state != State::Initial);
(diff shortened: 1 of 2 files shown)


bmazus (approved):
###

jasonsandlin (approved):

PR #963, opened 2026-04-28
LibHttpClient.Linux: refactor dependency install to optional step and eliminate forced sudo calls

ScottMunroMS:
LibHttpClient's linux build had `sudo` authorized calls sprinkled throughout its build by default to install dependencies. Generally, this is an anti-pattern. It means that to ever build libHttpClient for linux you need to have sudo privileges even if all required dependencies are installed. sudo should be isolated to and reserved for operations which require elevations. General builds should not require elevation.

Callers can continue to retain the old behavior by passing `--install-dependencies` to the old build script.

I've also eliminated some unnecessary sudo usage from the openssl build. it was primarily there to give write access to files in the global /usr/local folder. Instead of installing openssl build results there, install them to a local temp folder instead and the need for sudo goes away.


Utilities/Pipelines/Tasks/linux-build.yml
```diff
@@ -5,6 +5,11 @@ parameters:
 steps:
   - template: checkout.yml
 
+  - task: Bash@3
+    displayName: 'Ensure libHttpClient dependencies installed'
+    inputs:
+      filePath: './Build/libHttpClient.Linux/install_dependencies.bash'
```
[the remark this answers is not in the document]
ScottMunroMS: the intent of the PR is to make sudo optional for dev desk builds. the cloud builds need to continue to install dependencies.

Build/libHttpClient.Linux/install_dependencies.bash
```diff
@@ -0,0 +1,52 @@
...
+    -*|--*)
+      echo "Unknown option $1"
+      exit 1
+      ;;
+    *)
+      POSITIONAL_ARGS+=("$1") # save positional arg
+      shift # past argument
+      ;;
+  esac
+done
```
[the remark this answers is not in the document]
ScottMunroMS: this matches precedent in other scripts which also don't use positional args really. going to leave as-is

Build/libHttpClient.Linux/install_dependencies.bash
```diff
@@ -0,0 +1,52 @@
...
+  exit 0
+fi
+
+echo "Installing dependencies..."
+echo "Build dependencies: ${build_dependencies[*]}"
+echo "Library dependencies: ${library_dependencies[*]}"
+
+sudo hwclock --hctosys
+sudo apt-get update
+sudo apt-get install "${build_dependencies[@]}" "${library_dependencies[@]}"
```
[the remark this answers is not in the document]
ScottMunroMS: added the -y flag

Build/libHttpClient.Linux/openssl_Linux.bash
```diff
@@ -22,23 +22,24 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
-sudo hwclock --hctosys
-sudo rm -rf /usr/local/ssl
-sudo mkdir /usr/local/ssl
-sudo mkdir /usr/local/ssl/lib
-sudo mkdir /usr/local/ssl/include
-sudo mkdir /usr/local/ssl/include/openssl
+OPENSSL_INSTALL_DIR="$SCRIPT_DIR/../../Int/x64/$CONFIGURATION/openssl.Linux/"
```
jasonsandlin, line 34: **Severity: Critical** | **Confidence: high (both models)**

**Issue:** `curl_Linux.bash` still configures curl with `--with-openssl=/usr/local/ssl`, but this script now installs OpenSSL to a local project directory (`Int/x64/$CONFIGURATION/openssl.Linux/`). The curl build will fail to find OpenSSL headers/libraries, breaking the Linux build chain.

**Suggested Fix:** Update `curl_Linux.bash` to point `--with-openssl=` at the new local install directory, computing the path from `SCRIPT_DIR` and `CONFIGURATION` the same way this script does.

Build/libHttpClient.Linux/install_dependencies.bash
jasonsandlin: **Severity: Important**

**Issue:** `sudo apt-get install` is missing the `-y` flag. In CI (`linux-build.yml` calls this script without `--check`), the command will hang waiting for interactive confirmation and eventually time out.

**Suggested Fix:** Change to `sudo apt-get install -y ...`

Build/libHttpClient.Linux/install_dependencies.bash
```diff
@@ -0,0 +1,52 @@
...
+
+  echo "All dependencies are installed"
+  exit 0
+fi
+
+echo "Installing dependencies..."
+echo "Build dependencies: ${build_dependencies[*]}"
+echo "Library dependencies: ${library_dependencies[*]}"
+
+sudo hwclock --hctosys
```
jasonsandlin, line 55: **Severity: Important**

**Issue:** `sudo hwclock --hctosys` will fail in containers and VMs without an RTC device (common in CI). This will halt the script before dependencies are installed.

**Suggested Fix:** Either remove it (CI agents have correct time via NTP), or guard with `sudo hwclock --hctosys 2>/dev/null || true`.

Utilities/Pipelines/Tasks/linux-build.yml
jasonsandlin, line 11: **Severity: Important**

**Issue:** This pipeline step calls `install_dependencies.bash` without `--check`, meaning it runs the full `sudo apt-get install` path. Combined with the missing `-y` flag, this will hang in CI. If the CI image already has deps pre-installed, this should use `--check` instead.

**Suggested Fix:** Either pass `--check` to only verify dependencies, or ensure the script uses `apt-get install -y` for non-interactive usage.

Build/libHttpClient.Linux/libHttpClient_Linux.bash
```diff
@@ -56,15 +62,21 @@ done
...
+  bash "$SCRIPT_DIR"/install_dependencies.bash
+  if [ $? -ne 0 ]; then
+    echo ""
+    echo "Failed to install dependencies."
+    exit 1
+  fi
+else
+  bash "$SCRIPT_DIR"/install_dependencies.bash --check
+  if [ $? -ne 0 ]; then
+    echo ""
```
jasonsandlin: **Severity: Important**

**Issue:** The script now always enforces `install_dependencies.bash --check` with no true bypass option. `--skipaptget` is now a no-op. This can hard-fail builds on non-dpkg environments or custom toolchains where `dpkg -s` won't find packages installed via other means.

**Suggested Fix:** Restore a true bypass option (e.g., `--skip-dependency-check`) or make the check a warning rather than a hard failure for environments where dpkg probing is unreliable.

jasonsandlin (commented):
## Code Review (GPT-5.3-Codex + Claude Opus 4.6)

Dual-model AI code review. 5 findings: 1 Critical, 4 Important.

Build/libHttpClient.Linux/install_dependencies.bash
ScottMunroMS: this is a preexisting clock call that has just been moved to a new script. the AI feedback seems to not have a clear idea what it's trying to suggest either. It's concerned about CI containers not having an RTC but also suggests it can be removed because CI agents have correct time via NTP?

I'm going to leave this as-is and will leave it as a future exercise to LHC maintainers to strip out this clock call if they deem it so necessary.

for now I'm trying to minimize waffling about with unrelated changes/improvements

ScottMunroMS: oh wait! I misread the comment. it wants me to remove the hwclock call BECAUSE its expected to fail in those cloud containers. doh!

in either case, I'm still not planning to bother. this clock call predates this PR. I'm betting to deal with WSL clock sync shenanigans and its not actively broken so its outside the scope of this PR

Build/libHttpClient.Linux/openssl_Linux.bash
ScottMunroMS: fixt

Utilities/Pipelines/Tasks/linux-build.yml
ScottMunroMS: the pipelines previous intent is to force the dependencies to be installed. that intent is retained and no change is needed. also confirmed that this is infact installing dependencies? it seems to be installing clang surprisingly enough. may be interesting to double check how necessary that is out of band of this PR

Build/libHttpClient.Linux/libHttpClient_Linux.bash
ScottMunroMS: allowed dependency bypass

Build/libHttpClient.Linux/openssl_Linux.bash
```diff
@@ -22,48 +24,59 @@ while [[ $# -gt 0 ]]; do
...
 fi
-if [ ! -d /usr/local/ssl/lib ] ; then
-    echo "Directory /usr/local/ssl/lib does not exist"
+if [ ! -d "$OPENSSL_INSTALL_DIR/lib" ] ; then
+    echo "Directory $OPENSSL_INSTALL_DIR/lib does not exist"
     exit 1
 fi
-if [ ! -d /usr/local/ssl/include/openssl ] ; then
-    echo "Directory /usr/local/ssl/include/openssl does not exist"
+if [ ! -d "$OPENSSL_INSTALL_DIR/include/openssl" ] ; then
```
jasonsandlin, line 50: **Critical** (high-confidence - flagged by both GPT-5.3-Codex and Claude Opus 4.6)

`if [-f Makefile ]; then` is a bash syntax error - missing a space between `[` and `-f`. Bash interprets `[-f` as a command name which doesn't exist. Since it's in an `if` condition, `set -e` won't kill the script, but the condition always evaluates to false so `make clean` will **never** run - even when a Makefile exists. Stale build artifacts from a prior configuration could corrupt the build.

**Suggested fix:** Change to `if [ -f Makefile ]; then` (add a space after `[`).

Build/libHttpClient.Linux/libHttpClient_Linux.bash
```diff
@@ -56,15 +69,31 @@ done
...
-if [ "$DO_APTGET" != "false" ]; then
-  sudo hwclock --hctosys
-  sudo apt-get update
-  sudo apt-get install clang
-  sudo apt-get install make
-  sudo apt-get install autoconf
-  sudo apt-get install automake
-  sudo apt-get install libtool
-  sudo apt-get install zlib1g zlib1g-dev
+if [ "$INSTALL_DEPENDENCIES" = "true" ]; then
```
jasonsandlin, line 73: **Important** (single-model - Claude Opus 4.6)

With `set -e` active, if `bash install_dependencies.bash` returns non-zero the script terminates immediately at that line. The subsequent `if [ $? -ne 0 ]` check and the 'Failed to install dependencies.' message are dead code that will never execute.

(The `else` branch correctly handles this with `set +e`/`set -e` bracketing, but this branch doesn't.)

**Suggested fix:** Either wrap with `set +e` before the call and `set -e` after (like the check branch), or use `if ! bash install_dependencies.bash; then` which is exempt from `set -e` termination.

jasonsandlin (commented):
Code Review (v2) by GPT-5.3-Codex + Claude Opus 4.6 - 2 findings (1 Critical, 1 Important). Previous 5 findings from v1 review have all been addressed in this update.

Build/libHttpClient.Linux/libHttpClient_Linux.bash
ScottMunroMS: fixt!

Build/libHttpClient.Linux/openssl_Linux.bash
ScottMunroMS: fixt

jasonsandlin (approved):

PR #965, opened 2026-04-28
WebSocket Compression support

jhugard:
## Summary

> Notes:
> 1. `NetworkState` owns a single built-in WebSocket provider.
> 2. Windows-family mixed-backend behavior lives in `WinHttpHybrid_WebSocketProvider`.
> 3. The primary public control surface on the built-in implementations that expose the new behavior is `HCWebSocketSetOptions(...)`.
> 4. `LegacySemantics` explicitly preserves the existing behavior; non-legacy options select deterministic behavior, and compression requests route to the compression-capable path where needed.

This PR adds first-class WebSocket compression support to `libHttpClient` and includes the build, transport, API, and validation changes needed to use it safely on the important built-in paths.

At a high level, this PR:

- adds the dependency and build plumbing needed for built-in compression support across Linux, macOS, iOS, and the Windows-family paths that require extra transport support
- introduces Windows-family hybrid routing so Win32 and GDK can request compression without changing the caller-facing control surface
- adds `HCWebSocketSetOptions(...)` and `HCWebSocketOptions` as the primary public control surface for legacy preservation, deterministic behavior selection, and compression negotiation on the built-in implementations that expose that surface
- explicitly separates legacy behavior from a deterministic, message-oriented contract instead of keying behavior selection on compression alone
- makes deterministic behavior reject fragment callback registration and reject oversize receives with `HCWebSocketCloseStatus::TooLarge` socket close, effectively deprecating the fragment-handling APIs on the built-in implementations that expose the new options surface and enforcing consistent semantics across platforms
- adds upgrade-response header inspection, disconnect-with-status, and related WebSocket control improvements
- adds focused Windows compression coverage and unit-test updates that lock down the final behavior, including deterministic oversize handling without compression
- folds in the clean-branch fixes for websocketpp close-status handling and websocket async queue topology, plus the post-rebase connect-completion, no-websockets build-variant, strict-mode WinTLS build, and Apple strict-warning wrapper repairs

## Why this is needed

The motivating problem is straightforward:

- Some applications depend on compression being broadly available
- libHttpClient does not provide compression negotiation and control, nor close status control, nor response header access
- WinHTTP WebSockets do not directly support RFC 7692 `permessage-deflate`
- Existing (legacy) handling of over-sized messages is lossy, as well as inconsistent across platforms
- Win32 desktop and GDK currently default to WinHTTP for built-in WebSockets, so they need extra routing work to participate in that broader compression-capable path without also creating a second behavioral contract for deterministic mode

This PR closes that gap by adding compression-capable built-in WebSocket support across Linux, macOS, iOS, and the Windows-family paths that need it, while keeping the caller-facing configuration surface backend-agnostic and the deterministic receive contract consistent.

### Public options surface

- `HCWebSocketSetOptions(...)` is the primary control surface on the built-in implementations that expose deterministic behavior and compression negotiation
- `HCWebSocketOptions::LegacySemantics` explicitly preserves the existing legacy behavior by retaining the existing provider behavior and semantics
- `HCWebSocketOptions::None` selects deterministic behavior without requesting compression
- `HCWebSocketOptions::RequestCompression` selects deterministic behavior and requests `permessage-deflate`
- `CompressionServerNoContextTakeover` and `CompressionClientNoContextTakeover` refine the compression request and require `RequestCompression`

### Legacy behavior

Existing behavior, aka legacy behavior, on Win32 / GDK differs from other platforms:

- Once an incoming payload exceeds the receive buffer (20KB by default), WinHTTP starts forwarding partial chunks through a raw-byte binary fragment path
- Without a caller-registered fragment handler, the fallback is lossy and misleading instead of message-oriented:
  - oversized binary messages are surfaced as multiple partial binary whole-message callbacks
  - oversized UTF-8 messages surface as _binary_ whole-message callbacks for the earlier chunks plus a final _text_ whole-message callback for only the residual tail
  - if no binary _whole-message_ callback is installed, those earlier oversized text chunks are dropped entirely and instead deliver only a final text whole-message callback for only the residual tail
- With a fragment handler installed, receive processing becomes fragmented:
   - undersized binary and UTF-8 messages are delivered through their normal respective message delivery callbacks
   - both oversized binary and oversized UTF-8 messages are delivered to the fragment handler chunk-by-chunk, without distinguishing between binary and text messages
- No means is provided to notify the server when a too-large message is received

Other platforms:

- default to ~32MB max receive buffer size, with no functioning means to adjust that size
- automatically close the socket with a too-large status if an oversized message is detected
- do not provide fragment handling callback support

### Legacy versus deterministic behavior

- If `HCWebSocketSetOptions(...)` is never called, Win32 and GDK stay on the legacy WinHTTP-facing behavior
- `LegacySemantics` is an explicit no-op that preserves that same behavior
- Non-legacy options select deterministic behavior on supported paths
- Deterministic behavior provides a single message-oriented contract across those supported built-in paths: fragment callback registration is rejected, fragment callback delivery does not occur, and inbound message-size enforcement is consistent
- Deterministic behavior treats `HCWebSocketSetMaxReceiveBufferSize(...)` as a hard inbound message-size cap; when a message exceeds that cap, the socket is closed with `HCWebSocketCloseStatus::TooLarge` (`1009`)
- If the max receive buffer is not set explicitly before connect, deterministic behavior uses the default `32,000,000` byte limit
- On Win32 and GDK, `HCWebSocketOptions::None` keeps the existing WinHTTP transport while enforcing that same deterministic hard-cap behavior; compression requests route through the hybrid provider to the compression-capable path
- Android does not expose `HCWebSocketSetOptions(...)` or `HCWebSocketSetMaxReceiveBufferSize(...)`, as the OkHttp provider automatically negotiates compression and has a fixed ~32MB message limit
- WinRT / UWP remains on the existing `MessageWebSocket` path and is out of scope for this compression work: this PR does not add compression negotiation there, and the deterministic `HCWebSocketSetOptions(...)` / hard-cap receive-limit behavior does not apply to that provider

### Build and dependency groundwork

- refreshed the dependency graph needed for the websocketpp transport, including `asio`, `websocketpp`, and `boost-wintls`
- moved `asio` forward to the newest revision that still preserves the public APIs `boost-wintls` depends on, instead of taking the latest upstream breaking API removals in the same change
- updated the Windows shared project items, solution files, package metadata, Linux build helpers, and Apple Xcode project so the new transport builds through the existing entry points
- hardened the Linux helper scripts for containerized dependency installation, OpenSSL install, and cURL artifact export on bind-mounted workspaces
- disabled optional `libzstd` autodetection in the Linux vendored cURL helper so static CI-style builds do not pick up an undeclared zstd link dependency
- aligned the Apple Xcode minizip wiring with the Linux and Android build definitions by adding `ioapi.c` and the `contrib/minizip` include path
- localized Apple Clang `-Wdocumentation` suppression around the vendored websocketpp permessage-deflate include so strict Xcode warning-as-error builds do not fail on a third-party documentation comment defect
- enabled Linux and Apple build-time plumbing for WebSocket compression while preserving explicit build control

### Windows hybrid routing and receive-limit unification

- added the websocketpp provider and the supporting WinTLS transport adapter
- introduced `WinHttpHybrid_WebSocketProvider` so Windows-family backend selection stays inside a concrete provider instead of leaking into global selector state
- wired Win32 and GDK to route compression requests through the compression-capable transport while leaving deterministic non-compression connections on WinHTTP
- updated the WinHTTP receive path so deterministic mode enforces the same hard inbound message-size cap and `TooLarge` close behavior as the other built-in deterministic paths
- updated related WinHTTP, WinRT, Apple, and `NetworkState` plumbing so the new transport integrates with the existing lifecycle and shutdown paths

### Explicit WebSocket options semantics

- primary public control surface on the applicable built-in implementations is `HCWebSocketSetOptions(...)`, with `HCWebSocketOptions` providing the control knobs
- handle-level state tracks whether options and max receive buffer values were set explicitly before connect
- preserved legacy behavior explicitly through `LegacySemantics` and tightened deterministic behavior across the supported built-in paths
- updated exports, docs, and platform-facing surfaces to match the new options contract

### Samples and validation coverage

- added a local WebSocket echo-server sample used by the validation harnesses
- added dedicated Win32 compression coverage projects and certificate validation coverage
- added unit coverage around options validation, `LegacySemantics`, fragment-handler rejection, and explicit max-receive-buffer behavior

### Current websocketpp scale posture

- This branch does not modify the upstream websocketpp provider-owned background thread per WebSocket (`m_websocketThread = std::thread(...)` and `std::thread m_websocketThread`)
- This places severe limits on how many client connections can be made per process, due to thread thrashing and other internal contention
- The Win32/GDK uncompressed WinHTTP websocket path was tested up to 12K clients in a single process, without issue
- However, all paths using the websocketpp provider, including the Win32/GDK compressed path, are likely to start failing due to contention at around 5K and have high CPU utilization
- Adjusting the websocketpp architecture for better scalability is out-of-scope for this PR, but has been tested as feasible as a basis for an independent PR

### Review feedback fixes

- added `boost-wintls` to `NOTICE.txt` and `cgmanifest.json`; corrected the stale `asio` commit hash in both files
- added `DEFINE_ENUM_FLAG_OPERATORS(HCWebSocketOptions)` so callers can combine flags with `|` without casting through `uint32_t`
- simplified the remaining low-risk `HCWebSocketOptions` flag-handling sites in the runtime and tests now that the enum flag operators exist; removed leftover test-only helper functions and replaced the remaining raw legacy-options equality check
- added `m_stateMutex` lock to `SetOptions()` and `SetPingInterval()` for consistency with the locking pattern in `SetMaxReceiveBufferSize()` and the other pre-connect setters
- changed `GetResponseHeaderAtIndex()` to return `E_INVALIDARG` on out-of-range index, consistent with the existing `HCHttpCallResponseGetHeaderAtIndex()` contract
- added Win32 `.vcxitems.filters` entries for the new websocketpp and hybrid-routing files
- added documentation comments for the intentional `raw new` in `wintls_socket.hpp`, the detached-joiner shutdown fallback, and the duplicated proxy URI parser
- clarified the public `httpClient.h` and README wording around `LegacySemantics`, deterministic behavior, and receive-limit behavior; aligned the related WinHTTP wording/flow to match
- repaired a late merge regression in WebSocket connect completion handling
- fixed the Win32 and GDK no-websockets export surfaces and promoted Linux `HC_NOWEBSOCKETS` to a first-class build option so the post-rebase no-websockets variants stay buildable
- made the `WinHttpHybrid_WebSocketProvider` provider-selection helpers `const` in response to late review feedback
- fixed a latent WinTLS websocketpp `translate_ec(...)` compile issue so the branch still builds cleanly under stricter MSVC conformance modes used by downstream consumers
- moved the Release `/OPT:NOREF` intent for the Win32 compression test executables into the linker settings and paired it with plain `/LTCG` so the setting actually takes effect
- rejected `HCWebSocketSetMaxReceiveBufferSize(0)` with `E_INVALIDARG` to avoid invalid deterministic receive-buffer states
- removed the non-GDK always-true `curl_multi_poll` warning while preserving the GDK dynamic-loader fallback behavior

### Local Windows build matrix

- built `libHttpClient.vs2022.sln` for `Debug|x64`
- built `libHttpClient.vs2022.sln` for `Release|x64`
- built `libHttpClient.vs2022.sln` for `Debug|x86`
- built `libHttpClient.vs2022.sln` for `Release|x86`
- built `libHttpClient.vs2022.sln` for `Debug|ARM64`
- `Release|ARM64` is blocked locally by missing Spectre-mitigated libraries (`MSB8040`)
- `Debug|Gaming.Desktop.x64` and `Release|Gaming.Desktop.x64` are blocked locally by missing Microsoft GDK integration and missing local `v143` GDK toolset support (`MSB8020`)

### Local Windows test runs

- ran the full `x64 Debug` TE suite via `vstest.console.exe`
- all WebSocket unit coverage passed in that run, including the options and deterministic-semantics tests
- local TE noise remains in `TaskQueueTests`, as expected
- `AsyncBlockTests::VerifyFailureInDoWork` failed once in the full TE run but passed immediately when rerun in isolation
- built and ran `Tests\WebSocketCompression\WebSocketCompressionTests.Win32.vcxproj`
- validated deterministic oversize-close behavior, `permessage-deflate` negotiation, and the no-context-takeover combinations successfully against the local echo server
- built and ran `Tests\WebSocketCompression\WebSocketCompressionIntegrationTests.Win32.vcxproj`
- validated trusted localhost, wrong-host `CERT_E_CN_NO_MATCH`, untrusted-root `CERT_E_UNTRUSTEDROOT`, and refused-connect diagnostics successfully

### Linux and Android probe results

- Linux Debug builds completed for `libHttpClient.Linux.so` and `WebSocketCompressionTests.Linux`
- ran `WebSocketCompressionTests.Linux` directly under WSL and validated both the deterministic non-compression path (`HCWebSocketOptions::None`, including the `TooLarge` oversize close) and the negotiated compression path (`RequestCompression` plus the no-context-takeover variants) successfully against the local echo server
- Local machine does not currently have a configured Android SDK or NDK, so `assembleDebug` could not start

### Post-rebase Windows and Linux variant checks

- built `Build\libHttpClient.Win32\libHttpClient.Win32.vcxproj` for `Release|x64`
- ran `Out\x64\Release\WebSocketCompressionTests.Win32\WebSocketCompressionTests.Win32.exe` successfully
- built `Build\libHttpClient.Win32\libHttpClient.Win32.vcxproj` for `Release|x64` with `/p:HCEnableWebSocketCompression=false`
- built `Build\libHttpClient.Win32\libHttpClient.Win32.vcxproj` for `Release|x64` with `/p:HCNoWebSockets=true`
- built `Tests\WebSocketCompression\WebSocketCompressionTests.Win32.vcxproj` for `Release|x64` with the Release `/OPT:NOREF` + `/LTCG` follow-up applied
- built `Tests\WebSocketCompression\WebSocketCompressionIntegrationTests.Win32.vcxproj` for `Release|x64` with the same Release linker follow-up applied
- manually compiled `Source\WebSocket\Websocketpp\websocketpp_websocket.cpp` under `/permissive-` against the branch include/define set to confirm the WinTLS transport also builds under stricter MSVC conformance settings
- ran `Build\libHttpClient.Linux\libHttpClient_Linux.bash --install-dependencies -c Release` successfully inside an `ubuntu:22.04` container
- built the Linux Release shared library with the default compression-enabled configuration and ran `ctest --output-on-failure -R websocket-compression-linux` successfully under WSL
- reran the Linux Release helper build after the `CurlMulti.cpp` cleanup; `WebSocketCompressionTests.Linux` still linked successfully and the non-GDK warning disappeared
- built the Linux Release shared library with `HC_ENABLE_WEBSOCKET_COMPRESSION=OFF`
- built the Linux Release shared library with `HC_NOWEBSOCKETS=1`

### Not validated locally

- broader Linux matrix coverage beyond the WSL Debug runtime validation above was not rerun locally
- The latest iOS CI compile failure was diagnosed from the upstream log and fixed by localizing the websocketpp `-Wdocumentation` suppression in the configured permessage-deflate wrapper, but that rerun has not been observed locally
- GDK compilation, including the post-rebase no-websockets export-surface follow-up, is blocked locally by missing toolchain components, as noted above

## Scope boundary for this PR

- this PR does not patch `websocketpp` itself
- this PR keeps the Windows mixed-backend decision inside the built-in provider layer rather than surfacing backend selection to callers
- this PR intentionally does not preserve the old WinHTTP oversized-fragment behavior once callers opt into deterministic behavior
- this PR does not expose the new options surface on Android, which negotiates compression by default and has a fixed ~32MB maximum message size
- this PR does not add compression support or deterministic receive-limit enforcement to WinRT / UWP, which continues to use the existing `MessageWebSocket` provider behavior
- this PR does not move Android onto websocketpp; the built-in Android provider remains the existing OkHttp-based implementation

## Risks / review focus

Internal review should focus especially on:

- whether the websocketpp + WinTLS dependency footprint is acceptable for this PR
- whether `WinHttpHybrid_WebSocketProvider` is the right long-term ownership point for Windows-family mixed-backend behavior
- whether `HCWebSocketSetOptions(...)` is the right caller contract for expressing both legacy preservation and deterministic behavior selection
- whether the deterministic hard-cap behavior is the right replacement for the older fragment-overflow behavior when callers opt out of legacy semantics
- whether the Android posture of omitting the new options and max-buffer APIs is the right product choice
- whether the current compression and certificate coverage is sufficient without additional semantics-specific coverage

## Suggested review path

1. `7a3ef9b` — `Add WebSocket compression build and dependency plumbing`
   Dependency updates, shared-project wiring, Linux and Apple compression build plumbing, and solution/package metadata changes.

2. `b739781` — `Introduce the websocketpp-backed Windows transport`
   The websocketpp provider, WinTLS transport adapter, WinHTTP hybrid routing, and the platform-specific wiring for Win32 and GDK.

3. `5698ddb` — `Replace compression flags with explicit WebSocket options`
   The `HCWebSocketSetOptions(...)` surface, `LegacySemantics`, deterministic behavior, Android surface cleanup, and the export/documentation updates.

4. `818751b` — `Add WebSocket samples and compression validation coverage`
   The echo-server sample, compression and certificate integration coverage, and the unit-test updates that lock down the final behavior.

5. `a593c5b` — `Unify deterministic WebSocket behavior on WinHTTP`
   The WinHTTP deterministic receive-limit enforcement, hybrid routing adjustment for non-compression deterministic mode, deterministic oversize validation updates, and the final README/public documentation alignment.

6. `0fa4a01` — `Update third-party compliance metadata (AI Review)`
   Adds `boost-wintls` to `NOTICE.txt` and `cgmanifest.json`; corrects the stale `asio` commit hash.

7. `0a76816` — `Fix WebSocket API issues (AI Review)`
   Adds `DEFINE_ENUM_FLAG_OPERATORS`, mutex consistency for `SetOptions`/`SetPingInterval`, `E_INVALIDARG` for out-of-range header index.

8. `81436e8` — `Build hygiene and documentation comments (AI Review)`
   Win32 `.vcxitems.filters` entries, documentation comments for design choices and intentional duplication.

9. `d1e0f90` — `Simplify WebSocket option flag handling (AI Review)`
   Low-risk cleanup of the remaining option-flag handling in the runtime and tests after the enum flag operators landed.

10. `b1b8d97` — `Handle websocketpp close status consistently`
   Aligns websocketpp close-status handling with the rest of the deterministic behavior and removes a late clean-branch inconsistency in the websocketpp path.

11. `c475339` — `Fix websocket async queue topology`
   Keeps the clean branch's websocket async queue usage aligned with the intended control flow.

12. `10aed2d` — `Address PR feedback on WebSocket semantics docs`
   Clarifies the public `httpClient.h` and README wording around `LegacySemantics`, deterministic behavior, and receive-limit expectations; includes the small aligned WinHTTP wording/flow cleanup.

13. `5ba0134` — `Fix websocket connect completion merge`
   Repairs a late merge regression in `hcwebsocket.cpp` so connect completion follows the intended post-rebase control flow.

14. `6b8a612` — `Fix no-websockets build variants`
   Fixes the Win32 and GDK no-websockets export surfaces and the Linux `HC_NOWEBSOCKETS` build plumbing so the post-rebase no-websockets variants remain buildable.

15. `9fbc9c1` — `Make hybrid websocket helpers const`
   Applies the late review-feedback const-correctness cleanup to the non-mutating `WinHttpHybrid_WebSocketProvider` provider-selection helpers.

16. `cdaebe6` — `Fix strict-mode WinTLS websocket build`
   Fixes a latent `wintls_socket.hpp` transport-error translation compile issue so the websocketpp-backed Windows compression path still builds under stricter MSVC conformance settings.

17. `c90949a` — `Harden Linux helper scripts for container builds`
   Hardens the Linux dependency-install, OpenSSL, and cURL helper scripts for the containerized build path used in local and CI-style validation.

18. `6cdd639` — `Silence non-GDK curl multi poll warning`
   Removes the always-true non-GDK `curl_multi_poll` check while preserving the GDK dynamic-loader fallback behavior.

19. `cc5fa18` — `Align Apple minizip Xcode wiring`
   Adds `ioapi.c` and the `contrib/minizip` include path to the Apple Xcode project so its minizip wiring matches the Linux and Android build definitions.

20. `5f8ec91` — `Fix Win32 compression test linker settings`
   Moves the Release `/OPT:NOREF` control for the Win32 compression test executables into Link and forces plain `/LTCG` so that setting is compatible.

21. `fba869d` — `Reject zero WebSocket receive buffer size`
   Rejects `HCWebSocketSetMaxReceiveBufferSize(0)` with `E_INVALIDARG` to avoid invalid deterministic receive-buffer states.

22. `bae2321` — `Disable zstd in Linux curl helper`
   Disables optional zstd autodetection in the Linux vendored cURL helper so static builds keep a deterministic link surface and do not require undeclared `libzstd` symbols.

23. `7dad3d9` — `Silence Apple websocketpp documentation warning`
   Localizes Apple Clang `-Wdocumentation` suppression to the configured permessage-deflate wrapper so strict iOS/macOS warning-as-error builds do not fail on the vendored websocketpp header.


Source/WebSocket/hcwebsocket.cpp
```diff
@@ -412,22 +512,98 @@ HRESULT WebSocket::SetProxyDecryptsHttps(
...
+    RETURN_HR_IF(
+        E_INVALIDARG,
+        RequestsLegacyWebSocketSemantics(options) &&
+            options != HCWebSocketOptions::LegacySemantics);
+    RETURN_HR_IF(
+        E_INVALIDARG,
+        HasNoContextTakeoverWebSocketOptions(options) && !RequestsWebSocketCompression(options));
+
+    std::lock_guard<DefaultUnnamedMutex> lock{ m_stateMutex };
+    RETURN_HR_IF(E_HC_CONNECT_ALREADY_CALLED, m_state != State::Initial);
```
jasonsandlin, line 559: **Important**

`SetOptions` acquires `m_stateMutex` (line 553), then calls `HasBinaryMessageFragmentHandlers()` (line 562) which acquires `m_eventCallbacksMutex` (line 572). This establishes lock ordering: `m_stateMutex` -> `m_eventCallbacksMutex`. If any other code path (e.g., notification/callback dispatch) takes these locks in the reverse order, it creates a potential ABBA deadlock. The pre-connect state guard (`m_state != State::Initial`) reduces practical risk since callbacks won't be firing before connect, but the lock ordering should be verified across all paths for correctness.

**Suggested fix:** Audit all code paths that acquire both `m_stateMutex` and `m_eventCallbacksMutex` to confirm consistent ordering. Alternatively, refactor `SetOptions` to read `HasBinaryMessageFragmentHandlers()` before acquiring `m_stateMutex`, or extract the check outside the critical section.

Source/WebSocket/Websocketpp/wintls_socket.hpp
```diff
@@ -0,0 +1,294 @@
...
+        }
+        else
+        {
+            m_socket->async_shutdown(callback);
+        }
+    }
+
+public:
+    template <typename ErrorCodeType>
+    static lib::error_code translate_ec(ErrorCodeType)
```
jasonsandlin, line 225: **Important**

The generic `translate_ec` template converts all non-`lib::error_code` error types (including `wintls::error_code` carrying Schannel/SSPI status codes) into a single `pass_through` error, discarding the original platform error information. TLS handshake failures like `SEC_E_CERT_UNKNOWN` or `CERT_E_CN_NO_MATCH` will reach `HResultFromConnectError` as a generic error, producing `E_FAIL` rather than a meaningful HRESULT. The security audit document flags this: "TLS failure reporting is more generic than ideal."

**Suggested fix:** For `wintls::error_code` specifically, extract the underlying Windows HRESULT/Win32 error code and propagate it through the asio `system_category` so that `HResultFromConnectError` can produce a meaningful `HRESULT_FROM_WIN32`. This improves debuggability of TLS failures in production.

Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
```diff
@@ -948,60 +1752,70 @@ struct wspp_websocket_impl : public hc_websocket_impl, public std::enable_shared
...
         {
-            if (is_tls_client())
-            {
-                return reinterpret_cast<websocketpp::client<WebsocketConfig> &>(tls_client());
-            }
-            else
-            {
-                return reinterpret_cast<websocketpp::client<WebsocketConfig> &>(non_tls_client());
-            }
+            return *reinterpret_cast<websocketpp::client<WebsocketConfig>*>(client_storage());
```
jasonsandlin, line 1795: **Important**

`impl<T>()` performs an unchecked `reinterpret_cast` from `void*` (returned by `client_storage()`) to `websocketpp::client<T>*`. The old code had the same pattern but with only 2 config types (ws/wss). With 12+ config types now (ws, wss, ws_deflate, wss_deflate, ws_wintls, wss_wintls, ws_wintls_deflate, etc.), the risk of a mismatch between the template parameter and the actual stored type is higher. A mismatch would be silent undefined behavior.

**Suggested fix:** Add a debug-mode type check: store `std::type_info` or a type tag at construction time and `assert()` in `impl<T>()` that the requested type matches. This catches dispatch errors during development at zero cost in release builds.

Source/WebSocket/hcwebsocket.cpp
jhugard: Good catch on the nesting hazard. You're right that SetOptions established a m_stateMutex -> m_eventCallbacksMutex ordering. No current code path takes them in reverse (e.g. CloseFunc explicitly unlocks m_stateMutex before acquiring m_eventCallbacksMutex), but it's a latent trap for future changes.

Fixed in ea05d34: HasBinaryMessageFragmentHandlers() is now called before acquiring m_stateMutex, eliminating the nested lock entirely. This is safe because SetOptions is only valid in State::Initial (enforced immediately after acquiring the lock), so no concurrent callback registration or dispatch can be in flight.

Source/WebSocket/Websocketpp/wintls_socket.hpp
jhugard: This turns out to be a false positive for our build configuration. wintls::error_code is a typedef for std::error_code (established via WINTLS_USE_STANDALONE_ASIO in wintls/detail/config.hpp), and websocketpp::lib::error_code is also std::error_code (via websocketpp/common/system_error.hpp). Since they are the same type, the specific translate_ec(lib::error_code ec) overload wins over the generic template - TLS errors are not converted to pass_through.

The error propagates with its original SECURITY_STATUS value in system_category. HResultFromConnectError recognizes system_category and routes through HResultFromPlatformNetworkError, which handles negative values (which SECURITY_STATUS codes like SEC_E_CERT_UNKNOWN / 0x80090327 are) as direct HRESULT passthrough.

I've added a clarifying comment at the overload set in ea05d34 to make this resolution explicit for future readers.

Source/WebSocket/Websocketpp/websocketpp_websocket.cpp
jhugard: Agreed - with 10+ config types the type-erasure dispatch has enough indirection that a mismatch would be silent UB. The invariant is maintained today by the disciplined create_client -> invoke_wspp_client correspondence, but there's no runtime validation that the template parameter matches the concrete stored type.

Fixed in ea05d34: websocketpp_client_base now stores a m_typeTag (set to typeid(websocketpp::client<Config>).hash_code() at construction), and impl<T>() asserts the tag matches before the cast. This catches dispatch errors during development at zero cost in release builds (ASSERT compiles out).

SECURITY_AUDIT_BOOST-WINTLS.md
```diff
@@ -0,0 +1,255 @@
+# Security Audit: `libHttpClient` websocket compression backend (`websocketpp` + `boost-wintls`)
+
```
jasonsandlin: is this file needed?  seems like should be deleted.

jasonsandlin:
## Code Review Findings

Reviewed the core source changes (API surface, WebSocket core, WinHTTP providers, websocketpp). Overall well-engineered — architecture is sound, the deterministic/legacy split is clean, and decompression bomb risk is mitigated by `set_max_message_size`. Three items worth a look:

---

### 1. Latent 0-byte buffer in `winhttp_connection.cpp` (~line 1707)

When the receive buffer is already at `receiveBufferLimit`, the expansion logic computes `newSize = current * 2`, caps to `receiveBufferLimit`, then calls `Resize` with the current size — a no-op that leaves `GetRemainingCapacity() == 0`. The subsequent `WinHttpWebSocketReceive` would then get a 0-byte buffer.

Currently unreachable due to upstream control flow, but fragile if that changes.

**Suggested fix:** Guard before the `Resize` call — if `newSize <= m_websocketReceiveBuffer.GetBufferByteCount()`, return `E_UNEXPECTED` instead of calling Resize with the same size.

---

### 2. Silent failure in `winhttp_websocket_hybrid.cpp` (~line 649)

`ActiveProvider` returns `nullptr` (callers return `E_UNEXPECTED`) when the impl pointer doesn't match either provider type. If the connection failed and `impl` was never set, the user gets a generic `E_UNEXPECTED` with no diagnostic info.

**Suggested fix:** Add an `HC_TRACE_ERROR` before returning `nullptr`.

---

### 3. Dangling pointer risk in `hcwebsocket.cpp` (~line 603)

`GetResponseHeader` acquires `m_stateMutex`, reads the header map's `c_str()` pointer, then releases the lock. If `ClearResponseHeadersLockHeld` (called in `ConnectComplete` on failure) runs concurrently on another thread, the returned pointer is dangling.

This is a new risk introduced by the `ClearResponseHeaders` path in this PR.

**Suggested fix:** Either document that the returned pointer is only valid while the WebSocket is in Connected state, or check `m_state == State::Connected` inside the lock and return `E_UNEXPECTED` otherwise.

jasonsandlin:
## Backward Compatibility Note

Verified that **all changes in this PR are opt-in only** — existing consumers are unaffected:

- **No existing function signatures or callbacks changed** — the PR only adds new APIs (`HCWebSocketSetOptions`, `HCWebSocketSetCompression`, `HCWebSocketDisconnectWithStatus`, response header getters)
- **Default behavior preserved** — `m_options` defaults to `HCWebSocketOptions::None` with `m_optionsExplicitlySet = false`, so without calling `HCWebSocketSetOptions()`, sockets use legacy platform behavior exactly as before
- **Compression is off by default** — requires explicitly passing `HCWebSocketOptions::RequestCompression`
- **Custom providers not broken** — `IWebSocketProvider` gains a new virtual `OptionsResult()` with a default implementation returning `E_NOT_SUPPORTED`, so existing custom providers still compile without changes
- **Callback signatures unchanged** — `HCWebSocketMessageFunction`, `HCWebSocketBinaryMessageFunction`, `HCWebSocketBinaryMessageFragmentFunction`, and `HCWebSocketCloseEventFunction` are all unchanged. Semantics only change if the caller opts into deterministic/compression modes

jhugard:
> ## Code Review Findings
> Reviewed the core source changes (API surface, WebSocket core, WinHTTP providers, websocketpp). Overall well-engineered — architecture is sound, the deterministic/legacy split is clean, and decompression bomb risk is mitigated by `set_max_message_size`. Three items worth a look:
> [snip]

Addressed the above items in one commit.

Also, in a separate commit moved the websocket response header accessor functions to `httpClient.h`, from `httpProvider.h`, as these are intended for client usage as well as provider usage.  This commit also includes minor SAL fixes and doc edits to make the websocket handle ownership contract clear.

jasonsandlin (approved):

SECURITY_AUDIT_BOOST-WINTLS.md
jhugard: Agree.  I had left it in for review (to cover the new dependency), but in retrospect would have been better to include with the PR description, either in full or in summary.

File deleted.

jasonsandlin (approved):

Issue #971, opened 2026-05-08
XTaskQueueSubmitDelayedCallback can dispatch work before the requested delay elapses


jhugard:
## Summary

Callbacks submitted through `XTaskQueueSubmitDelayedCallback` can become
dispatchable before their requested delay has elapsed. The contract of this API
is that the callback will not be invoked until at least `delayMs` milliseconds
have passed, but under certain runtime conditions the callback surfaces early —
sometimes immediately.

This affects any code that relies on `XTaskQueueSubmitDelayedCallback` for
timed waits, retry back-off, debouncing, or deferred continuations. The
symptoms appear as spurious early timeouts, premature retry loops, and
timing-sensitive async failures that are difficult to reproduce in isolation.

## Public API affected

```c
STDAPI XTaskQueueSubmitDelayedCallback(
    _In_ XTaskQueueHandle queue,
    _In_ XTaskQueuePort port,
    _In_ uint32_t delayMs,
    _In_opt_ void* callbackContext,
    _In_ XTaskQueueCallback* callback);
```

No signature change is required — the issue is a runtime violation of the
existing delay guarantee.

## Expected behavior

A callback submitted with `delayMs = N` must not become dispatchable until at
least N milliseconds have elapsed since submission, regardless of:

- whether other composite queues sharing the same underlying port are
  terminated concurrently
- whether the internal delayed-callback timer has been retargeted for an
  earlier entry since submission
- whether two delayed callbacks happen to land on the same or nearly the same
  deadline

## Actual behavior

Three distinct failure modes violate the delay contract:

1. **Termination early-promotion** — When a composite queue is terminated via
   `XTaskQueueTerminate`, its pending delayed entries are flushed. If another
   composite queue shares the same delayed port, the termination path can make
   the sibling's not-yet-due entry dispatchable immediately.

2. **Stale timer notification** — A platform timer callback that was enqueued
   for an earlier deadline can arrive after the timer has been retargeted for a
   later entry. The old code treated any timer fire as proof that the current
   armed deadline had elapsed, promoting the later entry too early.

3. **Equal-deadline aliasing** — When two delayed callbacks are submitted with
   the same or very close delay values, the old promotion logic matched pending
   entries by exact timestamp identity. This made correctness depend on whether
   two `uint64_t` due times happened to be bitwise equal rather than on the
   simpler question of whether the callback's deadline had actually passed.

In consumer code the common external symptom is the same in all three cases:
[...]

PR #972, opened 2026-05-08
Fix XTaskQueueSubmitDelayedCallback dispatching work before the requested delay

jhugard:
## Summary

This PR fixes the runtime behavior of `XTaskQueueSubmitDelayedCallback` so that
callbacks are never dispatched before their requested `delayMs` has elapsed, as
described in issue #971.

The existing public API signature is unchanged — the fix corrects a contract
violation in the internal delayed-callback scheduling path that could make
pending work become dispatchable too early. Any consumer relying on the delay
guarantee (timed waits, retry back-off, deferred continuations, connection
timeout logic) is affected by the bug and benefits from this fix.

## Bugs addressed

Three distinct failure modes in the delayed-callback promotion path:

1. **Termination early-promotion** — `XTaskQueueTerminate` on a composite queue
   could flush shared delayed-port state in a way that made a sibling queue's
   not-yet-due callback appear ready immediately.

2. **Stale timer notification** — A platform timer callback enqueued for an
   earlier deadline could arrive after the timer had been retargeted for a later
   entry. The old code treated any timer fire as proof that the *current* armed
   deadline had elapsed, promoting the later entry too early.

3. **Equal-deadline aliasing** — The old logic matched pending entries by exact
   `uint64_t` timestamp. When two delayed callbacks had the same or
   nearly-identical due time, correctness depended on whether those timestamps
   were bitwise equal rather than on whether the deadline had actually passed.

All three surface the same external symptom: a timed operation completes or
times out before the requested wait duration.

### Timer backends (`WaitTimer_win32.cpp`, `WaitTimer_stl.cpp`, `ios_WaitTimer.mm`)

- Switched from wall-clock / `high_resolution_clock` absolute times to
  **monotonic `steady_clock` due times**. This eliminates sensitivity to
  wall-clock adjustments and removes the platform-specific
  `GetSystemTimePreciseAsFileTime` machinery on Win32.
- Exposed `GetCurrentTime()` (current monotonic time) and `GetDueTime(msFromNow)`
  (future deadline) in place of the old `GetAbsoluteTime` which conflated both.
- Win32: converts the stored `steady_clock` deadline to a relative wait when
  arming `SetThreadpoolTimer`, preserving the existing one-shot mechanism.
- iOS: same conversion for `NSTimer`'s relative interval.

### Promotion logic (`TaskQueue.cpp`)

- **`PromoteReadyPendingCallbacks`** (replaces `ScheduleNextPendingCallback`):
  sweeps *all* pending entries whose `enqueueTime <= now` and promotes them,
  then re-arms the timer for the next future deadline. This replaces the old
  "pop exactly one entry whose timestamp matches the armed due time" flow,
  eliminating both the aliasing problem and the stale-notification problem.

- **`SubmitPendingCallback`**: on an early timer callback (where `now < dueTime`),
  retries against the currently armed due time instead of silently returning.
  It only re-arms when `m_timerDue` still matches the observed due time, so a
  stale callback cannot overwrite a newer earlier deadline while still
  preventing stranded entries when the Win32 threadpool timer fires slightly
  ahead of the `steady_clock` deadline.

- **`CancelPendingEntries`**: defers `AppendEntry` calls (which signal the
  dispatcher) until after the pending-list traversal completes. This keeps queue
  wakeups out of `remove_if`'s critical section and prevents re-entrant
  interaction between termination and dispatch.

### Queue termination (`TaskQueue.cpp`)

- `CancelPendingEntries` no longer touches `m_timer` or `m_timerDue`. Sibling
  delegates sharing the same delayed port keep their timer state intact. If the
  terminated queue owned the earliest armed entry, the timer simply takes one
  blank fire and re-arms for the next real item via the sweep logic above.

## Public API surface

- **No signature changes** to any public `XTaskQueue` API.
- The behavioral contract of `XTaskQueueSubmitDelayedCallback` is tightened:
  the delay guarantee now holds correctly under shared-port topologies,
  concurrent termination, and timer retargeting.
- `XTaskQueueTerminate` behavior is unchanged from the caller's perspective —
  the terminated queue's callbacks still surface as canceled. The fix only
  prevents termination from affecting *sibling* queues' delayed work.

## Regression coverage

Two new deterministic unit tests, plus the existing control:

- `VerifyTerminationDoesNotEarlyPromoteSiblingDelayedCallback` — composite
  queues sharing a delayed port; verifies the sibling's callback stays pending
  until its own deadline after the other queue is terminated.
- `VerifyStaleDelayedCallbackDoesNotEarlyPromoteNextPendingEntry` — injects a
  synthetic stale timer notification after the first of two delayed callbacks
  has been dispatched; verifies the second callback is not promoted early.
- `VerifyDelayedCallbackTimerRaceOnManualQueue` — existing control coverage
  for the timer-dispatch interleaving on manual queues.

Test hooks used to drive the deterministic interleavings are compiled only in
unit-test builds (`HC_UNITTEST_API`).

## Validation

- All existing `libHttpClient` unit tests pass.
- Downstream consumer validation against the rebuilt Release runtime, including
  a sustained high-connection-count WebSocket benchmark (12,000 concurrent
  connections, 60-second soak) that originally surfaced the early-wake behavior.

## Scope

- Win32, STL (Linux/Android), and iOS timer backends are all updated.
- The change is strictly scoped to delayed-callback scheduling correctness.
  No changes to immediate dispatch, serialized threadpool dispatch, or the
  broader task queue lifecycle.

## Review focus

- Correctness of the monotonic due-time conversion in each platform backend
  (especially the Win32 `steady_clock` → relative FILETIME conversion and
  ceiling rounding).
- The promote-all-ready sweep: does `enqueueTime <= now` correctly handle edge
  cases where multiple entries share the same deadline?
- The re-arm-on-early-callback path: does the conditional `m_timerDue`
  compare/exchange guard correctly prevent stale callbacks from overwriting a
  newer earlier deadline while still covering legitimate early-fire cases?
- `CancelPendingEntries` collect-then-publish: is there any ordering concern
  when multiple terminated entries are published in sequence after the sweep?

### DIFF
--- a/Source/Task/TaskQueue.cpp
+++ b/Source/Task/TaskQueue.cpp
@@ -362,7 +362,10 @@ HRESULT __stdcall TaskQueuePortImpl::QueueItem(
     }
     else
     {
-        entry.enqueueTime = m_timer.GetAbsoluteTime(waitMs);
+        // Delayed callbacks are ordered by a monotonic due time so stale timer
+        // callbacks and wall-clock adjustments cannot make one pending entry
+        // masquerade as another.
+        entry.enqueueTime = m_timer.GetDueTime(waitMs);
         RETURN_HR_IF(E_OUTOFMEMORY, !m_pendingList->push_back(entry));
 
         // If the entry's enqueue time is < our current time,
@@ -959,12 +962,17 @@ void TaskQueuePortImpl::CancelPendingEntries(
     // share this port's delayed-callback timer state, so leave m_timer and
     // m_timerDue alone; if we removed the armed earliest entry, the existing
     // timer simply takes one blank fire and re-arms for the next real item.
+    LocklessQueue<QueueEntry> entriesToAppend(*m_queueList.get());
 
     m_pendingList->remove_if([&](auto& entry, auto address)
     {
         if (entry.portContext == portContext)
         {
-            if (!appendToQueue || !AppendEntry(entry, address))
+            if (appendToQueue)
+            {
+                entriesToAppend.push_back(std::move(entry), address);
+            }
+            else
             {
                 entry.portContext->Release();
                 m_pendingList->free_node(address);
@@ -975,6 +983,22 @@ void TaskQueuePortImpl::CancelPendingEntries(
 
         return false;
     });
+
+    while (appendToQueue)
+    {
+        QueueEntry entry = {};
+        uint64_t address = 0;
+        if (!entriesToAppend.pop_front(entry, address))
+        {
+            break;
+        }
+
+        if (!AppendEntry(entry, address))
+        {
+            entry.portContext->Release();
+            m_queueList->free_node(address);
+        }
+    }
 
 #ifdef HC_UNITTEST_API
     // Test hook: let unit tests enqueue a sibling delayed callback while this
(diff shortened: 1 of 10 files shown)


Source/Task/TaskQueue.cpp
```diff
@@ -1071,12 +1118,30 @@ bool TaskQueuePortImpl::ScheduleNextPendingCallback(
         return false;
     });
 
+    // Publish the ready entries after the pending-list walk finishes. This
+    // keeps queue wakeups and threadpool submissions out of remove_if's
+    // critical section while still preserving one ready-queue append per due
+    // timer callback.
```
jhugard: This comment is slightly stale and was based on an intermediate fix, rather than what actually landed. 
A better read would be:
```
// Publish the ready entries after the pending-list walk finishes. This
// keeps queue wakeups and threadpool submissions out of remove_if's
// critical section while still publishing each promoted ready entry
// individually once the sweep is complete.
```

jhugard:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/libHttpClient/pull/972#pullrequestreview-4255940433)

Include a fix for the following low confidence discovery from the earlier copilot review:
> Source/Task/iOS/ios_WaitTimer.mm:63
>
> ios_WaitTimer.mm: WaitTimerImpl::Start converts the remaining time to std::chrono::milliseconds (truncating toward 0) and then schedules NSTimer using that truncated value. When <1ms remains this can schedule an immediate/negative interval, causing repeated early timer fires and unnecessary churn until the steady_clock due time actually arrives. Clamp the interval to >= 0 and use ceiling rounding to avoid tight re-arm loops near the deadline.
```
    // NSTimer consumes a relative interval, so convert the stored steady-clock
    // deadline back into a relative delay right before arming it.
    auto timePoint = DeadlineFromDueTime(dueTime) - Clock::now();
    auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(timePoint);

    m_timer = [NSTimer scheduledTimerWithTimeInterval:ms.count() / 1000.0
                                                target:m_target
```

jhugard:
I am closing this PR in favor of a clean one, to remove the copilot "fix" spam.  While I agree that the tests need to be tighter post-fix, copilot choose to allow some clock jitter.  I would prefer to make them absolute to enforce the contract of "callback never fired before delayMS", rather than allow any jitter at all.

brianpepin:
James add me as a reviewer when you next submit. I looked this over and it looks good.

Issue #974, opened 2026-05-11
Linux STL XTaskQueue wait timer teardown can race in-flight delayed callback dispatch


jhugard:
## Summary

The Linux STL wait-timer backend allows delayed callback teardown to race the
worker thread after a due timer entry has already been popped from the heap.
The implementation stores raw `WaitTimerImpl*` pointers in the timer heap,
unlocks the queue, and then invokes the callback through that raw pointer.

If teardown destroyed the underlying timer object in that window, the worker
could invoke the callback through a freed timer object. In the current
implementation, `WaitTimer::Terminate()` funnels through `Cancel()` and then
destroys the impl, but `Cancel()` can no longer neutralize an entry that has
already been popped from the heap.

This is a lifetime and teardown correctness bug. The external symptom is
typically a crash or other undefined behavior during queue teardown,
especially when delayed callbacks run on immediate or otherwise synchronous
dispatch paths.

## Public API affected

```c
STDAPI XTaskQueueSubmitDelayedCallback(
    _In_ XTaskQueueHandle queue,
    _In_ XTaskQueuePort port,
    _In_ uint32_t delayMs,
    _In_opt_ void* callbackContext,
    _In_ XTaskQueueCallback* callback);

STDAPI XTaskQueueTerminate(
    _In_ XTaskQueueHandle queue,
    _In_ bool wait,
    _In_opt_ void* callbackContext,
    _In_opt_ XTaskQueueCallback* callback);
```

No signature change is required. The issue is an internal lifetime race in the
Linux STL wait-timer backend used to service delayed callbacks.

## Expected behavior

Once delayed callback teardown begins, the timer backend must either:

- prevent the callback from being dispatched at all, or
- wait for any already-started dispatch to quiesce before destroying the timer
  object and its callback context.

At no point should the worker thread retain a raw timer pointer that can outlive
the owning timer object.

## Actual behavior

The current STL backend allows the following race:

1. The worker thread popped a due timer entry containing a raw
   `WaitTimerImpl*`.
2. The worker released the queue lock before invoking the callback.
3. Another thread began teardown; in the current implementation, `WaitTimer::Terminate()` calls `Cancel()` and then destroys the timer object.
4. The worker resumed and invoked the callback through the stale raw pointer.

This could produce use-after-free behavior during delayed callback dispatch.
The risk is highest when callback execution can synchronously trigger teardown,
such as immediate-port usage or callback-driven queue destruction.

## Reproduction conditions

[...]

Issue #973, opened 2026-05-11
XTaskQueueSubmitDelayedCallback can lose its wake and strand delayed work until unrelated activity occurs


jhugard:
## Summary

Callbacks submitted through `XTaskQueueSubmitDelayedCallback` can remain
pending well past their requested delay, or indefinitely until unrelated
queue activity happens to wake the delayed port.

The failure mode here is a lost wake: the callback is never re-armed after a
same-port interleaving leaves the delayed queue with pending future work but
no active wake source.

This affects any code that relies on delayed callbacks for timeouts, retry
back-off, debouncing, or deferred continuations. In consumer code the symptom is
that a timeout or deferred action never fires when it should, even though no
cancellation occurred.

## Public API affected

```c
STDAPI XTaskQueueSubmitDelayedCallback(
    _In_ XTaskQueueHandle queue,
    _In_ XTaskQueuePort port,
    _In_ uint32_t delayMs,
    _In_opt_ void* callbackContext,
    _In_ XTaskQueueCallback* callback);
```

No signature change is required. The issue is a runtime violation of the
existing guarantee that delayed work will become dispatchable once its deadline
has elapsed.

## Expected behavior

A callback submitted with `delayMs = N` should eventually become dispatchable
once at least `N` milliseconds have elapsed, without requiring any unrelated
queue submission, timer retarget, or termination event to wake the delayed port.

That should hold even when:

- another delayed callback on the same port is being promoted concurrently
- the pending delayed list becomes temporarily empty during promotion
- a new future delayed callback is queued immediately after that empty sweep

## Actual behavior

A same-port lost-wake race can strand delayed work:

1. The delayed-callback promotion path sweeps the pending list and concludes
   there is no next future item.
2. Before it publishes the empty state by clearing `m_timerDue`, another thread
   queues a future delayed callback on the same port.
3. The new submitter observes the stale armed due time, decides it does not need
   to retarget the timer, and returns.
4. The sweep then clears `m_timerDue` to `UINT64_MAX`, leaving the new pending
   entry with no timer armed.

At that point the delayed callback remains pending until some unrelated later
activity happens to wake the queue, or forever if no such wake occurs.

## Reproduction conditions

The race is timing-sensitive and becomes easier to trigger under:

- same-port delayed callback traffic where one callback is being promoted while
  another future callback is queued concurrently
- workloads that repeatedly queue short delayed continuations on a manual or
  serialized port
- test or production environments where no unrelated later delayed submission is
[...]

PR #975, opened 2026-05-12
Fix XTaskQueue delayed-callback timing and wait-timer teardown races

jhugard:
## Summary

This PR fixes several correctness bugs in the `XTaskQueueSubmitDelayedCallback`
pipeline and the Linux STL wait-timer backend so delayed callbacks are neither
dispatched before their requested delay has elapsed nor stranded waiting for an
unrelated wake, and STL timer teardown can no longer race an in-flight callback
dispatch. See issues #971, #973, and #974 for additional detail.

The public `XTaskQueue` API surface is unchanged. The fixes are internal to the
delayed-callback scheduling path and timer backends, but they matter to any
consumer relying on timed waits, deferred continuations, retry back-off,
timeout enforcement, or queue teardown.

This branch also includes a small iOS backend cleanup identified during the
same audit: fractional-millisecond delays now round up instead of down, and the
iOS wait timer now routes destruction through `WaitTimer::Terminate()` for
consistency with the shared backend lifecycle. Unlike the STL change, however,
this iOS work does not add new teardown-quiescing logic.

## Bugs addressed

This PR addresses five distinct correctness issues in delayed-callback
scheduling and the STL backend, and also includes one small iOS cleanup:

1. **Termination early-promotion**. `XTaskQueueTerminate` on a composite queue
   could disturb shared delayed-port state and make a sibling queue's not-yet-
   due callback appear ready immediately.
2. **Stale timer notification**. An OS timer callback queued for an earlier
   deadline could arrive after the timer had already been re-armed for a later
   entry. The old code treated any timer fire as proof that the current armed
   deadline had elapsed.
3. **Equal-deadline aliasing**. The old logic matched pending entries by exact
   `uint64_t` timestamp instead of by whether the deadline had actually passed.
   Two entries with near-identical due times could see one promoted and the
   other stranded until the next unrelated timer fire.
4. **Same-port empty-sweep lost wake**. If a future delayed callback was queued
   on the same port while the dispatcher was clearing `m_timerDue` after an
   empty sweep, the new work could strand itself with no timer armed.
5. **STL timer teardown race**. The old Linux STL backend could pop a raw timer
   pointer, drop the queue lock, and then invoke the callback after
   `WaitTimer::Terminate()` had already destroyed the owning timer object.
6. **iOS timer rounding and lifecycle cleanup**. The iOS backend truncated
   fractional-millisecond delays and now routes destruction through
   `WaitTimer::Terminate()` for consistency with the shared `WaitTimer` lifecycle.

Across the five core correctness fixes, the external symptom is the same: a
delayed callback becomes runnable at the wrong time, either too early, only
after unrelated later work happens to wake the queue, or after teardown has
already started. The iOS cleanup is narrower and is limited to due-time
conversion, rounding, and lifecycle consistency.

### Delayed callback scheduling (`TaskQueue.cpp`)

- `PromoteReadyPendingCallbacks` sweeps all pending entries whose
  `enqueueTime <= now`, promotes every ready callback, and then re-arms the
  timer for the next future deadline. This replaces the old single-entry,
  exact-timestamp flow.
- `SubmitPendingCallback` now treats an early timer fire as a stale or early
  notification and retries against the currently armed due time instead of
  silently returning.
- The same-port path performs a rescue pass after clearing `m_timerDue` to
  `UINT64_MAX`, using a loop instead of recursive re-entry, which prevents the
  empty-sweep lost-wake race from leaving a newly queued delayed callback
  without an armed timer.
- `CancelPendingEntries` no longer mutates shared timer state during
  termination. It gathers canceled entries first and only publishes them after
  the pending-list traversal completes.

### Timer backends (`WaitTimer_win32.cpp`, `WaitTimer_stl.cpp`, `ios_WaitTimer.mm`)

- The backends use monotonic `steady_clock`-based due times instead of mixing
  absolute wall-clock and relative timers.
- Win32 and STL retain the existing one-shot timer model, but now convert the
  stored due time back to a relative wait only when arming the platform timer.
- The STL backend now keeps callback context in shared per-timer state,
  generation-tags heap entries instead of null-scanning the queue on every
  re-arm, and uses an explicit quiescing `Terminate()` path so teardown can
  wait for in-flight dispatch safely without leaking the timer queue for process
  lifetime.
- iOS now uses `std::chrono::ceil<std::chrono::milliseconds>` instead of
  truncating with `duration_cast<std::chrono::milliseconds>`.
- The iOS backend now implements `WaitTimer::Terminate()`, and its destructor
  routes through that entry point so timer cleanup follows the shared
  `WaitTimer` lifecycle shape. This is a narrow consistency cleanup, not the
  same quiescing teardown hardening used by STL.

### Thread pool (`ThreadPool_stl.cpp`)

- `SubmitWork` now uses `notify_one` instead of `notify_all` when waking a
  worker thread. Each queued call issues its own wake, so thundering-herd wakes
  on every submit are unnecessary.

### Test hooks and regression coverage

- Unit-test-only hooks are compiled only for unit-test builds (`HC_UNITTEST_API`).
- Commits are layered red-then-green so a reviewer can walk from failing repro
  to passing fix for each bug.

## Regression coverage

Deterministic coverage now includes:

- `VerifyDelayedCallbackTimerRaceOnManualQueue`
- `VerifyFutureDelayedCallbackQueuedDuringEmptySweepDoesNotStall`
- `VerifyTerminationDoesNotEarlyPromoteSiblingDelayedCallback`
- `VerifyStaleDelayedCallbackDoesNotEarlyPromoteNextPendingEntry`

Together these cover the original stale callback repro, the shared-port
termination bug, the stale-timer retargeting bug, and the same-port lost-wake
race. Correctness and hardening are further exercised by a downstream Win32
and Linux coroutines test suite and the benchmark validation called out below.

## Public API surface

- No public `XTaskQueue` signatures change.
- `XTaskQueueSubmitDelayedCallback` now correctly preserves its not-before-
  deadline behavior under shared delayed ports, concurrent termination, timer
  retargeting, and same-port requeue races.
- `WaitTimer::Terminate()` remains an internal backend concern. The STL fix only
  changes the internal lifetime model so delayed callbacks cannot outlive their
  timer object during teardown.
- `XTaskQueueTerminate` is unchanged from the caller's perspective: callbacks on
  the terminated queue still complete as canceled. The fix only stops that
  termination from affecting sibling queues.

## Validation

`libHttpClient` Windows unit test suite:
- Added regression tests to expose the bugs fixed in this PR
- Full TAEF suite passed (`87/87`)

Additional downstream integration validation:
- Exercised through an internal coroutines library that uses XTaskQueue
  delegate queues and delayed callbacks to implement sleep, timeout, and
  cancellation behavior.
- Also exercised through an internal WebSocket scale/throughput benchmark built
  on that coroutine layer.
- Repeated Windows and Linux soak and end-to-end runs remained stable after
  the fix, including workloads that had previously exposed the original
  shutdown hang.

Not performed:

- iOS build or runtime validation. The iOS cleanup was code-audited only; it
  was not exercised on-device or in an iOS CI environment.
- Performance regression testing. The unfixed code path is unstable under
  sustained load — the timer bugs cause crashes during connection scaling and
  throttle delayed-callback throughput to polling frequency — so no meaningful
  pre-fix performance baseline exists for comparison.

## Scope

- Win32 and STL carry the substantive timer correctness fixes; iOS receives a
  narrow due-time and lifecycle cleanup.
- The change is scoped to delayed-callback scheduling and teardown correctness.
  Immediate dispatch and unrelated task-queue behavior are unchanged.

## Review focus

- The promote-all-ready sweep and its `enqueueTime <= now` criterion.
- The stale-timer retry path in `SubmitPendingCallback`.
- The same-port rescue path after `m_timerDue` is cleared.
- The decision to keep shared timer state intact during termination and to let
  STL teardown quiesce in-flight dispatch instead of relying on raw timer
  pointers.
- The iOS due-time conversion, ceil rounding, and narrow lifecycle plumbing
  through `WaitTimer::Terminate()`.

## Potential follow-up: iOS hardening

The current PR deliberately stops short of a full iOS timer hardening pass.
The iOS changes here are limited to monotonic due-time conversion, ceiling
rounding, and routing destruction through `WaitTimer::Terminate()`; they do not
attempt to add the shared-state teardown quiescing used by the STL backend.

If we decide to pursue that Apple-platform hardening in a follow-up, the likely
shape is:

- Replace the current `NSTimer` plus raw `WaitTimerImpl*` `userInfo` handoff
  with a one-shot `dispatch_source_t` timer or another non-run-loop primitive.
- Keep callback payload and teardown bookkeeping in shared state so teardown can
  block new dispatch and wait for any in-flight callback to quiesce before the
  timer object is destroyed.
- Preserve one-shot semantics while ensuring canceled or superseded arms cannot
  dispatch later, for example by per-arm source replacement or generation
  tagging.
- Validate callback-thread behavior and performance on macOS/iOS before landing
  it, since moving away from `NSTimer` would change both the timer primitive and
  the execution context.

That would be a reasonable direction if we later decide to fix all timer
backends to the same lifetime standard, but it is intentionally out of scope
for this PR because we do not currently have Apple-platform correctness or
performance validation in hand.

### DIFF
--- a/Source/Task/TaskQueue.cpp
+++ b/Source/Task/TaskQueue.cpp
@@ -2363,4 +2363,21 @@ STDAPI XTaskQueueSetTestHooks(
     aq->SetTestHooks(hooks);
     return S_OK;
 }
+
+STDAPI XTaskQueueSubmitPendingCallbackForTests(
+    _In_ XTaskQueueHandle queue,
+    _In_ XTaskQueuePort port
+    ) noexcept
+{
+    referenced_ptr<ITaskQueue> aq(GetQueue(queue));
+    RETURN_HR_IF(E_GAMERUNTIME_INVALID_HANDLE, aq == nullptr);
+
+    referenced_ptr<ITaskQueuePortContext> portContext;
+    RETURN_IF_FAILED(aq->GetPortContext(port, portContext.address_of()));
+
+    auto* portImpl = static_cast<TaskQueuePortImpl*>(portContext->GetPort());
+    portImpl->SubmitPendingCallbackForTests();
+    return S_OK;
+}
 #endif
+
--- a/Source/Task/TaskQueue.cpp
+++ b/Source/Task/TaskQueue.cpp
@@ -1118,6 +1118,7 @@ bool TaskQueuePortImpl::ScheduleNextPendingCallback(
             // See VerifyDelayedCallbackTimerRaceOnManualQueue for full
             // analysis. The test hook here allows unit tests to verify
             // there is no race.
+#ifdef HC_UNITTEST_API
             m_attachedContexts.Visit([&](ITaskQueuePortContext* portContext)
             {
                 auto hooks = portContext->GetQueue()->GetTestHooks();
@@ -1128,6 +1129,7 @@ bool TaskQueuePortImpl::ScheduleNextPendingCallback(
                         noDueTime);
                 }
             });
+#endif
         }
     }
 
@@ -2341,6 +2343,7 @@ STDAPI_(bool) XTaskQueueUninitialize(
     return ApiRefs::WaitZeroRefs(timeoutMilliseconds);
 }
 
+#ifdef HC_UNITTEST_API
 /// <summary>
 /// Sets or clears test hooks on a task queue.
 /// </summary>
@@ -2354,3 +2357,4 @@ STDAPI XTaskQueueSetTestHooks(
     aq->SetTestHooks(hooks);
     return S_OK;
 }
+#endif
--- a/Source/Task/TaskQueue.cpp
+++ b/Source/Task/TaskQueue.cpp
@@ -362,7 +362,10 @@ HRESULT __stdcall TaskQueuePortImpl::QueueItem(
     }
     else
     {
-        entry.enqueueTime = m_timer.GetAbsoluteTime(waitMs);
+        // Delayed callbacks are ordered by a monotonic due time so stale timer
+        // callbacks and wall-clock adjustments cannot make one pending entry
+        // masquerade as another.
+        entry.enqueueTime = m_timer.GetDueTime(waitMs);
         RETURN_HR_IF(E_OUTOFMEMORY, !m_pendingList->push_back(entry));
 
         // If the entry's enqueue time is < our current time,
@@ -959,12 +962,17 @@ void TaskQueuePortImpl::CancelPendingEntries(
     // share this port's delayed-callback timer state, so leave m_timer and
     // m_timerDue alone; if we removed the armed earliest entry, the existing
     // timer simply takes one blank fire and re-arms for the next real item.
+    LocklessQueue<QueueEntry> entriesToAppend(*m_queueList.get());
 
     m_pendingList->remove_if([&](auto& entry, auto address)
     {
         if (entry.portContext == portContext)
         {
-            if (!appendToQueue || !AppendEntry(entry, address))
+            if (appendToQueue)
+            {
+                entriesToAppend.push_back(std::move(entry), address);
+            }
+            else
             {
                 entry.portContext->Release();
                 m_pendingList->free_node(address);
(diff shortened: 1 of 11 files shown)


brianpepin (approved):
This change looks good and I appreciate the addl testing.

PR #978, opened 2026-05-15
iOS build via CMake

### LINKED ISSUE #727: No way to consume ios/osx versions from CMake
I noticed it appears that the `CMakeLists.txt` isn't maintained when making changes to the `xcodeproj`. Shouldn't the `CMakeLists.txt` be modified to generate it and if so, would the maintainers be open to a PR making this change?

I think the workaround otherwise would be to `execute_process(xcodebuild...)` and link to the resulting library from CMake.

pablo-msft:
Closes #727 

- Add CMake to build libHttpClient static lib for iOS
- Add said CMake to iOS pipeline and updated out-of-date tooling

### DIFF
--- a/Build/libHttpClient.iOS.CMake/CMakeLists.txt
+++ b/Build/libHttpClient.iOS.CMake/CMakeLists.txt
@@ -48,6 +48,7 @@ set(COMMON_SOURCE_FILES
 set(IOS_SOURCE_FILES
     # Task
     "${PATH_TO_ROOT}/Source/Task/ThreadPool_stl.cpp"
+    "${PATH_TO_ROOT}/Source/Task/WaitTimer_stl.cpp"
     "${PATH_TO_ROOT}/Source/Task/iOS/ios_WaitTimer.mm"
     "${PATH_TO_ROOT}/Source/Task/iOS/ios_WaitTimer_target.mm"
     "${PATH_TO_ROOT}/Source/Task/iOS/ios_WaitTimer_target.h"
--- a/Build/libHttpClient.iOS.CMake/CMakeLists.txt
+++ b/Build/libHttpClient.iOS.CMake/CMakeLists.txt
@@ -95,5 +95,20 @@ target_include_directories(
     "${COMMON_INCLUDE_DIRS}"
     )
 
+set_target_properties(
+    "${PROJECT_NAME}"
+    PROPERTIES
+    CXX_STANDARD 17
+    CXX_STANDARD_REQUIRED ON
+    )
+
+set(IOS_MM_SOURCE_FILES
+    "${PATH_TO_ROOT}/Source/Task/iOS/ios_WaitTimer.mm"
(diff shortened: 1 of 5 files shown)


Build/libHttpClient.iOS.CMake/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+    "${WEBSOCKET_SOURCE_FILES}"
+    "${MOCK_SOURCE_FILES}"
+    "${HTTP_SOURCE_FILES}"
+    "${LOGGER_SOURCE_FILES}"
+    "${TASK_SOURCE_FILES}"
+    )
+
+set(IOS_SOURCE_FILES
+    # Task
+    "${PATH_TO_ROOT}/Source/Task/ThreadPool_stl.cpp"
```
jasonsandlin, line 50: Suggestion:

`WaitTimer_stl.cpp` and `ios_WaitTimer.mm` both define `OS::WaitTimer` and `OS::WaitTimerImpl` methods (Initialize, Start, Cancel, etc.). Including both in this target will cause duplicate symbol linker errors.

**Suggested fix:** Remove `WaitTimer_stl.cpp` from `IOS_SOURCE_FILES` — `ios_WaitTimer.mm` already provides the complete WaitTimer implementation for iOS.

jasonsandlin (approved):

Build/libHttpClient.iOS.CMake/CMakeLists.txt
pablo-msft: Attempted suggested solution, which generated this error:
```
ios_WaitTimerImpl.h:19:53: error: unknown type name 'WaitTimerCallback'; did you mean 'OS::WaitTimerCallback'? 
19 | HRESULT Initialize(_In_opt_ void* context, _In_ WaitTimerCallback* callback); 
| ^~~~~~~~~~~~~~~~~ 
[etc.]
```
Per discussion, reverted suggestion.

pablo-msft:
Ready for re-review.

Build/libHttpClient.iOS.CMake/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
+cmake_minimum_required(VERSION 3.6)
```
jasonsandlin: `cmake_minimum_required(VERSION 3.6)` may be too low — `CXX_STANDARD 17` requires CMake >= 3.8, and `CMAKE_SYSTEM_NAME=iOS` cross-compilation support requires >= 3.14. Consider bumping to:

```cmake
cmake_minimum_required(VERSION 3.14)
```

jasonsandlin (approved):

Build/libHttpClient.iOS.CMake/CMakeLists.txt
pablo-msft: Good catch. I will bump in this iOS file but not in the other cmake files in this repo.

jasonsandlin (approved):

PR #981, opened 2026-05-20
Split iOS check into three jobs

pablo-msft:
The iOS stages would take 30-40 minutes, whereas all other checks took fewer than 10 minutes. Split into iOS, macOS, and cmake builds jobs.


jasonsandlin (approved):
Clean split — all build steps preserved, parameters wired correctly, jobs now run in parallel.

PR #994, opened 2026-06-16
Fix XTaskQueue shared-timer starvation when a composite terminates with a pending delayed callback

rgomez391:
## Summary
Fixes a delayed-callback starvation in `XTaskQueue` that can occur when a composite queue sharing a port's Work stream is terminated while it still owns the earliest-due pending delayed callback. After such a terminate, the shared delayed-callback timer can be left armed for a deadline whose entry was just removed. On the STL `WaitTimer` backend (non-Microsoft platforms — Linux, Android, iOS, etc.) that orphaned fire can be lost, after which the remaining pending delayed callbacks never run and the queue effectively stops dispatching delayed work.

## Root cause
`TaskQueuePortImpl::CancelPendingEntries` intentionally leaves `m_timer` / `m_timerDue` untouched on terminate (sibling delegates share the port's timer state), relying on a "blank fire" of the now-orphaned deadline to re-arm for the next real item. That assumption holds on the Win32 threadpool timer backend but is fragile on the STL timer backend, where the orphaned notification can be dropped — stranding every remaining pending entry.

## Fix
Add `TaskQueuePortImpl::RearmTimerForEarliestPending()` and call it at the end of `CancelPendingEntries`. After pending entries are removed, it scans for the earliest surviving pending entry and re-points the shared timer at it:

- earlier than the armed deadline → `ArmTimerIfEarlier` (min-wins);
- later (the armed deadline belonged to a removed entry) → `ArmTimerForNextPendingDueTime` (the same verified helper the fire path uses, so a concurrent earlier publish is handled safely);
- none remain → CAS `m_timerDue` to `UINT64_MAX` (CAS only, never `Cancel()`, to avoid racing concurrent `Start` calls — matching the existing no-next-item branch).

## Test coverage
The regression only reproduces on the STL timer backend, which the existing Windows unit-test lanes don't exercise — that's why it wasn't caught. This PR adds a regression guard that runs on **both** backends from a single shared scenario:

- `Tests/Shared/CompositeQueueStarvationScenario.h` — a platform-neutral scenario (public `XTaskQueue` API + the standard library only): many concurrent submitters create composites over one shared Work port and drive self-resubmitting delayed-poll loops with tiny (1–2 ms) delays, while a canceller terminates a subset of composites that still have a delayed callback pending and pump threads dispatch the Work port. It returns failure if any request fails to complete (starvation).
- `Tests/UnitTests/Tests/TaskQueueTests.cpp` — `VerifyCompositeQueueDelayedCallbackStarvation` now calls the shared scenario (Win32 backend).
- `Tests/TaskQueueStarvation/TaskQueueStarvationRepro.cpp` + a CTest target in `Build/libHttpClient.Linux/CMakeLists.txt` — runs the same scenario on the STL backend, wired into the Linux CI lane in `Utilities/Pipelines/Tasks/linux-build.yml`.

Before the fix the scenario starves (times out) on the STL backend; after the fix it passes on both backends, and the full `TaskQueueTests` suite (30 tests, incl. the existing delayed-callback/termination regression tests) passes on both backends.

## Validation
- STL backend: scenario fails (~15 s timeout) without the fix; passes (~3–4 s) with it, consistently across repeated runs.
- Win32 backend + STL backend: full `TaskQueueTests` 30/30.
- Shared scenario header compiles standalone as an isolated C++17 translation unit.

### DIFF
--- a/Tests/UnitTests/Tests/TaskQueueTests.cpp
+++ b/Tests/UnitTests/Tests/TaskQueueTests.cpp
@@ -6,7 +6,7 @@
 #include "CallbackThunk.h"
 #include "PumpedTaskQueue.h"
 #include "XTaskQueuePriv.h"
-#include "../../Shared/CompositeQueueStarvationScenario.h"
+#include "CompositeQueueStarvationScenario.h"
 
 #define TEST_CLASS_OWNER L"brianpe"
 
--- a/Tests/UnitTests/Tests/TaskQueueTests.cpp
+++ b/Tests/UnitTests/Tests/TaskQueueTests.cpp
@@ -6,6 +6,7 @@
 #include "CallbackThunk.h"
 #include "PumpedTaskQueue.h"
 #include "XTaskQueuePriv.h"
+#include "../../Shared/CompositeQueueStarvationScenario.h"
 
 #define TEST_CLASS_OWNER L"brianpe"
 
@@ -98,6 +99,11 @@ struct HandleCloser
 };
 
 typedef AutoHandleWrapper<HANDLE, HandleCloser> AutoHandle;
+
+// The composite-queue delayed-callback starvation scenario lives in
+// Tests/Shared/CompositeQueueStarvationScenario.h so the exact same scenario is
+// exercised here (Win32 WaitTimer backend) and by the standalone Linux repro
+// binary (STL WaitTimer backend, where the regression manifests).
 
 DEFINE_TEST_CLASS(TaskQueueTests)
 {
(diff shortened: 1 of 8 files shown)


Tests/UnitTests/Tests/TaskQueueTests.cpp
```diff
@@ -6,6 +6,7 @@
 #include "CallbackThunk.h"
 #include "PumpedTaskQueue.h"
 #include "XTaskQueuePriv.h"
+#include "../../Shared/CompositeQueueStarvationScenario.h"
```
brianpepin: This code is shared with the Windows OS repo and it has a different branch structure. It would be great if the path to shared includes was #defined somewhere, or if the shared include path was added to the project include path so the structure wasn't stored here. I have a script that just copies the files to the right paths but I like to keep the core files identical so it's easy to move back and forth. Project files and pch are replicated in each repo so this is a handy place to introduce variance.

brianpepin (changes requested):
All this looks good. The only thing I would request is to somehow abstract the include path for CompositeQueueStarvationSequence.h in TaskQueueTests.cpp so I can more easily take this file verbatim over to the Windows repo. Either add ..\..\Shared to the project include path or provide a #define symbol I can redefine in Windows.

Tests/UnitTests/Tests/TaskQueueTests.cpp
rgomez391: Thanks Brian, this is effectively fixing a regression introduced by recent changes that Panic Button was hitting on their title. I will update the PR to address this.

brianpepin (approved):
Thanks. Ship it!

Tests/UnitTests/Tests/TaskQueueTests.cpp
rgomez391: Alright, moved the shared-include path out of the core file so TaskQueueTests.cpp stays identical across repos.

TaskQueueTests.cpp (and the Linux repro TaskQueueStarvationRepro.cpp) now use a bare `#include "CompositeQueueStarvationScenario.h"` with no embedded relative path.
The path variance lives in the per-repo project files instead:
- `libHttpClient.UnitTest.vcxitems` → added `$(HCRoot)\Tests\Shared` to `AdditionalIncludeDirectories` (next to the existing Support entries).
- `CMakeLists.txt` → added `${PATH_TO_ROOT}/Tests/Shared` to the repro target's include dirs.

PR #996, opened 2026-06-18
TaskQueue: use unnamed condition variables as an option on NDA platform

PushpadantK:
Add DefaultUnnamedConditionVariable wrappers and switch task queue, async, wait timer, and thread pool synchronization objects to unnamed condition-variable variants under HC_USE_UNNAMED_MUTEX.


rgomez391 (approved):
Let's remove the Platform specific title and description from the PR. For the actual changes LGTM :)

PR #998, opened 2026-06-23
Fix STL wait-timer strand: restore pre-#975 backend (close #973)

### LINKED ISSUE #973: XTaskQueueSubmitDelayedCallback can lose its wake and strand delayed work until unrelated activity occurs

rgomez391:
### Summary

This rolls back the **STL wait-timer backend rewrite** from #975 (WaitTimer_stl.cpp) while **keeping** all of the timer-arming improvements from #975, #980, and #983 in TaskQueue.cpp. The #975 backend rewrite introduced a delayed-callback strand that manifests as a **permanent hang** on platforms that compile the STL backend (Linux, Apple, and other non-Windows platforms), reproduced on-device as an infinite sign-in / inventory hang in a shipping title.

The change is scoped to a single source file (`WaitTimer_stl.cpp`) plus a regression test, and is validated on-device.

### Background: the #973 lineage

This is a continuation of the work on issue **#973** ("a delayed callback becomes runnable only after unrelated later work wakes the queue"):

- **#975** — *Fix XTaskQueue delayed-callback timing and wait-timer teardown races.* Reworked both the TaskQueue.cpp arming logic (wall-clock → monotonic due times) **and** rewrote the STL timer backend (WaitTimer_stl.cpp), replacing the O(N) pointer-keyed cancellation scan with an O(log N) **per-timer generation** scheme.
- **#980** — *Update TaskQueue with changes from Windows repo.* Added post-`Start()` verification in `SubmitPendingCallbacks`.
- **#983** — *Centralize timer-arming to close TOCTOU race.* Centralized the CAS+Start+verify logic into three helpers (`ArmTimerIfEarlier`, `ArmTimerForNextPendingDueTime`, `RearmTimerIfDueTimeUnchanged`).

The TaskQueue.cpp arming work in all three PRs is **correct and is retained by this change.** The regression is specifically in the **#975 STL backend rewrite**.

### The bug

The #975 STL backend's generation scheme can **drop a due callback**:

1. `TimerQueue::Set` (called by `Start()`) bumps the timer's generation and pushes a *new* heap entry; the worker is expected to discard the superseded entry on pop when `entry.Generation != state->Generation()`.
2. But a `Start()` that races the worker **between its `Peek()` of the earliest entry and its dispatch of that entry** bumps the generation first.
3. When the worker then pops that entry, it reads a now-stale generation and **discards it instead of firing it**. The re-pushed entry carries a *new* deadline, so the original due callback is never delivered.
4. With the entry gone, the worker sleeps on the next (later or absent) deadline and goes idle.

On a **single-port manual `XTaskQueue`** this is unrecoverable. The platform HTTP provider re-arms its poll via `XTaskQueueSubmitDelayedCallback(... Work ...)`; once that re-poll is dropped, there is no independent traffic to wake the queue and sweep the pending list again. The call freezes permanently.

Observed on-device: the delayed re-poll sits in `m_pendingList` while the timer thread waits in `_Cnd_timedwait` and `m_timerDue == UINT64_MAX` (idle) — i.e. a pending entry with no armed timer, which should be impossible.

### Why the arming-layer fixes (#975/#980/#983) could not address it

- All three hardened the `CAS → Start` bookkeeping on `m_timerDue` in `TaskQueuePortImpl`. That logic is correct, but it rests on the assumption that **`Start()` reliably arms the backend for the published deadline.** The generation backend violates that assumption by discarding the entry *after* `Start()`, so no amount of arming-layer post-`Start` verification can recover it.
- The defect only affects WaitTimer_stl.cpp. **Win32/UWP use `WaitTimer_win32.cpp`** (OS `CreateThreadpoolTimer`, no generation scheme) and are unaffected. Critically, the libHttpClient **Windows unit-test project compiles only the Win32 backend**, so the Windows suite stayed green when #975/#980/#983 merged. The Linux CI lane that does exercise the STL backend (`taskqueue-starvation-linux`) was added *after* these PRs, and its existing scenario has rescue traffic that masks this particular strand (see Follow-up). This test-coverage gap is the systemic reason the regression shipped.

### The fix

Restore the proven pre-#975 STL backend algorithm (libHttpClient 2.3.1, commit `0fa5f24`):

- **Pointer-keyed cancellation** and an **unconditional `notify_all` on every `Set`**, which cannot discard a due entry.
- Bridged to the post-#975 public `WaitTimer` API (`GetCurrentTime` / `GetDueTime` / `Start(dueTime)`) so TaskQueue.cpp is unchanged.
- Clock pinned to **`steady_clock`** (the 2.3.1 code used `high_resolution_clock`, which is `steady_clock` on libc++ but wall-clock `system_clock` on libstdc++). Pinning keeps delayed-callback ordering **monotonic on every platform**, preserving #975's monotonic-time intent.

Only the STL timer backend is reverted; the TaskQueue.cpp arming improvements from #975/#980/#983 are kept unchanged.

### Validation

- **On-device, in a shipping title:** the reproducing infinite sign-in / inventory hang no longer occurs with this backend combined with the current TaskQueue.cpp arming logic. Confirmed by the partner across multiple runs.
- Root cause was isolated by a controlled A/B: `new TaskQueue.cpp + new backend` hangs; `new TaskQueue.cpp + restored backend` does not. The only variable is WaitTimer_stl.cpp.

### Regression test (included in this PR)

A Linux CI test lane that exercises the STL backend **already exists** — `taskqueue-starvation-linux` (`TaskQueueStarvationTests.Linux`, added alongside the composite-queue starvation guard), which runs a shared scenario in `Tests/Shared/` against `WaitTimer_stl.cpp`. That existing scenario keeps many concurrent submitters and terminate races running for the whole measurement window, so it always has independent traffic that can rescue a momentarily-stranded timer — which is exactly why it does not catch this particular strand.

This PR adds a complementary scenario that reproduces the failure shape directly:

- `Tests/Shared/SingleQueueDelayedPollStrandScenario.h` — a **single self-resubmitting delayed-poll loop on one manual `Work` port** (mirroring a platform HTTP provider that re-arms its poll every few ms via `XTaskQueueSubmitDelayedCallback`). A short burst of unrelated one-shot delayed callbacks runs concurrently to create timer-arming contention, then stops. After the burst drains, the poll loop is the only remaining work — with no rescue traffic, a dropped re-arm stalls it permanently, and the test fails deterministically because it requires the loop to make measurable progress after the burst ends.
- `Tests/UnitTests/Tests/TaskQueueTests.cpp` — `VerifySingleQueueDelayedPollStrand` runs it on the **Win32** backend.
- `Tests/TaskQueueStarvation/SingleQueuePollStrandRepro.cpp` + the new `taskqueue-poll-strand-linux` CTest target run it on the **STL** backend (where the regression manifests), wired into the Linux CI lane.

The scenario uses only the public `XTaskQueue` C API plus the standard library, so the same translation unit compiles on every platform. Like the existing starvation guard, the strand is a probabilistic race; the test is a stress loop (many iterations) that maximizes exposure rather than a deterministic single-shot.

### DIFF
--- a/Source/Task/WaitTimer_stl.cpp
+++ b/Source/Task/WaitTimer_stl.cpp
@@ -77,7 +77,7 @@ namespace OS
         TimerEntry Pop() noexcept;
 
         DefaultUnnamedMutex m_mutex;
-        std::condition_variable m_cv;
+        DefaultUnnamedConditionVariable m_cv;
         std::vector<TimerEntry> m_queue; // used as a heap
         std::thread m_t;
         bool m_exitThread = false;
--- a/Source/Task/WaitTimer_stl.cpp
+++ b/Source/Task/WaitTimer_stl.cpp
@@ -1,110 +1,34 @@
 #include "pch.h"
 #include "WaitTimer.h"
 
+// NOTE: This is the proven pre-#975 STL wait-timer backend (libHttpClient
+// 2.3.1 / commit 0fa5f24), restored to fix a delayed-callback strand that
+// #975's rewrite introduced. The #975 backend added a per-timer "generation"
+// scheme where every Start() invalidated the previously pushed heap entry; a
+// Start() racing the worker as it was about to dispatch the due entry caused
+// the worker to discard that entry as stale and never fire it, permanently
+// stranding the callback (observed on-device as an infinite sign-in / inventory
+// hang on a single-port manual queue with no independent traffic to rescue it).
+//
+// This backend uses the original pointer-keyed cancellation and an
+// unconditional notify on every Set, which cannot drop a due callback. The
+// public WaitTimer API matches the post-#975 surface (GetCurrentTime /
+// GetDueTime / Start(dueTime)); both time helpers and the worker share a
+// single monotonic clock so the values TaskQueue compares stay consistent and
+// are immune to wall-clock adjustments.
+//
+// The pre-#975 backend keyed this clock off std::high_resolution_clock. That
+// alias is steady_clock on some standard libraries (libc++) but system_clock
+// on others (libstdc++), which is wall-clock and not monotonic. We pin it to
+// steady_clock explicitly so the ordering and "now < dueTime" comparisons in
+// TaskQueue are monotonic on every platform that compiles this backend.
+
 using Clock = std::chrono::steady_clock;
 using Deadline = Clock::time_point;
-using TimerDuration = std::chrono::nanoseconds;
-
-namespace
-{
-    // Keep the public WaitTimer surface on a plain integer so TaskQueue can use
-    // atomics without dragging chrono types through its state. The integer still
-    // represents steady-clock time, not wall-clock time.
-    Deadline DeadlineFromDueTime(uint64_t dueTime) noexcept
-    {
-        return Deadline(std::chrono::duration_cast<Clock::duration>(TimerDuration(dueTime)));
-    }
-
-    uint64_t DueTimeFromDeadline(Deadline deadline) noexcept
-    {
-        return static_cast<uint64_t>(
-            std::chrono::duration_cast<TimerDuration>(deadline.time_since_epoch()).count());
-    }
-}
 
 namespace OS
 {
     class TimerQueue;
-
-    // Keep callback payload and teardown coordination in shared state because the
-    // Worker can still hold a due heap entry after the owning WaitTimerImpl has
-    // been canceled or destroyed.
-    class WaitTimerState
-    {
-    public:
-        WaitTimerState(_In_opt_ void* context, _In_ WaitTimerCallback* callback) noexcept
-            : m_context(context), m_callback(callback)
-        {}
-
-        uint64_t NextGeneration() noexcept
-        {
-            return ++m_generation;
-        }
-
-        uint64_t Generation() const noexcept
-        {
-            return m_generation.load(std::memory_order_acquire);
-        }
-
-        void BeginTerminate() noexcept
-        {
-            m_terminating.store(true, std::memory_order_release);
-        }
-
-        bool TryBeginDispatch() noexcept
-        {
-            if (m_terminating.load(std::memory_order_acquire))
-            {
-                return false;
-            }
-
-            std::lock_guard<std::mutex> lock{ m_lock };
-            // Re-check under the lock so teardown cannot race with the in-flight
-            // count increment and then wait forever for a dispatch we started.
-            if (m_terminating.load(std::memory_order_relaxed))
-            {
-                return false;
-            }
-
-            ++m_inFlightDispatch;
-            return true;
-        }
-
-        void EndDispatch() noexcept
-        {
-            std::lock_guard<std::mutex> lock{ m_lock };
-            ASSERT(m_inFlightDispatch != 0);
-
-            --m_inFlightDispatch;
-            if (m_inFlightDispatch == 0)
-            {
-                m_quiesced.notify_all();
-            }
-        }
-
-        void WaitForQuiesce() noexcept
-        {
-            std::unique_lock<std::mutex> lock{ m_lock };
-            m_quiesced.wait(lock, [this]() noexcept
-            {
-                return m_inFlightDispatch == 0;
-            });
-        }
-
-        void InvokeCallback() noexcept
-        {
-            m_callback(m_context);
-        }
-
-    private:
-        void* m_context;
-        WaitTimerCallback* m_callback;
-        std::atomic<uint64_t> m_generation{ 0 };
-        std::atomic<bool> m_terminating{ false };
-        DefaultUnnamedMutex m_lock;
-        DefaultUnnamedConditionVariable m_quiesced;
-        uint32_t m_inFlightDispatch = 0;
-    };
 
     class WaitTimerImpl
     {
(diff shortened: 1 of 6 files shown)


Source/Task/WaitTimer_stl.cpp
```diff
@@ -161,10 +77,9 @@ namespace OS
         TimerEntry Pop() noexcept;
 
         DefaultUnnamedMutex m_mutex;
-        DefaultUnnamedConditionVariable m_cv;
+        std::condition_variable m_cv;
```
brianpepin: Should this be DefaultUnnamedConditionVariable?

Source/Task/WaitTimer_stl.cpp
```diff
@@ -185,16 +100,7 @@ namespace OS
...
-            {
-                // Immediate-port callbacks can tear down their own timer from the
-                // Worker thread. Joining from that path would deadlock.
-                m_t.detach();
-            }
-            else
-            {
-                m_t.join();
-            }
+            m_t.join();
```
brianpepin, line 103: 975 commit had a check here so we didn't join on the worker thread, and from looking at the task queue code I think that is in fact possible. Seems very safe to add in the 975 change here to detach if it's the same ID as the thread. I'd add that back.

brianpepin (changes requested):
Left a couple of comments that I think you should decide if you want to change.

Source/Task/WaitTimer_stl.cpp
rgomez391: Hey Brian, seems like that check alone could introduce a regression and we would need to make more changes in order to make it safe, here's a quick analysis:

- **`detach` alone → use-after-free.** The restored worker captures raw `this` and relies on the dtor always `join()`ing before destruction. If we `detach` instead, the worker keeps touching `this` (its mutex/queue) after the triggering callback returns, on a freed object.
- **`detach` + `keepAlive` (what I tried) → leak + hung thread.** Adding `keepAlive = shared_from_this()` to make detach safe creates a **refcount cycle**: the worker holds a ref → `TimerQueue` refcount stays ≥1 → the dtor (the *only* place `m_exitThread` is set in the old backend) never runs → the worker never exits → it never releases its ref. The queue leaks and the timer thread runs forever. Repeated timer create/destroy cycles would leak a thread each time.
- **The reason #975's detach was safe** is that #975 *also* added `RemoveTimer()`/`AddTimer()`, which set `m_exitThread` from **outside** the dtor (when the last timer is removed) — breaking the cycle. The old backend deliberately has none of that; it relies on the worker holding **no** ref so the dtor can run and `join()`.

brianpepin: Solid analysis. Let's leave it as just a join then. The win32 code also can deadlock here, and the solution is just in client code not to close the last queue handle in the callback.

Source/Task/WaitTimer_stl.cpp
rgomez391: This is fixed now!

brianpepin (approved):
Discussed offline; we believe there is a lot of merit to the 975 change but will commit this now to unblock a studio that is hitting this bug. Will re-visit bringing back once the studio is unblocked.

PR #1000, opened 2026-07-03
Fix NetworkState submit/cleanup TOCTOU races

### LINKED ISSUE #999: NetworkState: HTTP/WebSocket submission races cleanup (TOCTOU), causing crashes or orphaned requests
## Summary

`NetworkState` (`Source/Global/NetworkState.cpp`) does not serialize network-operation
*submission* against the *start of cleanup*. When a client calls `HCHttpCallPerformAsync` or
`HCWebSocketConnectAsync` on one thread while another thread calls `HCCleanup` / `HCCleanupAsync`,
there is a time-of-check-to-time-of-use window that can:

- dereference a moved-from `m_networkState` pointer and crash, and/or
- insert a request into the cleanup tracking sets *after* cleanup has already snapshotted them,
  orphaning the request so it is never canceled or awaited and can run against a torn-down provider.

Both symptoms stem from a single root cause and are intermittent/timing-dependent.

## Affected code

- `Source/Global/global_publics.cpp` — `HCHttpCallPerformAsync`
- `Source/WebSocket/websocket_publics.cpp` — `HCWebSocketConnectAsync`
- `Source/Global/global.cpp` — `http_singleton::CleanupAsyncProvider`, `singleton_access`
- `Source/Global/NetworkState.cpp` — `HttpCallPerformAsyncProvider`,
  `WebSocketConnectAsyncProvider`, `CleanupAsyncProvider`

Any consumer that can issue HTTP/WebSocket calls concurrently with library cleanup from a
different thread is exposed (e.g. XSAPI/PlayFab-style hosts that tear down while requests may still
be in flight).

## Background

A submission goes through the public entry point, which acquires the singleton and then calls into
`NetworkState`:

```cpp
// HCHttpCallPerformAsync (Source/HTTP/httpcall_publics.cpp)
auto httpSingleton = get_http_singleton();          // (A) strong ref; may be non-null
if (nullptr == httpSingleton) return E_HC_NOT_INITIALISED;
return httpSingleton->m_networkState->HttpCallPerformAsync(call, asyncBlock);  // (B) deref + submit
```

`get_http_singleton()` uses `singleton_access_mode::get`, which returns a copy of the shared
pointer **without incrementing the singleton use count**. `XAsyncBegin` dispatches the provider
`XAsyncOp::Begin` synchronously on the calling thread, so the submission's insert into the
tracking set happens synchronously inside step (B).

Cleanup runs as:

```cpp
// http_singleton::CleanupAsyncProvider::Begin (Source/Global/global.cpp)
singleton_access(cleanup, ...);                       // detaches: resets the static singleton
NetworkState::CleanupAsync(std::move(singleton->m_networkState), ...);  // moves NetworkState out
// -> NetworkState::CleanupAsyncProvider::Begin snapshots m_activeHttpRequests under m_mutex
```

The in-code comment claimed new performs cannot enter after cleanup begins because the singleton is
detached first. That is insufficient: detaching only prevents *future* `get_http_singleton()`
callers from acquiring the singleton. A caller that already passed the null check at (A) holds a
strong reference and is completely unsynchronized against cleanup.

## Race A — request inserted after the cleanup snapshot

`HttpCallPerformAsyncProvider::Begin` inserts into `m_activeHttpRequests`, and
`CleanupAsyncProvider::Begin` snapshots that set; likewise `WebSocketConnectAsyncProvider::Begin`
inserts into `m_connectingWebSockets`. Both the insert and the snapshot take `m_mutex`, but their
ordering is unconstrained:

1. Cleanup acquires `m_mutex`, snapshots the (empty) set, sees nothing pending, and proceeds to
   schedule provider teardown.
2. The submission then acquires `m_mutex` and inserts its context.

The inserted request is now absent from the cancel snapshot and is not counted by the pending-work
check, so cleanup can tear down the provider and complete while the request is still live. The
submission is never canceled or awaited and runs against a destroyed provider — use-after-free
and/or a request that completes long after the client believed cleanup finished. The submission
`Begin` path is the only mutation of the tracking sets that does not consult cleanup state, so a
late insert bypasses the entire cancel/await handshake.

## Race B — dereference of a moved-from `m_networkState`

1. Thread T1 executes (A): `get_http_singleton()` returns non-null and T1 holds a strong reference.
2. Thread T2 runs `http_singleton::CleanupAsyncProvider::Begin`, which resets the static singleton
   and then `std::move(singleton->m_networkState)`, leaving the singleton's `m_networkState`
   `UniquePtr` null. The moved-out `NetworkState` is subsequently destroyed when provider cleanup
   completes.
3. Thread T1 executes (B): `httpSingleton->m_networkState->...` reads the moved-from (null) pointer
   and dereferences it.

This is both a null dereference and a data race on the `m_networkState` `UniquePtr` (concurrent
read on T1 and move-write on T2 with no synchronization), i.e. undefined behavior. The singleton's
`use_count > 1` cleanup gate does not help: it defers destruction of the singleton *object*, but
`m_networkState` is moved out (and the `NetworkState` destroyed) independently of that gate.

Because the move nulls `m_networkState` before the submission reaches `NetworkState`, Race B often
manifests first on the HTTP submit path; when a submission does survive the move, Race A can then
occur. They are two facets of the same missing serialization between submission and cleanup-begin.

## Impact

- Intermittent crash (null dereference / UB) during shutdown when HTTP or WebSocket calls are
  submitted concurrently with `HCCleanup`.
- Orphaned in-flight requests that outlive cleanup and run against a torn-down provider
  (use-after-free), or that never complete/cancel as the client expects.
- Timing-dependent; more likely under load or with frequent init/cleanup cycles.

jhugard:
Fixes #999

## Summary

`NetworkState` did not serialize network-operation *submission* against the *start of cleanup*.
Calling `HCHttpCallPerformAsync` / `HCWebSocketConnectAsync` on one thread while another thread
runs `HCCleanup` / `HCCleanupAsync` could crash on a moved-from `m_networkState`, or orphan a
request that lands in the cleanup tracking sets after they were snapshotted. See the linked issue
for the full analysis.

This PR closes both facets. The public API surface is unchanged, and there is no new `shared_ptr`
or per-call allocation on the submission path.

Code review also found a cleanup wake-up edge case in the admission-gate implementation: a
rejected HTTP perform still runs its `XAsyncOp::Cleanup` provider path, and if that rejected
context is treated like a tracked request it can consume cleanup's one schedule slot or double-run
provider cleanup. The final fix only lets contexts that were actually present in
`m_activeHttpRequests` wake cleanup.

## Bug addressed

Two facets of one root cause — submission is not serialized against cleanup-begin:

- **Race A (late insert).** A perform/connect whose `Begin` op acquires `m_mutex` *after*
  `CleanupAsyncProvider::Begin` took its snapshot was inserted into `m_activeHttpRequests` /
  `m_connectingWebSockets` anyway, orphaning it past cleanup: never canceled or awaited, and able
  to run against a torn-down provider.
- **Race B (moved-from `m_networkState`).** `HCHttpCallPerformAsync` dereferences
  `httpSingleton->m_networkState` while `http_singleton::CleanupAsyncProvider::Begin`
  `std::move`s it out. `get_http_singleton()` does not increment the singleton use count, so an
  in-flight caller holding a strong reference is unsynchronized against the move — a null
  dereference / data race.

The PR also includes a code-review hardening fix for the rejection path itself: an HTTP perform
that is refused after cleanup begins is never inserted into `m_activeHttpRequests`, so its later
provider cleanup must reclaim only its local context and must not wake or reschedule cleanup.

### Race A — admission gate (`NetworkState`)

- Added `m_cleanupStarted`, set under `m_mutex` in `CleanupAsyncProvider::Begin` alongside the
  existing snapshot.
- `HttpCallPerformAsyncProvider::Begin` and `WebSocketConnectAsyncProvider::Begin` now check
  `m_cleanupStarted` inside the same `m_mutex` critical section as the tracking-set insert. If
  cleanup has begun, the submission is refused with `E_HC_NOT_INITIALISED` instead of being
  inserted. Because insert and snapshot share `m_mutex` and the flag is set under that lock, a
  submission either lands before the snapshot (and is captured/canceled) or is cleanly refused —
  no gap.
- `HttpCallPerformAsyncProvider::Cleanup` now wakes cleanup only when
  `m_activeHttpRequests.erase(performContext)` actually removed a tracked request. Rejected
  performs still reclaim their async context, but cannot consume cleanup's schedule slot or cause
  provider cleanup to run twice because they were never admitted to the active set.

### Race B — keep `NetworkState` owned by the singleton

- `NetworkState::CleanupAsync` now takes a non-owning `NetworkState*` and no longer transfers
  ownership; `global.cpp` passes `singleton->m_networkState.get()` instead of `std::move`.
- Cleanup no longer destroys `NetworkState` (`HttpProviderCleanupComplete` and the `DoWork`
  failure path no longer reclaim it). The instance stays owned by the `http_singleton` and is
  destroyed with the singleton, after the singleton's existing `use_count` gate confirms no other
  references remain. An in-flight caller holding a strong singleton reference keeps `use_count`
  above 1, so it can never observe a moved-from or destroyed `m_networkState`.
- Documented that gate in `global.cpp`, including the supporting invariant that no code may hold a
  `get_http_singleton()` reference across an async wait.

### Tests

- `TestCleanupKeepsNetworkStateForInFlightSingletonRef` (Race B) — a caller holding a live
  `get_http_singleton()` reference asserts `m_networkState` stays non-null across cleanup; fails
  pre-fix when it is moved out from under that reference.
- `TestHttpPerformRejectedAfterCleanupStarted` (Race A) — a perform whose `Begin` runs after
  cleanup has begun must be refused rather than inserted into `m_activeHttpRequests`; fails pre-fix
  when it is orphaned past the snapshot. The test was strengthened during code review with a real
  unscheduled cleanup async block, so it also fails if a rejected perform consumes cleanup's one
  allowed schedule.
- The original reproductions are deterministic (they rely on `XAsyncBegin` dispatching `Begin`
  synchronously) and were committed first as failing tests, then made to pass by the fixes. The
  code-review cleanup wake-up regression was also committed red, then fixed green.
- Adds a `TestSetCleanupStarted` `HC_UNITTEST_API` seam used by the Race A test; the seam can also
  attach a cleanup async block so the rejection path is tested against realistic cleanup state.

## Design rationale

- **Why the admission gate under `m_mutex`?** The insert and the cleanup snapshot already
  serialize on `m_mutex`. Setting/reading the flag in that same critical section makes the
  admission decision atomic with the insert/snapshot, closing the window without a new lock.
- **`m_cleanupStarted` is a plain `bool`, not atomic.** Every read and write occurs under
  `m_mutex`, which provides the required mutual exclusion and ordering; an atomic would be
  redundant and would misleadingly imply lockless access. The correctness of the guard depends on
  the read and the insert being in the *same* critical section as cleanup's write+snapshot, which
  the mutex — not an atomic — provides.
- **Why gate cleanup wake-up on `erase`?** A rejected HTTP perform still owns an async context and
  still receives `XAsyncOp::Cleanup`, but it was never part of the cleanup tracking set. Only an
  admitted request can be work that cleanup is waiting for, so the cleanup wake-up is tied to the
  active-set erase actually removing an entry. This prevents a refused submit from stealing the
  cleanup async block's schedule slot or invoking provider cleanup a second time.
- **Why keep `NetworkState` owned by the singleton instead of a `shared_ptr`?** The singleton
  already has a `use_count` gate that waits for all outstanding references before destruction.
  Reusing it (by not moving `NetworkState` out early) fixes Race B with no new `shared_ptr`, no
  new atomics, and no per-call cost — versus adding an in-flight submitter counter that would
  re-implement `use_count`, touch every public entry point, and add hot-path overhead. The
  trade-off is that `NetworkState` (and its already-cleaned providers) lives until the singleton is
  destroyed — a few milliseconds longer under the gate — and a documented invariant that callers
  must not hold a singleton reference across an async wait.

## Public API surface

- No public `HC*` signatures added, removed, or changed.
- `HCHttpCallPerformAsync` / `HCWebSocketConnectAsync` behavior is unchanged in the normal case;
  when they race a concurrent cleanup they now fail deterministically with `E_HC_NOT_INITIALISED`
  rather than crashing or orphaning the request.

## Validation

- Windows x64 Debug, `libHttpClient.UnitTest.TE`:
  - The two race reproductions: red on their reproduction commits → green after the fix commits.
  - Code-review regression: `TestHttpPerformRejectedAfterCleanupStarted` red with `E_UNEXPECTED`
    when the rejected perform consumed cleanup's schedule slot → green after
    `Skip cleanup wakeup for rejected HTTP performs`.
  - The full `GlobalTests` class passes 7/7, including the cleanup-focused cases
    (`TestAsyncCleanup`, `TestAsyncCleanupWithHttpCall`, `TestAsyncCleanupWithHttpCallPendingRetry`,
    `TestHttpCallCompletionCallbackIsNotCleanupCancelable`,
    `TestCleanupKeepsNetworkStateForInFlightSingletonRef`,
    `TestHttpPerformRejectedAfterCleanupStarted`).
  - Full TE suite passes on a clean rebuild (102–103/103 across runs). The only stragglers are
    load-induced global XTaskQueue teardown flakes (`VerifyGlobalQueueTermination`, and
    `TaskQueueTests::TestCleanup` `XTaskQueueUninitialize(250)`) that are unrelated to this change,
    pass in isolation on both `main` and this branch, and clear on re-run.

## Scope

- Change is scoped to `NetworkState` admission control and singleton↔`NetworkState` ownership; the
  cancel/await handshake for already-tracked requests is unchanged.
- No backend/provider-specific changes; all platforms pick up the fix through shared code.

## Review focus

- The `m_cleanupStarted` guard sharing the `m_mutex` critical section with the tracking-set
  insert/snapshot (the core Race A fix), and the `E_HC_NOT_INITIALISED` rejection semantics.
- The rejected HTTP perform cleanup path: only requests actually erased from
  `m_activeHttpRequests` may wake cleanup.
- The ownership change: `NetworkState` is no longer moved out or destroyed by cleanup, and is now
  destroyed with the singleton after the `use_count` gate (the core Race B fix).
- The documented invariant that no code may hold a `get_http_singleton()` reference across an async
  wait, and whether it holds for all current callers.

### DIFF
--- a/Source/Global/NetworkState.cpp
+++ b/Source/Global/NetworkState.cpp
@@ -197,6 +197,10 @@ HRESULT CALLBACK NetworkState::HttpCallPerformAsyncProvider(XAsyncOp op, const X
     case XAsyncOp::Cleanup:
     {
         std::unique_lock<std::mutex> lock{ state.m_mutex };
+        // Only a perform that was actually admitted (present in m_activeHttpRequests) may drive the
+        // cleanup wakeup. A perform rejected by the m_cleanupStarted guard was never inserted, so
+        // erase() returns 0 and we must not call ScheduleCleanup()/XAsyncSchedule for it -- doing so
+        // would spuriously (re)schedule cleanup's async block for a request that was never tracked.
         bool scheduleCleanup = state.m_activeHttpRequests.erase(performContext) != 0 && state.ScheduleCleanup();
         lock.unlock();
 
--- a/Source/Global/NetworkState.cpp
+++ b/Source/Global/NetworkState.cpp
@@ -151,6 +151,12 @@ bool NetworkState::CanCleanupCancelHttpRequest(XAsyncBlock* async) noexcept
     }
     return false;
 }
+
+void NetworkState::TestSetCleanupStarted(bool started) noexcept
+{
+    std::unique_lock<std::mutex> lock{ m_mutex };
+    m_cleanupStarted = started;
+}
 #endif
 
 HRESULT CALLBACK NetworkState::HttpCallPerformAsyncProvider(XAsyncOp op, const XAsyncProviderData* data)
--- a/Source/Global/NetworkState.cpp
+++ b/Source/Global/NetworkState.cpp
@@ -174,6 +174,15 @@ HRESULT CALLBACK NetworkState::HttpCallPerformAsyncProvider(XAsyncOp op, const X
         RETURN_IF_FAILED(XTaskQueueCreateComposite(workPort, workPort, &performContext->internalAsyncBlock.queue));
 
         std::unique_lock<std::mutex> lock{ state.m_mutex };
+        if (state.m_cleanupStarted)
+        {
+            // Cleanup has already begun and taken (or is taking) its snapshot of
+            // m_activeHttpRequests under this same mutex. Refuse the request instead of inserting
+            // it after the snapshot, which would orphan it (never canceled or awaited) and could
+            // run it against a torn-down provider.
+            lock.unlock();
+            return E_HC_NOT_INITIALISED;
+        }
         state.m_activeHttpRequests.insert(performContext);
         lock.unlock();
 
@@ -376,6 +385,13 @@ HRESULT CALLBACK NetworkState::WebSocketConnectAsyncProvider(XAsyncOp op, const 
         RETURN_IF_FAILED(XTaskQueueCreateComposite(workPort, workPort, &context->internalAsyncBlock.queue));
 
         std::unique_lock<std::mutex> lock{ state.m_mutex };
+        if (state.m_cleanupStarted)
+        {
+            // See the equivalent guard in HttpCallPerformAsyncProvider: reject connects that arrive
+            // after cleanup has begun rather than orphaning them past the cleanup snapshot.
+            lock.unlock();
+            return E_HC_NOT_INITIALISED;
+        }
         state.m_connectingWebSockets.insert(context->clientAsyncBlock);
         lock.unlock();
 
@@ -501,17 +517,19 @@ HRESULT CALLBACK NetworkState::CleanupAsyncProvider(XAsyncOp op, const XAsyncPro
         {
             std::unique_lock<std::mutex> lock{ state->m_mutex };
             state->m_cleanupAsyncBlock = data->async;
+            state->m_cleanupStarted = true;
             scheduleCleanup = state->ScheduleCleanup();
 
 #ifndef HC_NOWEBSOCKETS 
             HC_TRACE_VERBOSE(HTTPCLIENT, "NetworkState::CleanupAsyncProvider::Begin: HTTP active=%llu, WebSocket Connecting=%llu, WebSocket Connected=%llu", state->m_activeHttpRequests.size(), state->m_connectingWebSockets.size(), state->m_connectedWebSockets.size());
 #endif
-            // No new HTTP performs can enter m_activeHttpRequests after cleanup begins because
-            // http_singleton::singleton_access(cleanup) detaches the singleton before
-            // NetworkState::CleanupAsync runs. Snapshot requests here, then cancel them after
-            // releasing m_mutex. This prevents a race between holding the global cleanup mutex
-            // across XAsyncCancel and allowing completion to advance a request that cleanup has
-            // already decided to cancel.
+            // Setting m_cleanupStarted above (under m_mutex) closes the admission window: any HTTP
+            // perform or WebSocket connect whose Begin op acquires m_mutex after this point is
+            // refused, so nothing can be inserted into the tracking sets after the snapshot below.
+            // Requests that acquired m_mutex before us are already in m_activeHttpRequests and are
+            // captured by the snapshot here. Snapshot them under the lock and cancel them after
+            // releasing m_mutex; this prevents holding the lock across XAsyncCancel while still
+            // ensuring completion cannot advance a request cleanup has already decided to cancel.
             for (auto activeRequest : state->m_activeHttpRequests)
             {
                 auto expectedState = HttpPerformClientBlockState::CleanupMayCancel;
--- a/Source/Global/NetworkState.cpp
+++ b/Source/Global/NetworkState.cpp
@@ -152,10 +152,11 @@ bool NetworkState::CanCleanupCancelHttpRequest(XAsyncBlock* async) noexcept
     return false;
 }
 
-void NetworkState::TestSetCleanupStarted(bool started) noexcept
+void NetworkState::TestSetCleanupStarted(bool started, XAsyncBlock* cleanupAsyncBlock) noexcept
 {
     std::unique_lock<std::mutex> lock{ m_mutex };
     m_cleanupStarted = started;
+    m_cleanupAsyncBlock = cleanupAsyncBlock;
 }
 #endif
 
--- a/Source/Global/NetworkState.cpp
+++ b/Source/Global/NetworkState.cpp
@@ -197,8 +197,7 @@ HRESULT CALLBACK NetworkState::HttpCallPerformAsyncProvider(XAsyncOp op, const X
     case XAsyncOp::Cleanup:
     {
         std::unique_lock<std::mutex> lock{ state.m_mutex };
-        state.m_activeHttpRequests.erase(performContext);
-        bool scheduleCleanup = state.ScheduleCleanup();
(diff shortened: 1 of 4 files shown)


jhugard:
Identified another potential shutdown race not covered by these fixes.  Closing this PR until investigation completes & will resubmit once resolved one way or the other.

jhugard:
We feel comfortable that the other shutdown heisenbug we are seeing has a different root cause, so reopening this PR.

jasonsandlin (approved):

PR #1013, opened 2026-08-10
Add PLM suspend and request-limit features to WinHTTP, removing XCurl GDK path

jasonsandlin:


Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -18,18 +20,7 @@ Result<HC_UNIQUE_PTR<WinHttpProvider>> WinHttpProvider::Initialize()
...
-    {
-        RETURN_IF_FAILED(XNetworkingRegisterConnectivityHintChanged(provider->m_immediateQueue, provider.get(), WinHttpProvider::NetworkConnectivityChangedCallback, &provider->m_networkConnectivityChangedToken));
-    }
-    else
-    {
-        // XNetworking not available (e.g., PC GDK build), assume network is ready
-        provider->m_networkInitialized = true;
-    }
-
     RETURN_IF_FAILED(RegisterAppStateChangeNotification(WinHttpProvider::AppStateChangedCallback, provider.get(), &provider->m_appStateChangedToken));
```
amccalib: This PR removes the soft-fail handling from `CurlProvider::Initialize()`:

```cpp
HRESULT registerHr = RegisterAppStateChangeNotification(CurlProvider::AppStateChangedCallback, ...);
// Suspend handling is a resilience feature; failing to subscribe must not stop HTTP from working
HC_TRACE_ERROR_HR(HTTPCLIENT, registerHr, "...RegisterAppStateChangeNotification failed; suspend handling disabled");
```

but `WinHttpProvider::Initialize()` keeps `RETURN_IF_FAILED`. Net effect is that the WinHttp GDK path ends up less resilient than the XCurl path it replaces.

`RegisterAppStateChangeNotification` resolves from `api-ms-win-core-psm-appnotify`. If it isn't available - PC-GDK under Proton on Steam Deck being the case we care about - `Initialize()` hard-fails and the title gets no HTTP at all. That's the same fail-closed shape as the XNetworking gate this PR is otherwise removing, just reached by a different route.

Suggest porting the CurlProvider pattern here rather than dropping it: log, null the token, continue without suspend handling.

Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -280,22 +395,61 @@ HRESULT WinHttpProvider::CloseAllConnections()
...
+        // may already have been destroyed and so skipped by the callback-clearing loop above);
+        // without this its release would decrement a count that was just zeroed and underflow it.
+        ++m_requestEpoch;
+
+        abandoned.swap(m_pendingRequests);
+    }
+
+    for (auto& pending : abandoned)
+    {
+        HCHttpCallResponseSetNetworkErrorCode(pending.callHandle, E_ABORT, static_cast<uint32_t>(E_ABORT));
```
amccalib, line 496: `CloseAllConnections` now drains `m_pendingRequests` and completes each one with `E_ABORT`. The logic looks right and matches XCurl semantics.

Flagging it because it's new observable behavior for GDK titles rather than a defect: the shipping GDK vpack has no suspend today, so under Quick Resume a title with a deep queue will start seeing a burst of `E_ABORT` completions it has never seen before. Probably worth calling out in the release notes so it doesn't land as partner crash triage.

Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -280,22 +395,61 @@ HRESULT WinHttpProvider::CloseAllConnections()
...
             }
         }
-        WaitForSingleObject(closeContext.connectionsClosedEvent, INFINITE);
+        // Bounded on the PLM suspend path, where the title has a short budget to comply before the
+        // OS terminates it: a connection that fails to report closed in time must not turn a slow
+        // teardown into a watchdog kill. The close context is shared with the callback, so a late
+        // report after this point is still safe. Shutdown passes INFINITE instead, because there
+        // the wait is what guarantees no WinHTTP callback can still run against connections that
+        // allocate through the libHttpClient allocator after the provider and singleton are gone.
+        if (WaitForSingleObject(closeContext->connectionsClosedEvent, timeoutMs) == WAIT_TIMEOUT)
```
BerniME, line 435: shouldn't we retain or re-register connections that time out here? m_connections is already cleared here, so later on shutdown with INFINITE it can retry the ones we couldn't close here?

Source/Global/global_publics.cpp
```diff
@@ -81,6 +81,28 @@ try
 }
 CATCH_RETURN()
 
+STDAPI
+HCSettingsSetGlobalRequestLimit(_In_ uint32_t limit) noexcept
+try
+{
+    // Deliberately does not require initialization: the limit is process-wide state so it can be
+    // configured before HCInitialize creates the provider.
+    xbox::httpclient::SetGlobalRequestLimit(limit);
```
BerniME, line 90: can we act here if there are already request in queue? I think we check this on new request or onRequestsCompleted, but, if we have already requests waiting and we increase limit, can we start processing them immediately? instead of waiting for the next onRequestsCompleted or a new one to arrive.

Not sure how much we can actually win here, or if is a real scenario to change on flight, but wanted to ask

Source/HTTP/WinHttp/winhttp_provider.cpp
jasonsandlin: Good catch, and agreed - fixed in 838377f.

`WinHttpProvider::Initialize()` now matches the `CurlProvider` pattern instead of dropping it: on failure it logs, leaves `m_appStateChangedToken` null, and continues without suspend notifications. The destructor already null-guards the token, so the degraded mode is simply that `Suspend`/`Resume` never fire.

```cpp
HRESULT registerHr = RegisterAppStateChangeNotification(WinHttpProvider::AppStateChangedCallback, provider.get(), &provider->m_appStateChangedToken);
if (FAILED(registerHr))
{
    HC_TRACE_ERROR_HR(HTTPCLIENT, registerHr, "WinHttpProvider::Initialize: RegisterAppStateChangeNotification failed; suspend handling disabled");
    provider->m_appStateChangedToken = nullptr;
}
```

Your framing is the part worth keeping: hard-failing here is the same fail-closed shape as the cached network gate this PR removes, just reached by a different route. Losing PLM notifications should cost the title suspend handling, not HTTP.

Verified after the change: PC `lhc` suite 22/22, and the console `lhc-xbox` suite 12/12 on a Scarlett devkit. The console runs exercise real PLM suspend/resume, which confirms registration succeeds normally and the soft-fail is genuinely just a fallback.

Source/HTTP/WinHttp/winhttp_provider.cpp
jasonsandlin: Agreed, and thanks for separating this from a defect - no code change made here.

Confirming the behavior you describe is intentional: on suspend, `CloseAllConnections` fails anything still sitting in `m_pendingRequests` with `E_ABORT` rather than leaving callers waiting on requests that never reached WinHTTP and have no completion coming.

Two notes on the release-note framing:

1. This repo has no changelog or release-notes file, so there is nowhere in-tree to land it. I have added a partner-facing behavior-change note to the tracking work item instead so it is not lost at release time.
2. Worth stating the trigger precisely for whoever writes the notes: the burst is proportional to *queue depth at suspend*, so it is only visible to titles that exceed the concurrent request limit (default 12) and then suspend. A title that stays under the cap never sees an `E_ABORT`.

Related change in the same area since your review: the drain timeout is no longer a flat 2000 ms. `CloseAllConnections` is shared by `Suspend()` and `~WinHttpProvider`, and bounding the destructor risked a WinHTTP callback outliving the provider and its allocator. It now takes the timeout as a parameter - 2000 ms on suspend to protect the PLM budget, `INFINITE` on shutdown.

Source/HTTP/Curl/CurlMulti.cpp
```diff
@@ -11,27 +10,18 @@ namespace xbox
 namespace httpclient
 {
 
-// XCurl doesn't support curl_multi_timeout, so use a small, fixed delay between calls to curl_multi_perform
+// Some curl implementations do not support curl_multi_timeout, so use a small, fixed delay between calls to curl_multi_perform
```
ScottMunroMS, line 13: now that XCurl is gone I'm not sure you have any more curl implementations with this problem. I would not pollute this PR with an update to address this but you might want to file some tracking to consider addressing this in the future.

Build/libHttpClient.GDK/libHttpClient.GDK.def
```diff
@@ -85,6 +85,8 @@ EXPORTS
     HCRemoveCallRoutedHandler
     HCRemoveWebSocketRoutedHandler
     HCSetGlobalProxy
+    HCSettingsSetGlobalRequestLimit
+    HCSettingsGetGlobalRequestLimit
```
ScottMunroMS, line 89: I'm not sure if it's a goal of yours with this PR but I believe historically when we've needed to make additive changes to DLL ABIs, we've always added new functions to the end of the .def files

adding them in the middle breaks the ABI contract

in practice this may not be a big deal for you. it just means someone can't swap in the new dll without recompiling against the matching object lib. flagging it anyway.

if you action on this, don't forget about the other def files

Source/Global/global.cpp
```diff
@@ -19,6 +19,25 @@ using namespace xbox::httpclient;
 
 NAMESPACE_XBOX_HTTP_CLIENT_BEGIN
 
+// Bounds the memory held by in-flight requests, which matters most on memory-constrained titles.
+// Applied uniformly on all platforms so a single shipped binary behaves identically everywhere.
+constexpr uint32_t c_defaultGlobalRequestLimit = 12;
```
ScottMunroMS, line 24: I suspect this was set at 12 to match how LHC worked for xcurl platforms before this change

but by applying to all platforms I think you've now changed non-xcurl platforms using winhttp (ie Win32 vanilla and GDK.PC)

if you really care about keeping as close to the previous implementation as possible you may want this to stay unlimited for non-Xbox callers and only limit it on xbox.

Source/Global/global.cpp
```diff
@@ -19,6 +19,25 @@ using namespace xbox::httpclient;
...
+
+// Deliberately not part of http_singleton: this must be settable before HCInitialize.
+static std::atomic<uint32_t> g_globalRequestLimit{ c_defaultGlobalRequestLimit };
+
+void SetGlobalRequestLimit(uint32_t limit) noexcept
+{
+    // A limit of 0 would stall every request forever with no way to recover, so treat it as
+    // "restore the default" rather than silently wedging the title.
```
ScottMunroMS, line 32: this feels like a weird way to word this comment. rather than framing this as a workaround, just frame it as the API



```suggestion
    // Callers pass 0 to restore the default request limit.
```

Include/httpClient/httpClient.h
```diff
@@ -211,6 +211,43 @@ STDAPI_(void) HCRemoveCallRoutedHandler(
...
+/// once is capped.
+///
+/// The limit exists to bound the memory held by in-flight requests, which matters most on
+/// memory-constrained titles. The default is 12.
+///
+/// This may be called before HCInitialize(). Changing the value does not affect requests that are
+/// already in flight, and lowering it will not cancel them; the count drains naturally as they
+/// complete.
+///
+/// Platform support: the limit is currently enforced only on GDK (Xbox and PC) and Win32, which
```
ScottMunroMS, line 232: is there value in putting these APIs behind some kind of ifdef to avoid confusing titles? or is that not the convention for LHC.

another pattern could be to have these APIs return E_NOTIMPL on unsupported platforms to prevent titles from not reading these docs and believing they are configuring a limit which isn't taking effect

Source/HTTP/Curl/CurlMulti.h
```diff
@@ -24,14 +24,17 @@ class CurlMulti
...
-    // Curl_multi::WaitForActiveHandles until the suspend watchdog terminates the title
-    // (bug 63050439). The xCurl contract puts the "keep performing" duty on the multi
-    // consumer, so on suspend LHC drives the loop itself instead of relying on that queue.
+    // dispatched (see ScheduleTaskQueueCallback), so this exists for callers that must make
+    // progress without that queue - originally the GDK suspend path for bug 63050439.
+    //
+    // NOTE: currently unreferenced. Its only caller was CurlProvider's GDK suspend handling,
+    // which was removed when GDK moved to the WinHTTP provider. Kept because it is provider
+    // -agnostic and the same need arises for any caller that must drain without the queue;
+    // delete it if no such caller materializes.
```
ScottMunroMS, line 33: this is an internal implementation file right? why is deleting this unreferenced function considered a TODO rather than something to do right now? even if its provider agnostic don't you control all code which might have called it and can safely delete it now? or are you worried someone's forked implementation might rely on it?

its not clear how to tell when/if "no such caller materializes"

Source/HTTP/Curl/CurlMulti.h
```diff
@@ -24,14 +24,17 @@ class CurlMulti
...
     HRESULT PerformUntilDrained(uint32_t timeoutMs) noexcept;
 
-    // Number of requests still owned by this multi handle.
+    // Number of requests still owned by this multi handle. Currently only used by
+    // PerformUntilDrained.
```
ScottMunroMS, line 37: where? I don't think I see any calls within CurlMulti::PerformUntilDrained?

Source/HTTP/WinHttp/winhttp_provider.cpp
jasonsandlin: Good catch - this was a real hole, fixed in c5e5ba5.

You are right that `m_connections` is cleared up front, so a connection that missed the bounded wait was dropped from tracking entirely. That defeats the INFINITE wait on the shutdown path: the destructor would then go on to close the WinHTTP session handles a still-live connection was using.

One wrinkle that shaped the fix - re-closing them is not safe:

```cpp
if (m_connectionClosedCallback)
{
    // WinHttpProvider shouldn't close connection more than once
    assert(!m_connectionClosedCallback);
    return E_UNEXPECTED;
}
```

`WinHttpConnection::Close` is once-only and returns `E_UNEXPECTED` on a second call **without invoking the callback**, so re-registering into `m_connections` and closing again would have counted those connections into the new wait and then blocked on a callback that never fires - an outright hang under INFINITE.

So instead of re-closing, the provider now retains the close context those connections were already given, and a later call waits on that. Their original callback still fires into it, which is why the context is heap-allocated and shared.

One deliberate limitation worth flagging: that retained drain is bounded even when the caller passes INFINITE (5s on the shutdown path). A connection that already missed one deadline is the one least likely to ever report, and I would rather lose the guarantee in that rare case than have `HCCleanup` hang forever. The unbounded wait still applies to connections closed by the current call, which is the case that actually protects the session handles.

Verified: PC `lhc` 22/22, console `lhc-xbox` 12/12 on a Scarlett devkit.

Source/Global/global_publics.cpp
jasonsandlin: Good question, and your instinct is right - the gap is real, and actually a bit worse than "wait for the next completion".

`OnRequestCompleted` promotes **at most one** queued request per completion (it returns as soon as one starts successfully). So raising the limit while a backlog exists does not increase concurrency at all:

- limit 2, five requests submitted: 2 active, 3 queued
- caller raises the limit to 10
- first completion: count 2 -> 1, promote one -> count 2, return

Net effect is still 2 in flight. The extra 8 slots are never used until the queue fully drains, so the raise is effectively a no-op for as long as it would have mattered.

I have deliberately **not** changed it in this PR, for two reasons:

1. `SetGlobalRequestLimit` lives in `global.cpp` and is intentionally decoupled from the provider - it is callable before `HCInitialize` exists. Pumping the queue from there means either adding a method to `IHttpProvider` or reaching into `WinHttpProvider` specifically from the public API layer, which is a layering decision rather than a bug fix.
2. This PR already changed admission control once to fix a slot-count underflow that could wedge HTTP permanently. I would rather not make a second behavioral change to the same code path in the same PR without its own test coverage.

Two ways to fix it when we do, worth deciding between:
- Have the promotion loop fill the budget rather than promoting exactly one. Cheap, and makes the raised limit take effect on the very next completion - but still not immediate.
- Add an explicit "pump pending" path invoked when the limit increases. Immediate, but needs the interface work above.

Happy to do either here if you would prefer it in this PR - otherwise I will file it as a follow-up. Also worth noting the header currently only documents that lowering the limit will not cancel in-flight requests; whichever way we go, it should say what raising it does.

Tests/UnitTests/Tests/GlobalTests.cpp
```diff
@@ -296,6 +296,113 @@ DEFINE_TEST_CLASS(GlobalTests)
...
+
+    // Restores the shipping default so a failure part-way through a test cannot leak a custom
+    // limit into later tests. The limit is process-wide and deliberately survives HCCleanup,
+    // so nothing else resets it.
+    struct RequestLimitRestorer
+    {
+        ~RequestLimitRestorer() { HCSettingsSetGlobalRequestLimit(0); }
+    };
+
+    DEFINE_TEST_CASE(TestGlobalRequestLimitDefault)
```
ScottMunroMS, line 317: it feels like there's a more interesting test missing which actually verifies that whatever the configured value is actually limits the number of requests in flight

Source/HTTP/WinHttp/winhttp_provider.cpp
```diff
@@ -280,22 +405,96 @@ HRESULT WinHttpProvider::CloseAllConnections()
...
+    constexpr DWORD c_retainedCloseTimeoutMs = 5000;
+    DWORD retainedTimeoutMs = (timeoutMs == INFINITE) ? c_retainedCloseTimeoutMs : timeoutMs;
+
+    http_internal_vector<std::shared_ptr<CloseContext>> pendingCloseContexts;
+    {
+        std::lock_guard<std::mutex> lock{ m_lock };
+        pendingCloseContexts.swap(m_pendingCloseContexts);
+    }
+
+    for (auto& pendingContext : pendingCloseContexts)
```
amccalib, line 465: Should this drain be limited to the shutdown path?

It sits outside the `openConnections > 0` block, so it also runs on suspend - and there `retainedTimeoutMs` is 2000, waited per context in sequence. Worst case suspend becomes `2000ms + N x 2000ms`. Since these are, per your comment, the connections least likely to ever report, they'd burn the full timeout on each pass rather than returning early. It also delays the `E_ABORT` completions below by the same amount.

That seems to cut against the bounded wait's stated intent 30 lines up - "must not turn a slow teardown into a watchdog kill." Given the retained contexts exist to protect provider destruction, and the provider isn't destroyed on suspend, is anything gained by draining there? Wrapping the loop in `if (timeoutMs == INFINITE)` would keep it to shutdown.

Source/HTTP/WinHttp/winhttp_provider.h
```diff
@@ -121,25 +138,81 @@ class WinHttpProvider
...
     // Track WinHttpConnections so that we can close them on shutdown/suspend
+    // Shared with the connection-closed callbacks so a connection that reports closed after a
+    // bounded wait has already returned still has a valid context to signal. Defined in the .cpp.
+    struct CloseContext;
+
+    // Contexts from earlier CloseAllConnections calls whose bounded wait expired with connections
+    // still open. Those connections were dropped from m_connections and cannot be closed a second
+    // time (WinHttpConnection::Close is once-only and returns E_UNEXPECTED without invoking the
+    // callback), so waiting on the context they were originally given is the only way to observe
+    // them finishing. Shutdown drains these with an INFINITE timeout, which is what keeps a
```
amccalib, line 174: This says shutdown drains with `INFINITE`, but `CloseAllConnections` converts that to `c_retainedCloseTimeoutMs` (5s). The `.cpp` comment has it right - worth matching here so the header doesn't promise a guarantee the code intentionally gives up.

Source/HTTP/WinHttp/winhttp_provider.h
```diff
@@ -121,25 +138,69 @@ class WinHttpProvider
...
-    // Maintain a WinHttpSession for each unique security protocol flags
-    http_internal_map<uint32_t, HINTERNET> m_hSessions;
+    // Maintain a WinHttpSession for each unique (security protocol flags, secure scheme) pair.
+    //
+    // The scheme is part of the key because sessions are not interchangeable across schemes:
+    // GetHSession opens HTTPS sessions with WINHTTP_FLAG_SECURE_DEFAULTS, which permanently
+    // restricts that session to secure requests, and opens plain HTTP sessions with only
+    // WINHTTP_FLAG_ASYNC. Keying on the protocol flags alone let whichever scheme ran first win
+    // the cache slot, so an http:// or ws:// request that followed an https:// request reused the
+    // secure-defaults session and failed in WinHttpOpenRequest with ERROR_ACCESS_DENIED.
```
ScottMunroMS, line 148: nit: this led me to believe you distinguished between schemes and not scheme security levels

i.e. if you were distinguishing between schemes then HTTPS and WSS are different but they are not different here and intentionally reuse the same sessionkey


```suggestion
    // `isSecure` is part of the key because WinHTTP sessions cannot be reused across TLS and non-TLS 
    // connections. TLS-enabled sessions are restricted with WINHTTP_FLAG_SECURE_DEFAULTS which
    // then prevents those sessions from being used with insecure WS and HTTP schemes.
```

ScottMunroMS (commented):
some food for thought. no hard blockers

Tests/UnitTests/Tests/GlobalTests.cpp
ScottMunroMS: oh I think the tests I'm looking for may exist outside of this repo where some of the other LHC tests have ended up. is that correct?

Source/Global/global_publics.cpp
BerniME: sounds good, not blocker for me, a follow up task will do it

Source/HTTP/Curl/CurlMulti.cpp
jasonsandlin: Agreed, and thanks for flagging that it shouldn't land here. Filed [Task 63532904](https://microsoft.visualstudio.com/Xbox/_workitems/edit/63532904) to replace the fixed `PERFORM_DELAY_MS 50` with `curl_multi_timeout`.

You're right that the premise is gone: with XCurl removed, the only remaining consumer of this file is Linux against real libcurl (built by `Build/libHttpClient.Linux/CMakeLists.txt`), which supports `curl_multi_timeout` fine. Not touched in this PR.

Build/libHttpClient.GDK/libHttpClient.GDK.def
jasonsandlin: Good catch, fixed in 795af0b.

You're right that this is the convention, and the existing tails already show it: `HCHttpCallRequestGetMaxReceiveBufferSize` / `SetMaxReceiveBufferSize` sit at the end, out of alphabetical order. None of these .def files use explicit ordinals, so inserting mid-file shifted every subsequent export.

Moved the two new exports to the end of **all four** files (GDK, GDK.NoWebSockets, Win32, Win32.NoWebSockets). The diffs are now pure appends.

Source/Global/global.cpp
jasonsandlin: You're right that this changed behavior for non-Xbox platforms, and it's fixed in 795af0b.

One correction to the premise: there was no request limit on **any** platform before this PR, so 12 wasn't inherited from XCurl. It was new everywhere, which made the regression risk broader than intended.

The default is now 12 only on Xbox consoles, and unlimited elsewhere, so Win32 and GDK.PC keep their existing uncapped behavior unless a title opts in. It's resolved at **runtime** via `XSystemGetDeviceType()` rather than at build time, because the same Gaming.Desktop binary ships on Xbox, GDK PC, Steam and Steam Deck.

Source/Global/global.cpp
jasonsandlin: Applied your suggestion verbatim in 795af0b.

Include/httpClient/httpClient.h
jasonsandlin: Went with your `E_NOTIMPL` suggestion in 795af0b. Silently storing a value that never throttles was the worse failure mode.

Both `HCSettingsSetGlobalRequestLimit` and `HCSettingsGetGlobalRequestLimit` now return `E_NOTIMPL` on platforms whose provider has no admission control (that lives in the WinHTTP provider, which is only built for Win32 and GDK). Header docs updated to match.

This pairs with the scoping fix above: PC is no longer silently throttled, and platforms that can't enforce the limit are now detectable rather than quietly inert.

Source/HTTP/Curl/CurlMulti.h
jasonsandlin: Fair, no reason to defer it. Deleted `PerformUntilDrained` and `ActiveRequestCount` outright in 795af0b, along with the `<chrono>` / `<thread>` includes they were the only users of, which matters here because the Linux build compiles with `-Werror`.

Source/HTTP/Curl/CurlMulti.h
jasonsandlin: You're right, there were none: `PerformUntilDrained` had no callers at all. Both it and `ActiveRequestCount` are deleted in 795af0b. `PerformStepLocked` is retained, since `Perform()` still uses it.

Tests/UnitTests/Tests/GlobalTests.cpp
jasonsandlin: Correct on both counts. The unit tests here only cover the get/set contract; the enforcement tests live outside this repo, in PlayFab.C's `lhc-xbox` suite, which runs against a real devkit:

- **05** - get/set round trip
- **06** - a 12-request burst against a cap of 3 must queue at least 9 and still succeed
- **08** - the budget is restored after suspending at capacity
- **11** - repeated suspend/resume cycles leak no slots
- **14** - WebSockets must not consume HTTP budget

They're in the companion PR.

Source/HTTP/WinHttp/winhttp_provider.cpp
jasonsandlin: You're right, and it's fixed in 795af0b: the retained drain is now gated to the shutdown path only.

On suspend the cost was additive per retained context (`2000 + N x 2000`), which is exactly what the PLM watchdog punishes. Suspend stays bounded; only shutdown does the unbounded drain.

Source/HTTP/WinHttp/winhttp_provider.h
jasonsandlin: Fixed in 795af0b. The header now describes the bounded timeout instead of promising `INFINITE`, matching the .cpp.

Source/HTTP/WinHttp/winhttp_provider.h
jasonsandlin: Good nit, applied in 795af0b.

You're right that it's the security level rather than the scheme: HTTPS and WSS deliberately share a key (`Uri::IsSecure()` covers `wss`), and what can't be shared is a TLS session with a non-TLS one.

Source/Global/global_publics.cpp
jasonsandlin: Filed [Task 63532905](https://microsoft.visualstudio.com/Xbox/_workitems/edit/63532905) for this, thanks.

It captures the two details that make the fix non-trivial: `OnRequestCompleted` promotes at most one queued request per completion, and the setter lives in `global.cpp` as pre-`HCInitialize` process state with no handle to the provider. So the fix needs a provider hook rather than a change in the setter.

amccalib (approved):

amccalib (approved):

PR #1016, opened 2026-08-31
Fix _WIN32 macro checks

eofenstein:
Use `#if defined(_WIN32)` instead of `#if _WIN32` to avoid undefined macro warnings on non-Windows platforms.


eofenstein:
@jasonsandlin for review

jasonsandlin:
you'll need to re-create this PR not in a fork but from a branch of the primary LHC repo so the PR gates kick on.  Thanks

